PR #114, opened 2012-06-14
Added zindex options to dropdown

pier-oliviert:
On positionDropdown(), we assign zindex to the dropdown that can be specified in the options as zIndex.

By default this value is 2000 (Modal is usually in the 1000s)

### DIFF
--- a/select2.js
+++ b/select2.js
@@ -540,6 +540,7 @@
                 formatInputTooShort: function (input, min) { return "Please enter " + (min - input.length) + " more characters"; },
                 minimumResultsForSearch: 0,
                 minimumInputLength: 0,
+                zIndex: 2000,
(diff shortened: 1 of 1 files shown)


valaxy:
As version of v4.0.2, if you want to set zIndex to dropdown yourself, some tricks can be used:

```js
const $ = require('jquery')
const select2 = require('select2')

$.fn.select2.amd.require(['select2/dropdown/attachBody'], function (AttachBody) {
    let render                  = AttachBody.prototype.render
    AttachBody.prototype.render = function () {
        let result = render.apply(this, arguments)
        this.$dropdownContainer.css('z-index', this.options.get('zIndex'))
        return result
    }
})

$dom.select2({ zIndex: 999 })
```

Hope this help

PR #265, opened 2012-07-25
dropdownAttached option.

traversal:
It would be great to get this brought into master, as I have been having trouble with the way the drop down is appended to the body and positioned alongside the link control.

---

Added dropdownAttached option (false by default) which can simplify integration by avoiding issues where the drop down can break away from the link control when used inside jQuery UI dialogs etc.

Note that this setting does NOT allow for the newly introduced feature where the drop down is positioned above if there is no space, since it essentially prevents positionDropDown from being called altogether.

### DIFF
--- a/select2.js
+++ b/select2.js
@@ -742,6 +742,7 @@
 
         // abstract
         positionDropdown: function() {
+          if (!this.opts.dropdownAttached) {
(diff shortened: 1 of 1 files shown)


sooraj-rc:
www

PR #308, opened 2012-08-08
Fix HTML5 validation

### LINKED ISSUE #128: Original select has 'required' attribute, doesn't work fine
Hello,

I think this is the better plugin I've ever used ;)

My question is related to the 'required' attribute.

let me explain better
I have a html page like this:

``` html
<form name="myForm" action="myAction" >
   <select id="myId" name="myName" required="true" >
       <option value="myvalue">An Element</option>
   </select>
   <button type="submit" >Go</button>
</form>
<script>
    $("#myId").select2({placeholder: "Select"});
</script>
```

When I click on the submit button of the form containing the select above, I see the internal browser's message "Please select an element from the list". That's right, because I want browser shows the message.

The problem is that the message is not correctly shown, maybe caused by "display: none" of the original select when I do:
$("#myId").select2()

Is there a way to avoid this annoyance?

thanks in advance and good job again ;)

mcuelenaere:
This fixes HTML5 validation popups in browsers.
Instead of hiding the <select>, move it under the container.

This was briefly tested on Chrome, Firefox and Opera under Linux.

Under Chrome, some <select> tags seem to overlap for ~2px; I'm guessing calling .offset() to get the absolute position of the container isn't the correct way to do.

(this should fix #128)

### DIFF
--- a/select2.js
+++ b/select2.js
@@ -511,18 +511,16 @@
             this.initContainer();
             this.initContainerWidth();
 
-            // position element where the container is
-            this.opts.element
-                .css({
-                    position: "absolute",
-                    "z-index": 0,
-                    top: this.container.offset().top,
-                    left: this.container.offset().left,
-                    width: this.container.width(),
-                    height: this.container.height()
-                });
             // make sure the container is above the element
+            this.opts.element.css({
+                position: "absolute",
+                "z-index": 0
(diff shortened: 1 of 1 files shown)


radioactiveboyx:
i solved it using required="true"

Issue #798, opened 2013-02-04
Can't detect the Focus & Blur events


TheSisb:
I'm doing 

``` javascript
$('#el').select2();
$('#el').on('focus,blur,change', function(e){
    console.log($(this), e);
});
```

However only the change event is triggered.  I checked the documentation and see no mention of focus and blur.  I find that quite strange.  Is there any way to achieve that functionality?

PR #866, opened 2013-02-15
Add blur/focus event triggerings (#798)

iffy:
An attempt to add blur/focus events.  Related to issue #798

I ran in to problems using the real 'blur' and 'focus' events, so I simplified the problem by using 'select2:blur' and 'select2:focus.'

Also, when switching between the `container` and `search`, a blur and focus event are emitted.  I'd like to avoid that, if possible, but I'm not sure how to, and it's late.

### DIFF
--- a/select2.js
+++ b/select2.js
@@ -1737,7 +1737,7 @@ the specific language governing permissions and limitations under the Apache Lic
                     this.container.removeClass("select2-container-active");
                 }
                 if (!this.container.hasClass("select2-container-active")) {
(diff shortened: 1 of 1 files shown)


samholmes:
I can't capture these events either. What is the proper syntax?

arturoribes:
Same here, how to capture the focus event? I think it is quite a desirable feature to have the select2 to auto-open on focus, but at least it should be possible to bind the focus event. I tried focus, select2-focus and select2:focus with no luck.

PR #1604, opened 2013-08-09
Add 'nextSearchTerm' function

jdecuyper:
Clean version of [pull request #1519](https://github.com/ivaynberg/select2/pull/1519):
- Given a previous selection `nextSearchTerm` determines what the next search term should be.
- Remove some leftover tabs.
- Fix a typo: 'install the' instead of 'install sthe'.


michaelblyons:
Am I correct that this no longer works in 4.0?

jdecuyper:
YES you are!
@ivaynberg is there now an easy way to achieve this behavior with the latest version?

bpromasmb:
I would also like to know if there's a way to do this in v. 4.0

Issue #1538, opened 2013-07-16
Ability to hide currently selected value in dropdown


srigi:
When dropdown is opened, items are rendered inside of `ul.select2-results`. Every single item is wrapped in `li.select2-result`, the one item that is hovered by mouse is `li.select2-result.select2-highlighted`.

I would really appreciate if there is also some other class added to currently selected item in this `ul.select2-results`. That way I can target it by CSS and hide that item.

rahmathm1:
@kevin-brown  Is this working with list being fetched as ajax? thanks!

kikonen:
It looks like just hiding elements using "display: none" doesn't remove  them from keyboard iteration; so arrow  up/down hits hidden elements, which are invisible.

blazkovicz:
This issue was closed because of work in #2053 but that pull request was reverted, so the problem remains. Do authors of select2 4.0 intend to solve this isuue any time soon? I just updated select2 from 3.5 and discovered this annoing behaviour.

djtarazona:
+1

Hiding with CSS isn't a good solution as the options are still in the keyboard iteration and someones the drop down appears empty with no explanation / message to the user.

nhieu92:
Hi. I want to hide selected option. How to do that?

belov91:
@nhieu92  +1

guill3m:
> Select2 4.0 consistently marks selected objects, so you can now hide them through CSS

**Hiding options with CSS is not a good solution:** in fact, it is not a solution at all. As it has already been said if you try to navigate the options with the keyboard keys it will still cycle through the hidden options, which makes hiding the selected options with CSS an even a worse solution than leaving them there. So I would say this issue is not actually about _hiding_ but _removing_ the selected values from the dropdown.

> It's a new feature in 4.0.0 that was introduced largely for accessibility reasons.

@kevin-brown you mentioned that in #3850. Could you extend on those accessibility reasons please? I would like to understand your argument against this feature request but can't find anything that helps me understand it in the docs.

A couple of reasons to remove the already selected options from the dropdown:
-  In `select[multiple]` elements if there is several (maybe 10+) options and you have most of them selected, showing only the not selected ones helps the user to know what is not selected (the selected ones are already being shown anyway).  
- In `select[multiple]` elements to be able to quickly select several options (maybe even all of them) with the keyboard: space + enter + space + enter + ...

[...]

Issue #1267, opened 2013-05-07
Select a tag multiple times


aL3xa:
A dead simple one: I want to be able to select a tag, but multiple times. How? :smile:

kevin-brown:
It is now possible to do this if you inject the new tag as an `<option>` each time it is requested.

PR #2032, opened 2014-01-12
Add hideSelectionFromResult property

jdecuyper:
Add a property that allows the user to hide or display the selected item(s) in the result list. This functionality can be used for both single and multi select boxes. It was tested with inline options, data source, array data and tag array.

By default, the `hideSelectionFromResult` boolean property is `undefined` and leaves the expected behavior intact: the selected item from a single select box is visible inside the result list while the selected items from a multi select box are hidden from the result list. This default behavior can be overriden by setting the `hideSelectionFromResult` property to `true` or `false`.

Related issues: 
#1267 - Select a tag multiple times
#1538 - Ability to hide currently selected value in dropdown

### DIFF
--- a/select2.js
+++ b/select2.js
@@ -2225,13 +2225,14 @@ the specific language governing permissions and limitations under the Apache Lic
 
         // single
         postprocessResults: function (data, initial, noHighlightUpdate) {
-            var selected = 0, self = this, showSearchInput = true;
+            var selected = 0, selectedElm = null, self = this, showSearchInput = true;
 
             // find the selected element in the result list
 
             this.findHighlightableChoices().each2(function (i, elm) {
                 if (equal(self.id(elm.data("select2-data")), self.opts.element.val())) {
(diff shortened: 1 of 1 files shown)


ryudice:
Does this still works on the latest version? I cant find that option.

ryudice:
Nevermind, I just saw it was reverted. Is there an alternative way to achieve the same?

kevin-brown:
This should work in the latest version and give the same effect.

```css
.select2-results__option[aria-selected=true] {
  display: none;
}
```

enderahmetyurt:
@kevin-brown thank you :+1:

victoryam:
This doesn't fix everything (for me at least).

I seems like the hidden options are still being selected, so if I select the option "Bear", then type "B" in the search, the first option might seem as "Beetle" but it will not be selected by default, because "Bear" is selected even if not visible.

PR #2063, opened 2014-01-21
Add support for numbered (non-infinite-scrolling) pagination

LogicAndTrick:
As requested in #1475, this has been rebased onto the `4.0-wip` branch. 
- Set the `numPages` value in the results to the total number of pages to enable this style of pagination. This takes precedence over the `more` value, so if both keys are included in the results, numbered pagination will be used instead of the infinite scrolling method.
- When this pagination method is active, Page Up and Page Down will jump to the previous and next pages respectively. When Shift is held, they will jump to the first and last pages instead.

The final result looks like this:

select2 pagination

The 'Infinite scroll with remote data' example can be modified for explicit pagination by changing the results function like so:

``` javascript
function (data, page) {
    var numPages = Math.ceil(data.total / 10); // Get the total number of pages
    // Set the numPages value to use explicit pagination
    return { results: data.movies, numPages: numPages };
}
```


maxgalbu:
@kevin-brown Is this feature still available? I couldn't find these lines in the current source code and there's no example in the docs.

PR #2140, opened 2014-02-22
Add `shouldFocusInput` option [Fix #1541]

### LINKED ISSUE #1541: Virtual keyboard showing up always
If select2 is used the virtual keyboard is showing up always if a user opening or closing an select.

It's caused at least by two reasons: 
a) focus is called on search input field on opening select (line 1785 in select2-3.4.1)
b) focus is called on focusser input field on close (line 1808), focus (line 1817), cancel (line 1830)

The virtual keyboard is also showing up if search is disabled by minimumResultsForSearch. So users of mobile devices get virtual keyboard but can't input something. Also virtual keyboard is overlaying select options. So select is nearly not usable on mobile device.

kevin-brown:
This adds a `shouldFocusInput` option that determines whether or not the focusser should be automatically focused.  By default, the focusser will always be focused, unless `minimumResultsForSearch` is less than zero.

`close` on single selects took an option `params` argument which at one point was used to tell the focusser to not automatically focus.  This was [added at one point](https://github.com/ivaynberg/select2/commit/e162a4802cddfb95fdef8793fc83fc07897650ec) when tweaking the mask and [later reverted](https://github.com/ivaynberg/select2/commit/9bc68f089e4d81e0dd0e8de943071ae36a3394cd) to fix a bug that came as a result.  The leftovers of this code has been removed in this commit.

### DIFF
--- a/select2.js
+++ b/select2.js
@@ -1340,7 +1340,7 @@ the specific language governing permissions and limitations under the Apache Lic
                         if (self.opts.selectOnBlur) {
                             self.selectHighlighted({noFocus: true});
                         }
-                        self.close({focus:true});
+                        self.close();
                         e.preventDefault();
                         e.stopPropagation();
                     }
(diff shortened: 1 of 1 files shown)


ksikka:
Is the `shouldFocusInput` option meant to be used by the end user? Saw that it was not documented, wondering if that was intentional or not.

Issue #2226, opened 2014-04-01
Mobile (iOS 7) select2-result Selection Leaks Click Through to Next Screen


kimardenmiller:
When on iOS7 and tapping an item from the Select2 list:

![image](https://cloud.githubusercontent.com/assets/2165336/2585255/e8258264-b9e9-11e3-804c-ab94670ba963.png)

the tap selects item on the list fine, but then also gets picked up again on the next screen. In the above example the index view grid behind the select2-result got a clicked too, resulting in ...

![image](https://cloud.githubusercontent.com/assets/2165336/2585274/1d45f96a-b9ea-11e3-9eb7-03bb1acfb6b7.png)

what you now see: the show_detail screen that resulted from that leaked click.

Easily reproducible on iOS 7 [here](http://staging.spokenvote.org/proposals?hub=1).

Happens with a long list, short list, simulator, so is not fat fingers.

PR #2239, opened 2014-04-03
Fix for "leaking clicks"

antstorm:
Ok, this is another solution for clicks that are triggered on underlying elements on mobile devices.

This time `mousedown` and `mouseup` events are left intact. Instead touch-events are marked, but we are still waiting for a 'click' event to arrive. And once we got the click event we execute `selectHighlighted`, but only for marked events.

Fix for issue #2226.


branot89:
When I type a name and click on drop down list, select uses focus and I can't type anymore. Dropdown list is still visible but I can't type. How to fix that?

PR #2343, opened 2014-05-05
Added option to keep search results after select close

dmitry-dedukhin:
Hi!

This is a clean patch for https://github.com/ivaynberg/select2/pull/1902 (in 1902 I've made bad rebase and don't know how to fix it).


lewa:
It would be nice to have such feature in Select2 4.0.

PR #2505, opened 2014-07-01
Make it possible to force open the dropdown in a specified direction

matteodem:
In our project we had the specific case that our client wanted the dropdown to **only** open on south / north, even if there's no space anymore to display all the suggestions there.

### DIFF
--- a/select2.js
+++ b/select2.js
@@ -1311,6 +1311,11 @@ the specific language governing permissions and limitations under the Apache Lic
                 width: width
             };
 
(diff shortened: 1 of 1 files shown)


helarqjsc:
>This has come up a few times, and there does not appear to be any interest in putting it in the core.

There are like 10 issues about this exact problem, if not more. What do you mean "there is no interest"? I mean, it's your library and you can do whatever you want with it, but it's pretty obvious that a lot of people need this functionality.

Issue #2264, opened 2014-04-09
closeOnSelect = false does not work when minimumInputLength > 0


shiznit013:
The dropdown will close on select.  I have created a [jsfiddle](http://jsfiddle.net/7Nt7U/1/) to demonstrate the behavior.

This worked in 3.4.5.  I believe (although have not confirmed), that this may have been broken by pull request https://github.com/ivaynberg/select2/pull/1940

AtomStar:
This does not work in v4.0 beta2. @snichini fix is for v3.5. I did the following hack on lines 3384-3388 in v4, which seems to work:
``` javascript      
       Dropdown.prototype._onSelect = function () {
            if (this.options.options.closeOnSelect == false)
                return;
            this.trigger('close');
        };
```

kevin-brown:
@AtomStar If you want to roll that up into a decorator, it'd save me the time of adding it back myself. It shouldn't take too long and you can use https://github.com/select2/select2/pull/2914 as an example for adding in the option.

I was planning on adding it back for 4.0, but haven't had the time.

amamf:
The above fix provided by snichini doesn't work for me (3.5.2). It works for the first item only, when I select the second item the search results get closed. I came up with a different solution that works for me. The idea is not to update search results if they shouldn't be closed on select: 
```javascript
           // keep track of the search's value before it gets cleared
            this.nextSearchTerm = this.opts.nextSearchTerm(data, this.search.val());

            this.clearSearch();

            if (this.select || !this.opts.closeOnSelect) this.postprocessResults(data, false, this.opts.closeOnSelect===true);

            if (this.opts.closeOnSelect) {
                this.updateResults();
                this.close();
                this.search.width(10);
            } else { ...
```

AlexRocks:
If you do not want to hack the select2 3.5 code then following work ('select2-selecting' event launch before close event so we can manually close/open):
`$("#category_id").on("select2-selecting", function(e) {
                e.preventDefault();
//This one upload from remote server some additional info
                    uploadSubCats(e.val).then(function(res) {
//After that we can make decision close or not close dropdown
                        if (res.status == "ok") {
                            if (res.data === 0) {
                                $('#category_id').select2("val", val);
                                $('#category_id').select2("close");                                
                            } else {
                                $('#category_id').select2({"data": res.data});
                                $('#category_id').select2("open");
                            }
                        }
                    });                                    
            });`

joergrech:
[...]

PR #2608, opened 2014-08-12
Fix for issue #2264 Do not reset the search if minimumInputLength is > 1...

zacharyrankin:
..., multiple is true, and closeOnSelect is false


kevin-brown:

Issue #898, opened 2013-02-20
Selecting option with TAB should focus next input


eeng:
It would be nice if TAB works in the same way as in the Chosen library, i.e. when the dropdown is open and an option is selected with TAB key, the focus should change to the next input. Currently TAB is the same as ENTER, focus just stays in the select requiring another TAB to switch to the next input.

PR #2635, opened 2014-08-22
changing tab behaviour in multiselect; not focussing immediately on next field

murb:
This minor change is a partial revert of 4094c1c (that solved #898). I think the behaviour is requested in that issue is correct for the  traditional, *single* option select. For the multi select version, however, it makes more sense to keep the focus on the field, since you may want to add another value while typing. Although it is the same plugin, I don't think the different behaviour is conflicting: the appearance is entirely different. I wouldn't want my users to press Enter too often: because in normal text-input fields using Enter leads to submitting the entire form.

### DIFF
--- a/select2.js
+++ b/select2.js
@@ -2761,7 +2761,7 @@ the specific language governing permissions and limitations under the Apache Lic
                         return;
                     case KEY.TAB:
                         this.selectHighlighted({noFocus:true});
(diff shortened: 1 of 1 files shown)


formikea:
+1

kevin-brown:

PR #2648, opened 2014-08-28
Made selectOnBlur a function so user can decide when selection will occur upon blur event

broubach:
I had a situation where the selectOnBlur event could not be a static true or false. It had to be decided at runtime. I hope others can take advantage over this too.


kevin-brown:
Closing this off considering the upgrade path breaks and the upcoming release of 4.0.

PR #2698, opened 2014-09-24
Fixed up left and width css so left can't go negative and width adjusts ...

jeffreydwalter:
...when dropLeft goes negative.

When dropdownAutoWidth = true, and you resize your browser to a size where you have more room on the left than you do on the right hand side of the page, with the select2 options list open, the option list goes off the left-hand side of the page. 
I modified the css calculations to ensure that css.left is never less than 0, which keeps the options list from going off the page, and css.width = width + dropLeft which ensure proper width of the options list, only when dropLeft < 0.

### DIFF
--- a/select2.js
+++ b/select2.js
@@ -1314,8 +1314,8 @@ the specific language governing permissions and limitations under the Apache Lic
             }
 
             css =  {
-                left: dropLeft,
-                width: width
+                left: dropLeft > 0 ? dropLeft : 0,
+                width: dropLeft > 0 ? width : width + dropLeft
             };
 
             if (above) {


kevin-brown:

jeffreydwalter:
Why wasn't this pull request merged in seven months ago?

kevin-brown:
_Most likely_ because it was lacking a place where we could check the change out at. Also I have no idea where the width change comes into play.

jeffreydwalter:
Looks like the codebase has changed substantially. If I come across the bug
in the latest version I'll attempt to port this fix to the new code.

Thanks.

PR #2825, opened 2014-11-24
Result open and intermediately close on change .select2-drop

xavadu:
The problem is that on mousedown an element is selected, when the dropdown is open under the 'select' element (where you click to open the dropdown), the mouse pointer is over the dropdown then on release the mouse click the element over the mouse pointer is selected.

https://github.com/ivaynberg/select2/issues/2823


vladan-me:
Honestly, I would rather two times click than to have this "auto-select feature"... It can be really annoying. That's also a reason why I am reading this... Patch works fine though. Thanks!

Issue #1311, opened 2013-05-14
MultiSelect JavaScript error - no method 'showSearch'


timbonicus:
Uncaught TypeError: Object [object Object] has no method 'showSearch'

The MultiSelect2 class has no showSearch method. The error occurs with the newly-released 3.4.0.

This was introduced in https://github.com/ivaynberg/select2/commit/bcfa3fc702f29485d36ebdc9c5b12c05d1d551bd

To fix it temporarily, change line 1518:

``` javascript
if (initial && this.showSearch) this.showSearch(true);
```

PR #2828, opened 2014-11-25
Hide search box if select is empty

LudoMon:
Fix #2822.
When a select box is empty (this appends in my case, I have a dynamic select box), the search box should not be shown.


franckl:
This commit is creating a "Uncaught TypeError: this.showSearch is not a function" when using the latest stable/3.5 branch.

See https://github.com/select2/select2/issues/1311 (same issue already fixed in the past)

vevisystems:
I think this is an unfortunate change, because if createSearchChoice() is provided, the search box should be visible in order for the user to enter the desired text. What about changing:

this.showSearch(search.val());

with:

this.showSearch(search.val() || this.opts.createSearchChoice);

PR #2846, opened 2014-12-05
Fix bug where elements in select2 drop cannot be focused

colemanw:
Background: CiviCRM builds on select2 by adding some extra filters and controls to the select2 drop area:

select2_drop

Tracking down why the user was being prevented from interacting with the controls, I found that this chrome workaround had the undesired side-effect of preventing anything in the select2 drop from being focused (affected single but not multi selects for some reason). Making the condition more specific to the actual problem being worked around solves this.

### DIFF
--- a/select2.js
+++ b/select2.js
@@ -2181,7 +2181,7 @@ the specific language governing permissions and limitations under the Apache Lic
                 // without this the search field loses focus which is annoying
                 if (document.activeElement === this.body.get(0)) {
                     window.setTimeout(this.bind(function() {
-                        if (this.opened()) {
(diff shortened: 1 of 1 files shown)


colemanw:
Either way, it would be awesome to see this resolved in 3.5.x so that we can release the next version of CiviCRM with the official upstream code from select2.

colemanw:
Also just ran across the bug fixed by https://github.com/select2/select2/pull/2925
It would be awesome to see an 3.5 release that incorporates that fix as well as either of the proposed fixes for this issue.
Thanks again for all your hard work @kevin-brown

colemanw:
Great, thanks!

PR #2848, opened 2014-12-09
Fix performance issue with setVal on MultiSelect

tedliang:
As suggested in https://github.com/ivaynberg/select2/issues/781#issuecomment-30935226, significantly improves selection performance (90~100x faster) on a list of 2,000 items.

Also see: https://github.com/ivaynberg/select2/issues/905 and https://github.com/ivaynberg/select2/issues/2706

It will take 1s to 'select all' in patched version http://jsfiddle.net/hrro7j6L/3/ and 93s in head version http://jsfiddle.net/hrro7j6L/4/

### DIFF
--- a/select2.js
+++ b/select2.js
@@ -3261,11 +3261,9 @@ the specific language governing permissions and limitations under the Apache Lic
                 for (var j = 0; j < old.length; j++) {
                     if (equal(this.opts.id(current[i]), this.opts.id(old[j]))) {
                         current.splice(i, 1);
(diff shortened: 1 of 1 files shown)


TimMoore:
I see that this is in the 3.5 stable branch, but has not yet been released. Is there any chance of a 3.5.3 release for those of us that aren't ready to upgrade to 4.0 yet?

daveberlin:
+1, would like to go to 3.5.3

PR #2887, opened 2014-12-26
Added Serbian language

Vodzo:


kevin-brown:
This actually already exists (incorrectly) under the `rs` code in 3.5.x, and (correctly) under the `sr` code in 4.0.

PR #2891, opened 2015-01-02
Add ajax url function parameters

nikmartin:


kevin-brown:
Given that this [has changed in 4.0](https://select2.github.io/options.html#ajax) (and is better documented), I'm going to hold off from merging these into 3.5.x.

PR #2914, opened 2015-01-12
add maximum selection length option

smzchaudary:
Adds the ability to restrict the max number of selected items.


src/js/select2/data/maximumSelectionLength.js
```diff
@@ -0,0 +1,29 @@
...
+  function MaximumSelectionLength (decorated, $e, options) {
+    this.maximumSelectionLength = options.get('maximumSelectionLength');
+
+    decorated.call(this, $e, options);
+  }
+
+  MaximumSelectionLength.prototype.query =
+    function (decorated, params, callback) {
+
+    var count = this.$element.val() != null ? this.$element.val().length : 0;
```
kevin-brown: We might want to use `this.current()` here, to support cases where the current value is not actually stored in `$element.val()` (like the old hidden input, if we ever choose to support it again). This will also avoid the issue of it triggering an error if used on a non-multiple select (instead it just won't work, which more _slightly more_ sense).

kevin-brown:
This looks pretty good, just one inline comment that might need to be fixed.

Can you add tests so we can ensure that this isn't unintentionally broken in the future? And perhaps some docs for it (and/or add an example)?

smzchaudary:
@kevin-brown: Added tests as well as an example.
Let me know if you need anything else to merge this in.

docs/examples.html
```diff
@@ -76,6 +76,35 @@
...
+      <h2>Example code</h2>
+
+      <pre data-fill-from=".js-code-multiple-limit"></pre>
+
+<script type="text/x-example-code" class="js-code-multiple-limit">
+$(".js-example-basic-multiple-limit").select2({
+  maximumSelectionLength: 2
+  });
+
+<select class="js-example-basic-multiple" multiple="multiple">
```
kevin-brown: This can probably be dropped, there isn't anything special about the HTML for this, so I don't think it needs to be reinforced.

docs/examples.html
```diff
@@ -76,6 +76,35 @@
     </div>
   </section>
 
+  <section id="multiple_max" class="row">
+    <div class="col-md-4">
+      <h1>Limited Multiple select boxes</h1>
```
kevin-brown: "Limiting the number of selections" might be a better title for this.

smzchaudary:
@kevin-brown example changed and moved, as requested

docs/examples.html
```diff
@@ -33,7 +33,7 @@
 
       <pre class="code" data-fill-from=".js-code-basic"></pre>
 
-<script type="text/x-example-code" class="js-code-basic">
+      <script type="text/x-example-code" class="js-code-basic">
```
kevin-brown: The code blocks aren't indented because the indentation is copied over to how it is displayed. So if the code is indented in slightly in the markup, the example code in the docs will be also.

It tends to look out of place: http://select2.github.io/select2/#disabled-options

smzchaudary:
Fixed spacing issue.

kevin-brown:
Looks good, thanks for doing this. :+1:

kevin-brown:
Looks good, thanks for doing this. :+1:

PR #2916, opened 2015-01-13
Documentation update

AlaaAttya:
Updated: ajax will not work fine without specifying the 'result' and 'formatResult' functions.

Signed-off-by: Alaa Attya Mohamed <vidooman@gmail.com>


kevin-brown:
Both of these were removed in 4.0, these look like notes for the 3.5 docs (which already have them).

PR #2917, opened 2015-01-13
Added czech translations

solcik:
Aded czech translations (cs.js).


kevin-brown:
Thanks for the translations!

PR #2918, opened 2015-01-13
Adds the dutch translation

stefanvermaas:
As the title of this PR says: it adds the dutch translations.


stefanvermaas:
Saw that was already a dutch PR.

PR #2919, opened 2015-01-13
Small bugfix in german translation

apfelbox:
Just added a missing parameter in the function definition.

I know that the translation isn't complete yet, but this one just caught my eye.


kevin-brown:
Thanks! When I was converting this over I must have missed it.

PR #2922, opened 2015-01-14
Hide element if template returns a false-like value

fallenby:
This allows you to hide elements when using the template% formatting functions.

### DIFF
--- a/src/js/select2/results.js
+++ b/src/js/select2/results.js
@@ -466,7 +466,13 @@ define([
   Results.prototype.template = function (result, container) {
     var template = this.options.get('templateResult');
 
-    container.innerHTML = template(result);
+    var result = template(result);
+
+    if (result) {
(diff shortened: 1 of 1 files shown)


kevin-brown:
This is going to need tests, so we don't lose this in the future.

src/js/select2/results.js
```diff
@@ -466,7 +466,13 @@ define([
   Results.prototype.template = function (result, container) {
     var template = this.options.get('templateResult');
 
-    container.innerHTML = template(result);
+    var result = template(result);
```
kevin-brown: You're going to need to use a different variable here, we're clobbering `result`.

kevin-brown:
Committed in https://github.com/select2/select2/commit/0fe54b6d3be21023248ed6fefd9e51984e03bd64.

PR #2925, opened 2015-01-15
Properly off() the events in the "select2" namespace

gkostov:
The dot, denoting a namespace and not an event name, was missing


kevin-brown:
Thanks, looks like I copied that over incorrectly in one of the commits.

colemanw:
One event was still not getting cleared. This fixes it in the 3.5 branch, haven't tried in 4.x.
https://github.com/select2/select2/pull/3186

PR #2943, opened 2015-01-19
Avoid error on empty search result in MultiSelect. Fixes #2942

### LINKED ISSUE #2942: No search hits causes MultiSelect2 to throw 'showSearch is not defined'
Since the showSearch is defined in SingleSelect and this code is defined in AbstractSelect
```javascript
if (data.results.length === 0 && checkFormatter(opts.formatNoMatches, "formatNoMatches")) {
    render("<li class='select2-no-results'>" + evaluate(opts.formatNoMatches, opts.element, search.val()) + "</li>");
    this.showSearch(search.val());
    return;
}
```
an error will be thrown for MultiSelect when there are no search result hits.
Fixing to: 
```javascript
if(this.showSearch) {
    this.showSearch(search.val());
}
```
should do the trick.

vinntreus:


kevin-brown:

PR #2967, opened 2015-01-26
Fixed return value of AjaxAdapter.prototype.processResults.

loic:


kevin-brown:
A notice was added in https://github.com/select2/select2/commit/5b4e1eab42078313640761d8cc2b86629272493f whenever the results were not in the format of an object with the `results` key. We previously allowed them as just an array, but that changed in https://github.com/select2/select2/commit/05cbaf7bf7951a04f967637521e2f2e1777d1b8c.

So we can't have `processResults` assume that it is always an array, but we can do a check for it (using `$.isArray`) _within_ the function and have it give them a recommendation on how to override it.

loic:
Ah I see now, I didn't see it mentioned in the docs and the example returns an object that doesn't have a `results` key.

PR #2996, opened 2015-02-01
Cleanup composer dependencies and set right files

kartik-v:
Cleanup dependency on any other components to allow installing select2 plugin standalone -- and set all JS and CSS files for access in dist folder


kartik-v:
Probably not the right way to do this -- deleting the PR.

PR #3000, opened 2015-02-04
Fix for MaximumInputLength

lenon:
I18n translations expect a "maximum" argument for inputTooLong message, but MaximumInputLength is passing "minimum".
screen shot 2015-02-04 at 6 54 52 am
This renames "minimum" to "maximum" to fix this problem.


kevin-brown:
Thanks! Looks like we need some tests for this option.

PR #3010, opened 2015-02-06
Fixing errors in FR locale. Added missing pl/pt-BR locales in dist.

yanickrochon:
This PR includes the files generated by Grunt. Only the FR locale was modified, but the generated `select2.*.js` files seemed to be different also. Not sure why. Two locales were missing, too, which are included as well.


kevin-brown:
Thanks for the fixes.

> but the generated select2.*.js files seemed to be different also.

This is from a commit that didn't include the `dist` build.

> Two locales were missing, too, which are included as well.

Same thing, they weren't build in the PR that they were added in.

PR #3012, opened 2015-02-06
fix - add event listener for opera 12

burnb:


kevin-brown:
You don't appear to be unbinding the event [in `destroy`](https://github.com/burnb/select2/blob/307ee39bc22f54bf082592c1f107191444ccf2a9/src/js/select2/core.js#L439-L446). This will cause issues if Select2 is destroyed and the event is still bound.

burnb:
I added unbinding, as you said. And about this fix, i think it not disturbe, but it fix some problem in old Opera.

kevin-brown:
Thanks! Once I get Sauce Labs working with the Select2 tests, I'll try to get Opera 12 checked as well.

PR #3024, opened 2015-02-08
IE7/8 Compatability fixes

th3byrdm4n:
[].indexOf() isn’t supported in Compatability mode for IE10 (specifically IE7/8 Document mode). Utilizing jQuery’s $.inArray() allows for Select2 to work in IE10's default compatibility mode.


th3byrdm4n:
Yes, I know, it's hard to care about being compatible with old IE browsers, but I am required to support it in my current environment, and the fix, at this point, is extremely simple.

kevin-brown:
Thanks for the fixes!

Until we get our tests hooked up to a browser lab, it's going to be difficult to ensure that we don't have any regressions as far as this.

PR #3025, opened 2015-02-09
Add Portuguese translation

lagden:
To avoid the `warn message` when load pt-BR


kevin-brown:
This looks like the same data as `pt-BR`. I'd rather have `pt` contain the translations for what would normally be `pt-PT` instead of `pt-BR`.

Feel free to correct me if I'm wrong though.

lagden:
Hold on!
I will correct the translation to work the right way...

kevin-brown:
Closing this per my comments in https://github.com/select2/select2/pull/3043.

PR #3043, opened 2015-02-13
Create pt-PT.js

faio11:
European Portuguese Translation

### DIFF
--- a/src/js/select2/i18n/pt-PT.js
+++ b/src/js/select2/i18n/pt-PT.js
@@ -1,5 +1,5 @@
 define(function () {
-  //Portuguese
+  // European Portuguese
   return {
     errorLoading: function () {
       return 'Os resultados não puderam ser carregados.';
@@ -16,7 +16,7 @@ define(function () {
     inputTooShort: function (args) {
       var remainingChars = args.minimum - args.input.length;
 
-      var message = 'Digite ' + remainingChars + ' ou mais caracteres';
+      var message = 'Introduza ' + remainingChars + ' ou mais caracteres';
 
       return message;
     },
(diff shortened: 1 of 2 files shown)


kevin-brown:
I'd love it if we could have the file be named just `pt`. This would allow us to have a better, more complete fallback chain for the Portuguese language.

faio11:
So you want just one translation file to work for both pt-PT and pt-BR, is that correct?
I will try to make the text work for both variants of the language, but it will not be perfect because there are quite some differences between the languages.

kevin-brown:
Nope, I'm actually looking for `pt-PT` to be housed in `pt`. This way `pt-PT` will still work (because of fallback chains), and Portuguese will still be under the correct language code (`pt`).

Granted, if `digite` is the only difference then I'm fine with `pt` working in both contexts, it's just not required.

faio11:
Actually it is not, for example "A procurar" [PT] and "Buscando" [BR] (translating "Searching"), are valid in both languages, but in Portugal it is really not acceptable to use the BR version, although I assume both are acceptable in Brazil. In the project I am using Select2 for, I would have to make my own translation if the provided translation was in Brazilian Portuguese.

So my advice here is to keep both variants of the translation.

kevin-brown:
> although I assume both are acceptable in Brazil.

This is more or less what I was assuming.

My hope was to have `pt` house the "European Portuguese" translations and `pt-BR` house the "Brazilian Portuguese" translations. This would be following the scheme of [the Windows Language Code Identifiers](https://msdn.microsoft.com/en-us/library/ms533052%28v=vs.85%29.aspx) ([updated pdf](https://msdn.microsoft.com/en-us/library/cc233965.aspx)), which appears to handle fallback chains similar to Select2.

faio11:
Very well, I have reverted my commit.
Since this is my first contribution I ask you to please double check what I have done and tell me if is there anything else I need to do.
Thank you

kevin-brown:
Once again, thanks for creating the Portuguese translation. :+1: 

I've committed your first two commits as https://github.com/select2/select2/commit/86c23c0f5cec3bd055617a847f118a4a2b793de8 (along with the compiled `dist` file).

PR #3045, opened 2015-02-14
Browser compatibility and styling fixes.

th3byrdm4n:
`[].indexOf(value)` should be avoided in favor of `$.inArray([], value)` to preserve x-browser compatibility. I also noticed in the CSS where there were styles that suggest `word-wrap: nowrap` should be utilized alongside `text-overflow: ellipsis;`

### Side Note:
There's another [].indexOf() in i18n/uk.js ... I tried to define jquery.js in the file, but the compile was failing.


kevin-brown:
This should be easier to test once we get the tests compatible with IE 8. Right now they are all importing jQuery 2, but it's not compatible with IE 8 and will cause all tests to fail.

The tests failed because we can test on individual browsers in pull requests [because of Travis security restrictions](http://docs.travis-ci.com/user/pull-requests/#Security-Restrictions-when-testing-Pull-Requests), which I'm in the process of fixing.

As far as `i18n/uk.js`, the `[].indexOf` can be replaced by a simple greater than/less than check, as they are just looking to make sure the number is within a range.

kevin-brown:
Committed these directly to master and added https://github.com/select2/select2/commit/e6f249f66dbcd7e214085b67af683ddbb0398f70 to fix `i18n/uk.js`.

PR #3051, opened 2015-02-17
Added compatibility with `<input type="text" />` tags

### LINKED ISSUE #3022: Is tagging via <input> elements no longer supported?
Here's an example of this not working. It used to work in 3.5.2:
http://jsbin.com/yonavaqiva/2/edit?html,output

Is this no longer supported or is this just something that needs to be worked out prior to the 4.0.0 release? FWIW, my use-case actually requires a comma-separated input rather than a multiselect.

kevin-brown:
This adds backwards compatibility with old `<input type="text" />` tags.

- [x] Basic decorator
- [x] Tests still need to be added
- [x] A warning needs to be added for hidden inputs
- [x] Documentation

This will close #3022 when it is merged.


wimleers:
The title says `<input />`. Which I read as *any* `<input />`. Wouldn't it be better to explicitly say `<input type="text" />`? Because you (still) don't want to support `<input type="hidden" />`, right? Or does the title imply that things like `<input type="tel" />` are supported?

kevin-brown:
I think this PR is set to go as it currently stands.

> Wouldn't it be better to explicitly say <input type="text" />?

To avoid confusion in the future, I've adjusted the title (and documentation) to point out text inputs specifically.

PR #3058, opened 2015-02-21
Update examples.html

mariohs22:


kevin-brown:
It might be better if we explicitly call out the infinite scrolling support in a separate example, [like we used to](http://select2.github.io/select2/#infinite). I also definitely need to update the [AJAX option documentation](https://select2.github.io/options.html#ajax) to mention it as well.

kevin-brown:
Closing this off as we have renamed the working branch back to master.

PR #3061, opened 2015-02-23
Document folder link updated

janurag:
Document folder link was giving "404 Page Not Found Error" so I updated to "https://select2.github.io/"


kevin-brown:
When we finally switch `select2-ng` over to `master` (any day now), this link will fix itself.

PR #3066, opened 2015-02-24
Place select control outside of & before script element

erikeckhardt:
See http://stackoverflow.com/questions/28705192/select-element-in-html-document-not-displaying/28705288 for why

### DIFF
--- a/docs/examples.html
+++ b/docs/examples.html
@@ -40,17 +40,16 @@ slug: examples
 
       <pre class="code" data-fill-from=".js-code-basic"></pre>
 
-<script type="text/x-example-code" class="js-code-basic">
-$(document).ready(function() {
-  $(".js-example-basic-single").select2();
-});
-
-<select class="js-example-basic-single">
-  <option value="AL">Alabama</option>
-    ...
-  <option value="WY">Wyoming</option>
-</select>
-</script>
+      <select class="js-example-basic-single">
+        <option value="AL">Alabama</option>
+          ...
+        <option value="WY">Wyoming</option>
+      </select>
+      <script type="text/x-example-code" class="js-code-basic">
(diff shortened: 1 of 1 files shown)


kevin-brown:
Just to further expand on the comment I made on the question...

`<script>` elements are only executed by the browser if they are specifically typed to `text/javascript` using the `type` attribute (also the default). In the documentation, we use `<script type="text/x-example-code">` to allow for arbitrary text to be placed in there without having to worry about further escaping it. This is used by [a few](http://handlebarsjs.com/) [JavaScript templating systems](http://icanhazjs.com/), and you can see how it looks yourself [in the generated documentation](https://select2.github.io/examples.html).

You can read more about it [at this Stack Overflow question](http://stackoverflow.com/q/4912586/359284).

erikeckhardt:
Kevin, you're completely right, the alternate script type `text/x-example-code` that you provided was indeed something I wholly overlooked. My apologies.

Please do note that a plain old script tag `<script>` will be treated in browsers exactly as if it had been `<script type="text/javascript">`. Please do try it. I confirmed this before responding to you at the thread.

What I think you meant to say is that type attributes *other than* `text/javascript' (or no type attribute at all) will be ignored by javascript parsers--which I wholeheartedly agree with, and always have.

kevin-brown:
> Please do note that a plain old script tag `<script>` will be treated in browsers exactly as if it had been `<script type="text/javascript">`. Please do try it. I confirmed this before responding to you at the thread.

You're right, this is what I meant by "(also the default)" in my comment here. A lack of an explicit `type` attribute implies `text/javascript`.

In any case, I'm glad we got this cleared up. :)

erikeckhardt:
Yeah, I'm sorry about the furor.

PR #3069, opened 2015-02-25
! Allow HTML5 validation message to be displayed in the right position

PikachuEXE:
For #128


kevin-brown:
This pull request is aimed at 3.5.x, and we don't know when the next release for that will be.

If you're interested in fixing it for 4.0 as well, feel free to make a pull request against `select2-ng`.

PikachuEXE:
I think this should merged anyway

I haven't used `select2-ng` yet, so I got no idea if that is fixed or not (or how to fix)

leods92:
@PikachuEXE Your solution doesn't work in Opera :/

PR #3072, opened 2015-02-25
Fix: TypeError: $ is not a function

elliott-stocks:
I've noticed this error on two different (WordPress) sites. It happens when focusing on the select box.

`TypeError: $ is not a function /select2.js Line 641`

Apologises if it's already been reported, I couldn't find the issue reported.

### DIFF
--- a/src/js/select2/utils.js
+++ b/src/js/select2/utils.js
@@ -202,7 +202,7 @@ define([
     // http://codereview.stackexchange.com/q/13338
     // and was designed to be used with the Sizzle selector engine.
 
-    var $el = $(el);
+    var $el = jQuery(el);
     var overflowX = el.style.overflowX;
     var overflowY = el.style.overflowY;
 


kevin-brown:
jQuery should be imported on the second/third line of the file, and the tests are failing because we are using `noConflict` in tests to ensure that this isn't an issue.

elliott-stocks:
@kevin-brown I'm not really sure I understand how to resolve the issue? How should jQuery be imported?

kevin-brown:
jQuery should be included before Select2 on the page. As long as this happens, Select2 will pick up jQuery [through the shim](https://github.com/select2/select2/blob/4eaa8c5488e17963d46e9a195ae4954e5f3bcf97/src/js/jquery.shim.js) and it should be imported correctly through the rest of Select2.

On the [second line of `utils.js`](https://github.com/select2/select2/blob/b5c833f4348c30dcfcf88bfb0223c8f007aeda12/src/js/select2/utils.js#L2) is where we are actually importing jQuery into the file.

If this isn't actually happening, I'd need to see a fiddle or link to a live site where this is an issue, so I can try to debug the issue.

PR #3091, opened 2015-03-02
Update attachBody.js

qingyuexiao:
Sometimes $dropdown is not positioned right when showing above, for the initial width of $dropdownContainer is 0(when position is absolute, 'width:100%' doesn't work) and it makes the outerHeight of $dropdown incorrect.


kevin-brown:
Can you create a jsfiddle or jsbin where this issue can be reproduced at?

qingyuexiao:
I find that the latest version of select2.full.js runs all right when I tried to create a jsfiddle. So, the problem is fixed, I think.

PR #3099, opened 2015-03-03
added Korean translation.

yoonsy:


src/js/select2/i18n/ko.js
```diff
@@ -0,0 +1,44 @@
...
+
+      return message;
+    },
+    loadingMore: function () {
+      return '불러오는 중…';
+    },
+    maximumSelected: function (args) {
+      var message = '최대 ' + args.maximum + '개까지만 선택 가능합니다.';
+
+      if (args.maximum != 1) {
```
kevin-brown: I don't think this is needed? It doesn't look like there are any advanced pluralization rules to worry about.

src/js/select2/i18n/ko.js
```diff
@@ -0,0 +1,44 @@
...
+  return {
+    errorLoading: function () {
+      return '결과를 불러올 수 없습니다.';
+    },
+    inputTooLong: function (args) {
+      var overChars = args.input.length - args.maximum;
+
+      var message = '너무 깁니다. ' + overChars + ' 글자 지워주세요.';
+
+      if (overChars != 1) {
```
kevin-brown: Doesn't appear to be needed.

src/js/select2/i18n/ko.js
```diff
@@ -0,0 +1,44 @@
+define(function () {
+  // English
```
kevin-brown: If you can swap this out with "Korean", that'd be awesome.

kevin-brown:
Thanks for the translation!

PR #3111, opened 2015-03-05
Add Farsi (Persian) language support.

ahmadazimi:
I've added **Farsi (Persian) language** support to Select2.
Also **.idea** editor config folder and **.DS_Stroe** files are added to `.gitignore` file.
I've add some jshint options to the `i18n/fa.js` file to prevent build errors:
```
/* jshint -W100 */
/*jslint maxlen: 130 */
```


src/js/select2/i18n/fa.js
```diff
@@ -0,0 +1,38 @@
+/* jshint -W100 */
```
kevin-brown: Where was W100 being triggered? I don't see why this error would be triggered in this file, considering it hasn't been triggered in other files.

src/js/select2/i18n/fa.js
```diff
@@ -0,0 +1,38 @@
...
+/*jslint maxlen: 130 */
```
kevin-brown: I'd rather avoid squashing this warning considering it's quite easy to break up strings in JavaScript.

.gitignore
```diff
@@ -1,2 +1,4 @@
 node_modules
 dist/js/i18n/build.txt
+.idea
```
kevin-brown: These should be ignored in your user-level `.gitignore` (the `~/.gitignore` file), not in the project-level one.

src/js/select2/i18n/fa.js
ahmadazimi: We can use **half-space character** in Farsi language to join some related words together. So if I omit this jshint options, `This character may get silently deleted by one or more browsers.` error will be triggered.

```
Running "jshint:code" (jshint) task

   src/js/select2/i18n/fa.js
     26 |      var message = 'شما تنها می‌توانید ' + args.maximum + ' آیتم را انتخاب نمایید';
                                        ^ This character may get silently deleted by one or more browsers.
     31 |      return 'هیچ نتیجه‌ای یافت نشد';
                               ^ This character may get silently deleted by one or more browsers.

>> 2 errors in 81 files
Warning: Task "jshint:code" failed. Use --force to continue.
```

src/js/select2/i18n/fa.js
ahmadazimi: I change the maxlen to 86. This the edge for my strings to be valid via jshint.

.gitignore
ahmadazimi: I've fixed it.

src/js/select2/i18n/fa.js
kevin-brown: Ah, thanks for clearing that up.

kevin-brown:
Thanks for the translations!

ahmadazimi:
You're welcome buddy.

amirazad1:
hi ahmad,
i use select2.js on my project but search not work for some characters like 
ک  or ی
then add fa.js but problem not fix.
any idea?

ahmadazimi:
@amirazad1 
I think the problem is About arabic typo. Maybe you type Arabic ی or ک 
but you have Persian typo in your data.

PR #3122, opened 2015-03-09
Fixes issues with inputData module. When unselecting an item it would previously unselect all items

cristi-badila:


kevin-brown:
[`inputData` has tests](https://github.com/select2/select2/blob/master/tests/data/inputData-tests.js) and while I believe that this change works and improves things, it'd be useful to have a test to prevent future regressions.

cristi-badila:
@kevin-brown sorry about that. I actually was looking for a spec file for this module but I didn't see a "compat" folder in tests and assumed there weren't any tests for it (now that you mentioned it, I took a closer look and found it in the "data" folder). I'll update the pull request with the proper specs.

cristi-badila:
@kevin-brown updated the pull request with the necessary tests and the updated dist files.

tests/data/inputData-tests.js
```diff
@@ -59,6 +59,41 @@ test('test that options can be unselected', function (assert) {
...
+      },
+      {
+        id: 'test2',
+        text: 'Test2',
+        selected: true
+      },
+      {
+        id: 'test3',
+        text: 'Test3',
+        selected: true
```
kevin-brown: While setting `selected` here works because of the internals, I'd rather see the `value` set directly on the `<input />` when doing tests (just in case we remove/rename this).

kevin-brown:
Thanks for the tests, hopefully they will ensure that this doesn't regress in the future.

I've left one last inline comment about pre-selecting options in tests and then I think we should be all set to merge.

cristi-badila:
@kevin-brown updated the pull request as per your recommendations. Hope everything looks good.

kevin-brown:
Thanks a bunch for making the changes. Combined with the tests hopefully this come back as an issue in the future. :)

cristi-badila:
Glad I could help.

PR #3130, opened 2015-03-11
browserify compatibility

pagenoare:
Now installing via:

`npm install pagenoare/select2`

works well:

tested code:
```js
require('Select2');
$('el').select2();
```

Before this I was getting 
```
>> Error: Cannot find module 'jquery.select2' from '/project/path/node_modules/Select2/dist/js'
```


JohanLopes:
:+1:

pagenoare:
I can rebase it against master if u like to :-) (since it's not merge-able anymore)

kevin-brown:
Does browserify require that the JS file name matches the module name?

I'm not familiar with how it works, so I don't quite understand why the module renames happened (or why there is a casing difference).

pagenoare:
TBH it was the fastest solution I've found (I needed it badly). The best one, would be to make  `jquery.select2` recognizable to browserify (probably by doing some aliases, or shims or whatever it's called). 

I'll look at this after my daily job (in around 8hours) and let you know what I've found out.

Regards!

pagenoare:
Just checked that out with the latest version.

I think https://github.com/select2/select2/commit/d1e8a6fdad700deba721aaf83c912e09db80eb5d fixes it. 

Check out the demo I've made: https://github.com/pagenoare/select2-browserify-example

Thanks a lot!

Issue #3137, opened 2015-03-12
4.0.0 rc1 : translation not working


qcaron:
Hi guys,

I downloaded the first release candidate for version 4 but I could not manage to make translation working. I tried the following for both French and Spanish without success:

First, I import javascript files as follow:
```
<script src="path/to/select2/dist/js/select2.min.js"></script>
<script src="path/to/select2/dist/js/i18n/fr.js"></script>
```

Then I apply options on my select as follow:

```javascript
$("#my-select").select2({
    maximumSelectionLength: 3,
    language: "fr",
});
```

My console shows no error. The same goes using select2.js instead of select2.min.js.

Thanks!
Quentin

kevin-brown:
I cannot reproduce the issue you are seeing.

Here is the test case I used: http://jsbin.com/vononirehu/1/edit

qcaron:
Okay maybe I should have added a bit more details here. Anyway, your explanation is a bit elusive to me.

Please check out that JS Bin and let me know if you see some French when you try to select a second option: http://jsbin.com/bugawebiqa/1/

Cheers

kevin-brown:
The issue is that `maximumSelected` [is misspelled in the French locale](https://github.com/select2/select2/blob/e3d7cd2d6952fc4412e953f2b749ede316e70dc5/src/js/select2/i18n/fr.js#L29).

qcaron:
You're right! There are 15 languages with the same spelling error (according to GitHub search results). I will fork and ask for pull request.

Thanks for your time!

PR #3145, opened 2015-03-13
Updated 15 of the translation files so 'maximumSelected' is used instead of 'maximumSelection'

qcaron:
Hi there !

I have updated code and compiled with Grunt according to issue https://github.com/select2/select2/issues/3137.
Please let me know if anything is wrong so I can change the code.

Cheers!
Quentin


kevin-brown:
I did a quick check and everything looks good, thanks!

PR #3156, opened 2015-03-14
Dynamic jsonpCallback

nemanjan00:
Can be used if jsonpCallback change depending on term.

### DIFF
--- a/src/js/select2/data/ajax.js
+++ b/src/js/select2/data/ajax.js
@@ -55,6 +55,10 @@ define([
     if (typeof options.url === 'function') {
       options.url = options.url(params);
     }
+    
+    if (typeof options.jsonpCallback === 'function') {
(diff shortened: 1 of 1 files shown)


nemanjan00:
Hello,
Can you tell my why this wasn't merged? 
Thank you

kevin-brown:
This was not merged because it is not a commonly requested feature, and we are trying to cut back on the number of special cases which are handled within the core. Note that you can still implement this fairly easily using a custom `ajax.transport` method, as described in the following two questions on Stack Overflow.

- [Select2 use a dynamic Ajax URL on call](http://stackoverflow.com/q/30373451/359284)
- [Select2: searching for results from remote data sources using AJAX requests](http://stackoverflow.com/q/29189050/359284)

`ajax.url` and `ajax.data` are allowed to be functions for backwards compatibility reasons.

PR #3180, opened 2015-03-20
Remove jQuery as a jsHint global

colemanw:
Let's remove jQuery as a jsHint global so attempts to access it out-of-scope will be caught by linting.


kevin-brown:

PR #3181, opened 2015-03-21
fix jquery.mousewheel require path

braznaavtrav:
Previously, when compiling with browserify, the require call for the `jquery.mousewheel` module was throwing `Error: Cannot find module 'jquery.mousewheel'`. I've added a relative path to grab jquery.mousewheel from the vendor folder.


kevin-brown:
Was this all that was needed to get it working with browserify? Also, were you using the `dist` build (and that was triggering it) or were you building it directly from `src`?

If this just happens with the `dist` builds, I'd prefer a shim (like we do for jQuery). If it happens from `src` builds, is it possible to tell browserify to shim the module with a blank ([like `:empty`](http://requirejs.org/docs/optimization.html#empty) for r.js)?

braznaavtrav:
Yes, this got it working with browserify & npm (I installed with `npm install --save braznaavtrav/select2`) and I'm currently using it in a project. The build I am using is dist, which is what is set as `main` in package.json.

The problem this solves is in browserify (as well as node.js environments) if you require a module without providing a path, e.g. `require('jquery.mousewheel')`, browserify will by default look in the node_modules folder. Since `jquery.mousewheel` doesn't get installed as an npm dependency, it can't be required without providing a relative path.

kevin-brown:
I see, that makes sense (somewhat).

I'd rather avoid having a relative path for jQuery mousewheel, especially one that breaks from the main directory (`..` being the concern). But I _think_ there might be a few ways to get around it which would also improve how we have that line set up.

1. Require the `require` module in `jquery.select2` and use that instead of the global one. This is what we do [in the `select2/defaults` module](https://github.com/select2/select2/blob/9318980651e5bf471c412834139a03bf5de54f94/src/js/select2/defaults.js#L3) for optionally loading dependencies, and that appears to work well. My guess is that this has something to do with `require` being a local variable in the scope, so it gets ignored.
2. Shim the `jquery.mousewheel` dependency [like we do for `jquery`](https://github.com/select2/select2/blob/master/src/js/jquery.shim.js). This would probably be better in the long run anyway, for cases where jQuery mousewheel is used outside of the full build.

I have a feeling we're going to end up needing to do both in order to get it working correctly, but I have no clue how to test it properly to see if it fixes the browserify issue.

braznaavtrav:
Maybe I'm not following you totally on those solutions, but at first glance it doesn't seem like either of them would work. I'm not using requirejs in my project, I'm building with browserify, solution 1 seems like a AMD-only approach. Solution 2 looks like it requires a global instance of jquery and jquery.mousewheel, which pretty much defeats the purpose of using any module system. If you don't want to require jquery.mousewheel by providing a path, I'm fairly certain that the only solution would be adding it (as well as jquery) to the list of dependencies in the package.json file and running `npm install`.

kevin-brown:
Fixed in https://github.com/select2/select2/commit/da1ec4e02865f3b23d7634eba509e3438d4ab874.

MojoJojo:
I see this has been marked as closed, but I ran into the same issue with the latest download from the website (4.0.0-rc.2). Not sure if the fix has made its way into the release candidate. I am fairly new to AMD/JS scene, so please excuse me if the following doesn't make any sense.
While researching the issue independently I found that that jquery mousewheel plugin is also available as browserify module from the original author (https://github.com/jquery/jquery-mousewheel). However, the module is exported as 'jquery-mousewheel' instead of 'jquery.mousewheel' as require'ed in select2.js.

I first installed the jquery-mousewheel plugin via npm in my project and then changed the 'jquery.mousewheel' to 'jquery-mousewheel' in select2.js and it worked!

kevin-brown:
It's on master (upcoming 4.0.0-rc.3).

PR #3192, opened 2015-03-24
fix error : "Uncaught ReferenceError: require is not defined"

dbwhddn10:


kevin-brown:
This should also be fixed in [`jquery.select2.js`](https://github.com/select2/select2/blob/9318980651e5bf471c412834139a03bf5de54f94/src/js/jquery.select2.js).

dbwhddn10:
@kevin-brown this is fatal error: see https://jsfiddle.net/dbwhddn10/7zek638x/

kevin-brown:
Right, I understand that hidden inputs are broken right now because of this. I was just pointing out that we've also got an issue when loading in jquery.mousewheel, but it's being silenced and it might be useful to also get that fixed.

kevin-brown:
Committed in https://github.com/select2/select2/commit/da1ec4e02865f3b23d7634eba509e3438d4ab874.

PR #3207, opened 2015-03-30
[3.5] Update PL translation

piotrantosik:
Add missing formatAjaxError translation


kevin-brown:
In Select2 4.0 we use a [slightly different translation](https://github.com/select2/select2/blob/cc1eefb78f88ce0ca17c6059fb940c0fda84a82e/src/js/select2/i18n/pl.js#L18), but both of them _appear_ to mean the same thing.

kevin-brown:

Issue #2941, opened 2015-01-19
composer.json without components/jquery


remdan:
Why this requirement? I don't like it then I work not with "components".
With that I have the "components" in my projekt but it's not used, then I include jquery on a other way...

"require": {
    "robloach/component-installer": "*",
    "components/jquery": ">=1.7.2"
  }

kevin-brown:
I'm considering doing this for `bower.json` and `composer.json` which both specify their dependencies.

PR #3218, opened 2015-03-31
add jquery dependency

sankargorthi:
this makes sure that `component(1)` environments can use the jquery dependency needed by select2

### DIFF
--- a/component.json
+++ b/component.json
@@ -7,6 +7,9 @@
   "keywords": [
     "jquery"
   ],
+  "dependencies": {
+    "components/jquery": "^2.1.3"
+  },
   "main": "dist/js/select2.js",
   "styles": [
     "dist/css/select2.css"


sankargorthi:
Note, this will only work with `component(1)` `v1.x`

component.json
```diff
@@ -7,6 +7,9 @@
   "keywords": [
     "jquery"
   ],
+  "dependencies": {
+    "components/jquery": "^2.1.3"
```
sankargorthi: the assumption is that the version will be maintained for future upgrades.

kevin-brown:
This used to be required, but we removed it in https://github.com/select2/select2/commit/695123a1d2d50936f56380953937250cc9754ef6 because of https://github.com/select2/select2/issues/2941.

sankargorthi:
I suppose but component(1) needs the dependency declared in the manifest `component.json` or it will be unable to find the jquery dependency.

sankargorthi:
@timaschew is my comment valid for the current component version?

Issue #3065, opened 2015-02-24
Render native select out of screen or with clip: rect(0 0 0 0), height: 0


ova2:
I realized a probelm with Select2 when using with some web framworks having client-side validation. Some frameworks use this selector to figure out elements to be validated while form submission: ```input:visible:enabled:not(:button)[name]```. Saidly, but the native select element in Select2 has ```display: none``` and can not be validated. Can you render the native select out of screen please or apply this jQuery UI style class (normally applied to the surrounded div)?

```css
.ui-helper-hidden-accessible {
    border: 0;
    clip: rect(0 0 0 0);
    height: 0px;
    margin: -1px;
    overflow: hidden;
    padding: 0;
    position: absolute;
    width: 0px;
}
```

I quess, you should also set ```aria-hidden``` attr. to make it invisible for screen readers.

penihel:
I have many problems with this. It would be wonderful if the version 4 the Select2 to become compatible with JQuery Validation Plugin.

Today when ajax he does not properly validate

kevin-brown:
Pull requests are always appreciated. This one is going to need a pair of jsfiddles for demonstration purposes.

penihel:
https://jsfiddle.net/pbrjetqh/
This is a example of the problem

ova2:
Thank you!

elanhami:
This caused some trouble for me on Chrome when working with a large dataset (>7000 elements).
Since the native select is still considered part of the render tree, selecting an option in select2 has a significant delay because displaying/removing the dropdown causes a page reflow. The browser applies the reflow to all the options in the DOM, whereas with `display:none` this is not the case.

Adding this here to save others time I spent finding the cause...

PR #3219, opened 2015-04-01
Render native select out of screen or with clip

penihel:
https://github.com/select2/select2/issues/3065#issuecomment-85731663

### DIFF
--- a/src/js/select2/core.js
+++ b/src/js/select2/core.js
@@ -78,7 +78,8 @@ define([
 
     // Hide the original select
     $element.addClass('select2-hidden-accessible');
-
+	$element.attr('aria-hidden', 'true');
+	
     // Synchronize any monitored attributes
     this._syncAttributes();
 
@@ -490,6 +491,7 @@ define([
     this.$element.attr('tabindex', this.$element.data('old-tabindex'));
 
     this.$element.removeClass('select2-hidden-accessible');
+	this.$element.attr('aria-hidden', 'false');
     this.$element.removeData('select2');
 
     this.dataAdapter.destroy();
--- a/src/js/select2/core.js
+++ b/src/js/select2/core.js
@@ -77,7 +77,7 @@ define([
     });
 
     // Hide the original select
-    $element.hide();
+    $element.addClass('select2-hidden-accessible');
 
     // Synchronize any monitored attributes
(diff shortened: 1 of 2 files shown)


src/scss/core.scss
```diff
@@ -33,5 +33,16 @@
   filter: alpha(opacity=0);
 }
 
+.select2-hidden-accessible {
+    border: 0;
+    clip: rect(0 0 0 0);
+    /*height: 0px;*/
```
penihel: comment this line to $('select').is(':visible') return true

kevin-brown: jQuery UI uses `height: 1px` and `width: 1px` to work around this.

https://github.com/jquery/jquery-ui/blob/0de27b0609e8f6f9751ab7cce28492e18206d86d/themes/base/core.css#L17-L26

src/js/select2/core.js
```diff
@@ -77,7 +77,7 @@ define([
     });
 
     // Hide the original select
-    $element.hide();
+    $element.addClass('select2-hidden-accessible');
```
kevin-brown: It might also be useful to do `$element.attr('aria-hidden', 'true')` here, so screen readers know to not focus on the element.

src/js/select2/core.js
```diff
@@ -489,7 +489,7 @@ define([
     this.$element.off('.select2');
     this.$element.attr('tabindex', this.$element.data('old-tabindex'));
 
-    this.$element.show();
+    this.$element.removeClass('select2-hidden-accessible');
```
kevin-brown: And `$element.attr('aria-hidden', 'false')` here.

src/scss/core.scss
penihel: that's it. I will change is the commit here

src/js/select2/core.js
penihel: that's it. I will change is the commit here

src/js/select2/core.js
penihel: that's it. I will change is the commit here

kevin-brown:
Committed in https://github.com/select2/select2/commit/c24293f2ba3d1de3b2dd1421177d3d91b05be887.

PR #3220, opened 2015-04-01
Change touchstart and mousedown to prevent unexpected swipe opening

Frumcheg:
I try to fix trouble with scrolling on iOS: if I start dragging by touch for scrolling the page, the select will open if my first touch was over the select box.

### DIFF
--- a/select2.js
+++ b/select2.js
@@ -2250,7 +2250,7 @@ the specific language governing permissions and limitations under the Apache Lic
                 }
             }));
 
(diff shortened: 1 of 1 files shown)


kevin-brown:
Closing this off because while this is inconvenient, the behavior is expected in 3.5.

In 4.0.0, we [no longer listen](https://github.com/select2/select2/blob/cc1eefb78f88ce0ca17c6059fb940c0fda84a82e/src/js/select2/selection/single.js#L38) to the `touchstart` event.

PR #3276, opened 2015-04-22
add japanese translation

chungth:


kevin-brown:
[Tests are failing](https://travis-ci.org/select2/select2/builds/59548111) because you appear to have used [`dist/js/i18n/en.js`](https://github.com/select2/select2/blob/master/dist/js/i18n/en.js) instead of [`src/js/select2/i18n/en.js`](https://github.com/select2/select2/blob/master/src/js/select2/i18n/en.js).

That should also help to fix the JSHint errors that are being triggered.

chungth:
@kevin-brown   sorry for late reply. I fixed it

kevin-brown:
I forgot to close this off when https://github.com/select2/select2/pull/3477 was merged. That pull request _also_ added Japanese translations.

If there are any discrepancies between the two translations, feel free to note them in the other pull request.

PR #3280, opened 2015-04-24
feature: select2 don't search by parent when inited with custom data

kovalenko-sb:
When select2 is populated from the hierarchical array, search is working on the parent elements.

For example,
```
var data = [
  {
    text : 'parent 1',
    children: [
      { id: 1, text: 'child 1' },
      { id: 1, text: 'child 1' },
    ]
];
$('#e').select2({ data: data });
```
select2 initialized with this code returns `parent 1` as a search result if you search by `par`.

Now, parents don't have `id` field, so they wouldn't be searched by.

### DIFF
--- a/select2.js
+++ b/select2.js
@@ -518,7 +518,7 @@ the specific language governing permissions and limitations under the Apache Lic
                     }
                     group.children=[];
                     $(datum.children).each2(function(i, childDatum) { process(childDatum, group.children); });
-                    if (group.children.length || query.matcher(t, text(group), datum)) {
+                    if (group.children.length || (query.matcher(t, text(group), datum) && datum.id !== undefined)) {
                         collection.push(group);
                     }
                 } else {


kevin-brown:
Select2 provides a `matcher` option that can be overridden if you want to change how options are matched.

kovalenko-sb:
ok, I'll try to use matcher option, but I have a question.

When I have html

```
<select id="e1">
	<optgroup label="parent 1">
	   <option value="1">child 1</option>
	   <option value="2">child 2</option>
	</optgroup>
	...
</select>
```

and init select2

```
$(document).ready(function() { $("#e1").select2(); });
```

If I search by "par", I dont have "parent 1" in result.

But if I init select2 like this

```
var data = [
  {
    text : 'parent 1',
    children: [
      { id: 1, text: 'child 1' },
      { id: 1, text: 'child 1' },
    ]
];
$('#e').select2({ data: data });
```

If I search by "par", I have "parent 1" in result.


Is it correct? Select2 works differently on same data, if it initialized on differently way.

PR #3290, opened 2015-04-27
Improving Foundation Zurb compatibility

Sopamo:
As Foundation (http://foundation.zurb.com/) has a box shadow on it's input fields we need to remove them in the inline input field.


kevin-brown:
> we need to remove them in the inline input field.

Now I understand where the issue is coming from. Thanks for the fix.

PR #3291, opened 2015-04-27
Swap inputTooLong and inputTooShort error messages

aitor:
"Engada"  is galician for "to add" and therefore should be the base text used for the inputTooShort method, asking the use to add more chars. It seems to be mistakenly defined exactly the opposite it should be (as inputTooLong was containing the right text).


kevin-brown:
Good catch, thanks!

aitor:
My pleasure :).

PR #3301, opened 2015-04-30
Add `!important` to `.select2-hidden-accessible`.

fk:
`.select2-hidden-accessible` is the CSS class that is added to the original `<select>` element (that Select2 is initialized on) to hide it.
In cases where the original `<select>` has other style declarations assigned which conflict with those of `.select2-hidden-accessible`, the latter styles may be overridden, possibly resulting in the original `<select>` element not being properly hidden.

Similar to what https://github.com/select2/select2/pull/1549 did for Select2 v3's CSS, this adds the `!important` rule to all style declarations for `.select2-hidden-accessible` to make it exponentially harder to override them.


kevin-brown:
Thanks!

PR #3305, opened 2015-05-01
Fix undefined function error

colemanw:
PR #2828 caused an "undefined is not a function" error when no results are found in a multiselect, because it attempts to call a function that only exists for single selects. This fixes it.


colemanw:
@kevin-brown looks like this didn't make it into today's release :(
Would you like me to rebase it to fix the merge conflicts?

kevin-brown:
Looks like this was fixed in 3.5.3 with https://github.com/select2/select2/commit/95e6f806198ba589aee3492214aa514896ff6b7e.

colemanw:
Cool thanks for checking :)

Issue #3304, opened 2015-05-01
iOS Safari "Unexpected CSS token: :" 


RedoAoura:
Hey guys,

I do get 4 "Unexpected CSS token: :" errors on iOS Safari browser. I found the bug It is caused by this kind code snippets where "filter: " expect a string but it gets "progid: ". 

<code>
filter: progid:DXImageTransform.Microsoft.gradient(startColorstr='#ffffff', endColorstr='#eeeeee', GradientType=0); }
</code>

Please dubbel quote the value of "filter: " like this

<code>
filter: "progid:DXImageTransform.Microsoft.gradient(startColorstr='#ffffff', endColorstr='#eeeeee', GradientType=0)"; }
</code>

Thanks!

fk:
While I can reproduce this in the desktop version of Safari for both the minified and the unminified `select2.css`, it's worth noting that at least on desktop Safari these are only warnings, not errors:

![image](https://cloud.githubusercontent.com/assets/21834/7442496/2f5b0840-f114-11e4-80a4-6c47699d7312.png)

@RedoAoura Can you confirm that these are only warnings on iOS Safari, too? They should be only visible in Safari's "Web Inspector" console – I guess you discovered them when you were debugging iOS Safari via desktop Safari's "Web Inspector" and a device (iPad, iPhone) hooked up to a Mac?

The CSS raising the warnings is generated via a [Sass mixin](https://github.com/select2/select2/blob/master/src/scss/mixins/_gradients.scss) Select2 [borrowed from Bootstrap](https://github.com/twbs/bootstrap-sass/blob/a416cb86a3ddc6906892b2c0472deea306989fea/vendor/assets/stylesheets/bootstrap/mixins/_gradients.scss) – [here's what the Bootstrap documentation has to say about these kind of validation errors](http://getbootstrap.com/getting-started/#support-validators) (TL;DR: don't worry).

Related: https://github.com/FortAwesome/Font-Awesome/issues/5043.

I vote to close this issue – although we could look into invalid and valid syntaxes (mentioned e.g. [here](https://css-tricks.com/css3-gradients/)), I'd rather trust Bootstrap on this.

[...]

PR #3307, opened 2015-05-02
Update Sass vertical gradient mixin to latest version from Bootstrap Sass

fk:
As described in https://github.com/select2/select2/issues/3304#issuecomment-98395238, this updates the mixin to the latest stable version from Bootstrap Sass which adds Sass' http://sass-lang.com/documentation/Sass/Script/Functions.html#ie_hex_str-instance_method to ensure that colors are converted "into the format understood by IE filters".

The second commit also brings indention in line with both Select2's and Bootstrap Sass' style.


kevin-brown:
Thanks!

Issue #2944, opened 2015-01-19
Error on using infinite scroll


pujadeora:
I am using a select2 dropdown with infinite scroll for remote data that comes with paging mechanism. Some of the ajax calls seem  to be getting cancelled. I am not sure whether they are getting aborted by this plugin or due to some other network error. 

I have used the same ajax request at various other places in my application (without select2) and it doesn't get cancelled. I see that this plugin cancels pending requests when the user modifies the query term. Is there any chance that some of these paging calls are getting cancelled by select2?

Also, when these calls get cancelled, the select2 code encounters an error in the populateResults function because the results array is empty and this function is trying to read results.length. Can you help me to handle this error?

Thank you in advance.

kevin-brown:
What version of Select2 are you using? It looks like 3.5.2 based on your comment about `populateResults`, but I'm not sure.

> Also, when these calls get cancelled, the select2 code encounters an error in the populateResults function because the results array is empty and this function is trying to read results.length.

Can you include the traceback that is generated for this error? I haven't seen it, so hopefully the traceback can help up determine what the issue is?

vincentjames501:
Oddly, I can't reproduce this on my mac, but we are able to reproduce this on one of our QA linux machines. Working on getting more information...

```
Uncaught TypeError: Cannot read property 'length' of undefined
select2.min.js:947 
clazz.prepareOpts.$.extend.populateResults.populateselect2.min.js:994 
clazz.prepareOpts.$.extend.populateResultsselect2.min.js:1660 (anonymous function)
select2.min.js:681 
clazz.bindselect2.min.js:456 
window.setTimeout.$.extend.errorjquery.js:3094 
n.Callbacks.jjquery.js:3206 
n.Callbacks.k.fireWithjquery.js:8261 
xjquery.js:8010 
n.extend.ajax.v.abortselect2.min.js:428 (anonymous function)
```

vincentjames501:
Looks like it issues ```loadMoreIfNeeded``` two times back to back before the first request has finished. The first request gets aborted and causes the js exception as mentioned in my previous post. We should handle this gracefully by simply adding the following:

```javascript
populateResults: function(container, results, query) {
                    var populate, id=this.opts.id, liveRegion=this.liveRegion;

                    populate=function(results, container, depth) {

                        var i, l, result, selectable, disabled, compound, node, label, innerContainer, formatted;

                        results = results || []; //This line is added
                        results = opts.sortResults(results, container, query);
```

[...]

PR #3323, opened 2015-05-07
Fix for undefined results

sam159:
This fixes an issue where results may be undefined during loading the next page of results

See: https://github.com/select2/select2/issues/2944


kevin-brown:

Issue #3222, opened 2015-04-02
Add container to templateResult


Knoxvillekm:
Container need for set classes, styles, etc to option.

example:
```js
select2({
  templateResult: function(result, container) {
    if (!result.id) { return result.text; }
    $(container).addClass(result.resultClass);
    return result.text;
  }
});
```

code:
```js
  Results.prototype.template = function (result, container) {
    ...
    var content = template(result, container);
    ...
  };
```

stretch4x4:
Yep I could use this too, going to work on a pull request for this item

Igloczek:
Thumbs up, I need this feature too, because Select2 doesn't work well with FastClick (need extra "needsclick" class)

hubsonx:
I have applied Select2 to my picklists in Salesforce Lightning components and everything works fine on the desktop and on my mobile phone (android 5), but on the Ipad mini nothing happens when I click on the select box.

I have discovered that lightning components use fastclick natively and I couldn't find the resolution for this in the boards. 

I have tries to apply the 'needsclick' class to my selects by it has not helped me.
Any ideas how to fix this?

superKalo:
@hubsonx here's how I managed to finally fix it.

I have Select2 4.0.1 and FastClick 1.0.6 (both latest by the time I am writing this)

```javascript
var $customSelects = $('select');

$customSelects.select2({
    /**
     * Adds needsclick class to all DOM elements in the Select2 results list
     * so they can be accessible on iOS mobile when FastClick is initiated too.
     */
    templateResult: function(result, container) {
        if (!result.id) {
            return result.text;
        }
        container.className += ' needsclick';
        return result.text;
    }
});

/**
 * Additional to tweaking the templateResult option in Select2,
 * add needsclick class to all DOM elements in the Select2 container,
 * so they can be accessible on iOS mobile when FastClick is initiated too.
 *
 * More info about needsclick:
 * https://github.com/ftlabs/fastclick#ignore-certain-elements-with-needsclick
 *
 */
$customSelects.each(function(index, el){
    $(el).data('select2').$container.find('*').addClass('needsclick');
});
```

Here's [jsfiddle with an example](https://jsfiddle.net/superKalo/oj0sLr1n/), so you can test and make sure my code works ;-)

PS: Thanks @Knoxvillekm for your initial suggestion that fixes the dropdown list elements. I only changed it to apply class via Vanilla JS, instead of calling jQuery for that (this should perform faster).

nobruds:
[...]

PR #3324, opened 2015-05-07
Adding container support issue #3222

stretch4x4:
Fixes issue #3222
Now templateSelection can have access to the parent container so that
classes can be added for styling

### DIFF
--- a/tests/selection/multiple-tests.js
+++ b/tests/selection/multiple-tests.js
@@ -33,6 +33,41 @@ test('display uses templateSelection', function (assert) {
   assert.equal(out, 'test');
 });
 
+test('templateSelection can addClass', function (assert) {
+  var called = false, found = false;
+
+  var templateOptions = new Options({
+    templateSelection: function (data, container) {
+      called = true;
+      container.addClass('testclass');
+      return data.text;
+    }
+  });
+
+  var selection = new MultipleSelection(
+    $('#qunit-fixture .multiple'),
+    templateOptions
+  );
+
+  var $container = selection.selectionContainer();
+  
+  var out = selection.display({
+    text: 'test'
+  }, $container);
+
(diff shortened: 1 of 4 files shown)


tests/selection/multiple-tests.js
```diff
@@ -33,6 +33,41 @@ test('display uses templateSelection', function (assert) {
...
+    templateOptions
+  );
+
+  var $container = selection.selectionContainer();
+  
+  var out = selection.display({
+    text: 'test'
+  }, $container);
+
+  for (var i = 0; i < $container[0].classList.length; i += 1) {
```
kevin-brown: You should be able to just use `$container[0].classList.contains('testclass')` here. Though that isn't supported in IE8, I'd just use `$container.hasClass('testclass')` honestly.

stretch4x4: I did think that was a bit average but was struggling to think of a better way last night.
Thanks for the feedback, also helped me pick up that I had forgotten to rename the variable so it wasn't actually being tested anyway :(

kevin-brown:
The only thing holding me back on merging this is that the single and multiple selection templates now have different parameters. If we can integrate this in with `SingleSelection`, that would be perfect.

stretch4x4:
I didn't really think it was applicable to single select due to it not really using the tag template and it seemed to fall back ok if you didn't use the container but it shouldn't be hard to add to the single selection and no doubt somebody will find a use.
Will see if I can find 5 minutes to solve that.

stretch4x4:
Finally got around to adding the single select support, hopefully that is ready to merge now

kevin-brown:
Looks good, thanks for adding support for the single select as well.

stretch4x4:
No worries

Issue #70, opened 2012-05-28
Open dropdown on mousedown instead of click and select the option on mouseup istead of click


aziz:
Inorder to mimic native selects behavior, **_Open dropdown on mousedown instead of click and select the option on mouseup instead of click.**_ 
I've made a patch for chosen long ago (https://github.com/harvesthq/chosen/pull/165) to fix this. 
I think it's crucial to have the same behavior as native components. It's even more crucial in multi selects where you want to select multiple items. In my old patched version each selection needed only one mouse click. Press Mouse button and hold over the select to open the dropdown, then hover over the option you want and release the mouse button.

PR #3369, opened 2015-05-18
Change listener from 'mouseup' to 'click' on dropdown results

cerlestes:
This prevents the results dropdown from closing immediately again when a result was positioned directly under the cursor, for example due to CSS styles that position the dropdown with its results directly above the select2 container.

### DIFF
--- a/src/js/select2/results.js
+++ b/src/js/select2/results.js
@@ -292,7 +292,7 @@ define([
         return;
       }
 
-      $highlighted.trigger('mouseup');
+      $highlighted.trigger('click');
     });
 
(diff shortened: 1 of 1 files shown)


kevin-brown:
As mentioned [in the ticket that added this](https://github.com/select2/select2/issues/70), the select box provided natively by most browsers has support for opening, selecting an option, and closing the dropdown in a single click action (press down, move, release).

For those who need to reposition Select2 in a way that prevents them from being able to support this, we recommend overriding these event handlers on your own so you can control how the selection flow works.

PR #3370, opened 2015-05-18
Remove unnecessary _resolveWidth() method

cerlestes:
The way this method is used at the moment completely destroys the responsiveness of Select2. It's not really needed anyway, unless I've overlooked something. A fixed container width is achievable using CSS without breaking responsiveness. As anticipated it's working just fine without this in all major browsers.


cerlestes:
See #3371 for a better way to fix this.

Issue #3278, opened 2015-04-23
Issues with responsive width


fredmalone:
Unless I am missing something (I apologize if I have), select2 is not fully responsive.  Look at this code:

```
    <div class="col-sm-6">
        <div class="form-group">
            <label>Submit to</label>
            <select class="form-control" id="SubmitQueue" name="SubmitQueue">
                <option value="1">Subsequent Civil</option>
                <option value="2">Subsequent Criminal</option>
                <option value="3">Judge Order - Civil</option>
                <option value="4">Review Sign</option>
                <option value="5">Subsequent Domestic Family</option>
                <option value="6">Judge Order - Criminal</option>
            </select>
        </div>
    </div>
```

The original select has 100% width from the form-control class.  Select2 is creating a fixed width container.  Thus, when I resize the window the select2 container does not resize as it should since the col-sm-6 class is a percentage.  

So did I miss something in my initialization or is this an issue?

EDIT:  OK. I looked throught the code and I noticed that the _resolveWidth doesn't even attempt to get a percentage width from CSS and only works with an inline style.    I can live with that.  How did it work in 3.5.2 without the inline style?

eobeda:
What I was seeing is that the 100% width is being converted into 100px. Hence the loss in responsiveness.  Are your element initially hidden ?  I was using the following to work-around for some of the issues. I never got the place-holder text working on long strings. Manually clearing the selection worked, but I never got a programmatic solution.
 //// Fix a select2 ver 4.x issue with widths on initially hidden elements
        if ($(element).not(':visible')) {
            $('.select2-search__field').width("100%");  // The placeholder text needs help as well.

fredmalone:
No, they are not initially hidden.  I will look at that options.width to see if it works for me.

alexcroox:
I've found that if you have select 2 in a hidden div, when you later reveal that div the width is wrong, is there a method to re-initialise the positioning?

alexcroox:
Temporary workaround is to specify width in the init() I can get away with this as all mine happen to work fine 100% width.

`.select2({ width: '100%' });`

hhubik:
Also having this issue.  Have a select2 (v4) in a Bootstrap v3 accordion panel that is closed on page load.  When I expand the panel for the first time, the width of the placeholder text is set to 100px even though the select2 element is much wider based on the markup and the browser window size.  After I enter one or two values and then delete them, the placeholder displays correctly.

My HTML markup is 
[...]

PR #3371, opened 2015-05-18
Change default value for 'width' option to NULL

cerlestes:
Select2 should not set a specific width by default, as this breaks the responsiveness of the plugin. Setting the default value to NULL will make Select2.prototype._resolveWidth() return NULL as well, which in turn prevents Select2.prototype._placeContainer() from applying a CSS-width to the container, which would make it unresponsive.

### DIFF
--- a/src/js/select2/defaults.js
+++ b/src/js/select2/defaults.js
@@ -375,7 +375,7 @@ define([
         return selection.text;
       },
       theme: 'default',
-      width: 'resolve'
+      width: null
     };
   };
 


kevin-brown:
This is a breaking change, and I'm curious what the impact is for those who are not overriding the CSS and are using Select2 with an inline select?

bcdude87:
Would a default width of '100%' in the element calculation also solve this issue without breaking other functionality?  I certainly don't know all of the use cases, but NULL was not a good solution for me in a number of places I tried this, 100% simply filled the parent element (as I style my div containers, not the input elements directly) and scaled properly in responsive designs.  This could still be overridden when initializing or styling directly if the user then specifies NULL.

cerlestes:
A width of 100% is almost never a good idea, as it interferes with paddings and borders in various scenarios, including those of the parent element. width:auto; is the only way to go, IMHO.

I just don't think that the user should have to specify that a plugin should behave 'normally'/as expected by overriding the config; instead the plugin should follow the default behaviour of block-elements, with the option for the user to override this when wanted/needed.

bcdude87:
I understand my use cases account for about 10 out of (millions?), but I would like to offer another 2 cents... In each of the projects I've implemented Select2, I override width to 100% because the template (specifically, the parent DIV) is already styled to fit where it needs to be.  I've spent too much time around the template to ever want to consider targeting another plugin directly.  Consider placing this into Pure (purecss.io) with a responsive grid.  I've not run into any issues with padding and borders and it looks great on my phone and my massive desktop display.

But as said, the ability to override when wanted is there, so while I do agree with NOT defining a static width in px, it doesn't much matter in my case what the end result is.

kevin-brown:
We're looking to find a better way to handle responsive designs than requiring the width to be specific inline, but defaulting to `null` _defininitely is not the solution_. If you have any solutions for detecting a percentage width by default, [we're all ears](https://github.com/select2/select2/issues/3278).

The only rule is that it can't be a significantly breaking change. Select2 suddenly becoming responsive instead of having a fixed width is not a breaking change, it's a bug that has been fixed. When Select2 no longer sets a width in the first place, then it becomes a breaking change.

---

I decided to do some testing with the `width` options right now as they compare to the normal width of the `<select>`:

_**Warning:** huge full-page screenshots of [the examples page](https://select2.github.io/examples.html)_

- [When Select2 is destroyed (browser handles sizing)](https://i.imgur.com/o7p2YAk.jpg)
- [When `width: 'resolve'` is the default (current default)](https://i.imgur.com/jyhAB2p.jpg)
- [When `width: null` is the default (proposed default)](https://i.imgur.com/YBEGWFv.jpg)

You'll notice that `width: 'resolve'` most closely matches the sizing when Select2 is not initialized.

----

> instead the plugin should follow the default behaviour of block-elements, with the option for the user to override this when wanted/needed.

Hang on there, [a `<select>` is inline by default](https://developer.mozilla.org/en-US/docs/Web/HTML/Inline_elemente). That of course doesn't mean that it shouldn't handle block-level cases like one would expect, but it does mean that you can't just throw the inline case out of the window.

PR #3375, opened 2015-05-19
Fixed AjaxAdapter's super constructor call

vyrotek:
Replaced ArrayAdapter super constructor call with correct AjaxAdapter super constructor call. This will allow users to initialize an Ajax select with data.


kevin-brown:
Thanks!

penihel:
and the example of how to use?

Fivell:
@vyrotek  is this related to #3828 ?

PR #3386, opened 2015-05-21
Set select2 id container

carlospalacin:
set id container set property data-s2-id-container on select:

<select name="myselect" data-s2-id-container="s2country">
	<option value="">Select a option</option>
	<option value="1">Option 1</option>
	<option value="2">Option 2</option>
	...
</select>

### DIFF
--- a/dist/js/select2.js
+++ b/dist/js/select2.js
@@ -4718,6 +4718,14 @@ S2.define('select2/options',[
         this.options.dir = 'ltr';
       }
     }
+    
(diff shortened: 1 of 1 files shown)


kevin-brown:
The outer container itself doesn't have an id (usually you don't need it), but the selection container has the `[instance-id]-container` id set, the results container uses `[instance-id]-results`, and individual results use `[instance-id]-result-[result-id]`.

Issue #3209, opened 2015-03-30
[multiselect] Disable showing dropdown when select2-selection__choice__remove is clicked


greggulrajani:
Hello All,

Thanks for the great component. We have managed to get very close with our implementation but we are struggling with one last detail.

When the user removes a multiselect item the dropdown menu is shown. Is there a way to disable the behavior ?  

Thanks in advanced.
Greg

vooydzig:
For me quick fix/workaround/patch was stopping event propagation for click event in select2/selection/multiple.js file: 

    this.$selection.on('click', '.select2-selection__choice__remove',
      function (evt) {
      evt.stopPropagation();
      evt.preventDefault();
      var $remove = $(this);
      var $selection = $remove.parent();

      var data = $selection.data('data');

      self.trigger('unselect', {
        originalEvent: evt,
        data: data
      });
    });

Please verify if it's acceptable solution, as I see no downside to this.

kevin-brown:
Select2 sends the original jQuery Event with the `select2:unselect` event, so you can actually stop the propagation of this event without modifying core Select2 code.

http://jsbin.com/qiweciveve/1/edit?html,js,output

```js
$("select").on("select2:unselect", function (evt) {
  if (!evt.params.originalEvent) {
    return;
  }
  
  evt.params.originalEvent.stopPropagation();
});
```

piotr-cz:
Thanks, this is IMHO optimal way to handle this.

leizard:
it doesn't work with xeditable select2.

```
	$('.x-editable').on('shown', function(e, editable) {
		editable.input.$input	.on('select2:unselect', function (evt) {
			if (!evt.params.originalEvent) return;
			evt.params.originalEvent.stopPropagation();
		});
	});
```

polyspot:
Going from @kevin-brown 's solution, this will prevent the dropdown from showing in all cases when removing a multiselect item, even if you prevent the removal. Note the use of `.args` which is required for some reason within the `unselecting` event.
```
$("select").on('select2:unselecting', function(evt) {
  if (!confirm("Remove \"" + evt.params.args.data.text + "\"?")) evt.preventDefault();
  if (!evt.params.args.originalEvent) return;
  evt.params.args.originalEvent.stopPropagation();
});
```
I found this approach preferable (simpler) in my case - i.e. preventing the dropdown can be done in the `unselecting` event instead of handling `unselect` separately.

mrafei:
[...]

Issue #3169, opened 2015-03-18
selectOnClose stack overflow


timmclaurin:
When I have SelectOnClose = true; I am recieving this error: I am using the most recent version of 4rc2.

```
Uncaught SyntaxError: Invalid regular expression: /(^|\.)select2\.(?:.*\.|)undefined(\.|$)/: Stack overflow		jQuery 
b.event.remove		jQuery 
(anonymous function)		jQuery 
b.extend.each		jQuery 
b.fn.b.each		jQuery 
b.fn.extend.off		jSelect2.
S2.define.AttachBody._detachPositioningHandler		jSelect2.
S2.define.Utils.Decorate.calledMethod		jSelect2.
(anonymous function)		jSelect2.
S2.define.Observable.invoke		jSelect2.
S2.define.Observable.trigger		jSelect2.
S2.define.Select2.trigger		jSelect2.
(anonymous function)		jSelect2.
S2.define.Observable.invoke		jSelect2.
S2.define.Observable.trigger		jSelect2.
S2.define.CloseOnSelect._selectTriggered		jSelect2.
S2.define.Utils.Decorate.calledMethod		jSelect2.
(anonymous function)		jSelect2.
S2.define.Observable.invoke		jSelect2.
S2.define.Observable.trigger		jSelect2.
S2.define.Select2.trigger		jSelect2.
(anonymous function)		jSelect2.
S2.define.Observable.invoke		jSelect2.
S2.define.Observable.trigger		jSelect2.
(anonymous function)		jQuery 
b.event.dispatch		jQuery 
b.event.add.v.handle		jQuery 
b.event.trigger		jQuery 
(anonymous function)		jQuery 
b.extend.each		jQuery 
b.fn.b.each		jQuery 
b.fn.extend.trigger		jSelect2.
S2.define.SelectOnClose._handleSelectOnClose		jSelect2.
S2.define.Utils.Decorate.calledMethod		jSelect2.
(anonymous function)		jSelect2.
S2.define.Observable.invoke		jSelect2.
S2.define.Observable.trigger		jSelect2.
S2.define.Select2.trigger		jSelect2.
(anonymous function)		jSelect2.
S2.define.Observable.invoke		jSelect2.
S2.define.Observable.trigger		jSelect2.
S2.define.CloseOnSelect._selectTriggered		jSelect2.
S2.define.Utils.Decorate.calledMethod		jSelect2.
(anonymous function)		jSelect2.
S2.define.Observable.invoke		jSelect2.
[...]

PR #3397, opened 2015-05-22
Select current result when the tab key is pressed.

rywall:
This results in much more natural behaviour when using a multi-select. Users expect tab completion instead of having it clear what you just typed.

### DIFF
--- a/src/js/select2/core.js
+++ b/src/js/select2/core.js
@@ -298,7 +298,7 @@ define([
       var key = evt.which;
 
       if (self.isOpen()) {
-        if (key === KEYS.ENTER) {
+        if (key === KEYS.ENTER || key === KEYS.TAB) {
           self.trigger('results:select');
 
           evt.preventDefault();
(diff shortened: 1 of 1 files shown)


kevin-brown:
We have the `selectOnClose` option for this.

rywall:
@kevin-brown selectOnClose seems to be broken: http://jsfiddle.net/m8n4obgx/

It hangs my browser for a couple seconds and throws some javascript errors. Am I using it wrong?

kevin-brown:
@rywall https://github.com/select2/select2/issues/3169

rywall:
Ahh I see. Okay I've tried this. `selectOnClose` blurs the input once tab is pressed, unlike my pull request which keeps the input focussed to allow the application of multiple tags. I know this is a breaking change, but would you agree that this is a superior user experience? Maybe this is something for 4.1? Is there a situation where this would result in a worse or confusing user experience?

MrHubble:
@rywall what solution did you end up using for this?

I currently use https://github.com/select2/select2/issues/3209#issuecomment-149663474 to allow multiselect but am wondering if there are better solutions out there.

I would also like to remove the ability for the user to select by clicking outside the select element with `selectOnClose` set to true.

Issue #3297, opened 2015-04-29
Wrong .select2-search__field width right after item is deleted by backspace button


vedmant:
Right after i press delete (backspace) button it sets .select2-search__field to 0.75em and visible only one character http://screencast.com/t/kAQNuyX0yf  http://screencast.com/t/ZGZe9n4jMd , when I press button second time it sets right width http://screencast.com/t/yGhfV8RxmT

nazrhyn:
:+1:

Here's a GIF showing the bug. The key I'm pressing to remove is backspace.

![thinselect2](https://cloud.githubusercontent.com/assets/1672405/7414421/9a22a8e0-ef20-11e4-8b84-0ece09df3d9f.gif)

rrhyne:
I'm seeing this problem with the placeholder text as well. To repro, enter a long place holder text over 15 characters.

temporary solution is to set select2-search__field width to 100%

kevin-brown:
> I'm seeing this problem with the placeholder text as well. To repro, enter a long place holder text over 15 characters.

Sounds like a totally different issue, might want to open a ticket with a jsfiddle about it.

thejae:
Actually setting width to 100% works quite well. There doesn't seem to have any nasty side effect either.

```css
.select2-container--default .select2-search--inline .select2-search__field{
  width: 100% !important;
}
```

 :+1: @rrhyne

PR #3398, opened 2015-05-22
Resize search after removing a choice.

rywall:
Otherwise the input isn't sized correctly.

### DIFF
--- a/src/js/select2/selection/search.js
+++ b/src/js/select2/selection/search.js
@@ -132,6 +132,7 @@ define([
     this.trigger('open');
 
     this.$search.val(item.text + ' ');
+    this.resizeSearch();
   };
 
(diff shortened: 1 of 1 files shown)


rywall:
@kevin-brown to replicate this:

1. Go to: http://jsfiddle.net/1o04w80r/
2. Type "al".
3. Press Enter.
4. Press Delete.
5. Notice the input is sized incorrectly.

kevin-brown:
I need to do some more testing with this, and figure out if there are any related issues that should be handled as well.

rywall:
Will fix #3297

King2500:
:+1:

althaus:
:+1:

robikovacs:
:+1:

Issue #3363, opened 2015-05-15
Select2 v4 is not responsive at all and thus breaks every variable-width layout


cerlestes:
**Abstract**
The newest version of Select2 sets a fixed width on its container. This breaks about every page its used on in every browser when resizing the parent element(s), even the official example page:
![Example](http://i.imgur.com/QkAIqvF.png)

**Steps to reproduce**
  1. Open page with variable width and Select2
  2. Resize browser window
  3. Select2 container remains at old fixed width

**Possible solution**
EDIT: Set the default value for the 'width' option to NULL in order to prevent Select2 from applying a fixed width to the container by default. This is neither needed nor wanted in most cases.
~~Remove the _resolveWidth() method all together and have the Select2 container retain its ```width:auto;``` value. There's no need for setting an arbitrary width to an element that does not require one. 
Optionally set the container to ```display:block;``` by default, so it scales to full width.~~

cerlestes:
See #3370

cerlestes:
See #3371 for a pull-request.

kevin-brown:
Merging into https://github.com/select2/select2/issues/3278.

joetidee:
This is still hapenign in v4.0.3

liamrathke:
Happening for me in v4.0.6 as well

Issue #3361, opened 2015-05-14
Accessing element from the ajax.data callback function


SilverFire:
Using Select2 v.3.5, ajax.data callback function had `this`, pointing to the element, so i was able to get it's properties. In Select2 v.4, `this` points to the ajax configuration object, so I have no access to the element.

This does not work anymore:

```js
$('select').select2({
  ajax: {
    url: 'example.com',
    data: function (param) {
      console.log(param, this);
      return {results: {
        q: param.term,
        module: $(this).data('module')
      }};
    }
  }
});
```

You can see it on the [jsbin](http://jsbin.com/cuzaxe/6/)

How can I access element from `ajax.data` callback function?

penihel:
I have the same problem

penihel:
I did this temporarily workaround:

```javascript
var opts = {
  element: $('#s1'),
  ajax: {
    url: 'example.com',
    data: function (param) {
      console.log(param, this);
      return {results: {
        q: param.term,
        module: $(this).element.data('module')
      }};
    }
  }
};
$('#s1').select2(opts);
```

See the aditional line **"element: $('#s1'),"**

SilverFire:
I have tried this before I posted this issue. I use Select2 as a part of my plugin. The element contains the options storeage, instance of which is stored in `data`. 

If I use this crutch, the recursive link relation appears: the element has a config, which points back on the element with config and so on. This breaks a deep copy of the element with jQuery `$.extend`, which is called when plugin initializes.

Don't remember exec line, but it's [here](https://github.com/select2/select2/blob/a993ed9ca9b2d47804f2ac4dec3358b88e54dca7/src/js/select2/options.js#L94) or [here](https://github.com/select2/select2/blob/a993ed9ca9b2d47804f2ac4dec3358b88e54dca7/src/js/jquery.select2.js#L20)

penihel:
Fixing...
```javascript
var opts = {
  ajax: {
    url: 'example.com',
    element: $('#s1'),
    data: function (param) {
      console.log(param, this);
      return {results: {
        q: param.term,
        module: $(this.element).data('module')
      }};
    }
  }
};
$('#s1').select2(opts);
```

kevin-brown:
[...]

PR #3410, opened 2015-05-27
Fixed bug #3363: `this` in ajax.url, ajax.data points on element

SilverFire:
For further details visit #3361 

P.S.: Grunt updated some CSS. It seems, that previous contributor didn't run grunt.


SilverFire:
@kevin-brown ready to merge

dist/js/select2.js
```diff
@@ -3306,11 +3306,11 @@ S2.define('select2/data/ajax',[
     }, this.ajaxOptions);
 
     if (typeof options.url === 'function') {
-      options.url = options.url(params);
+      options.url = options.url(self.$element, params);
     }
 
     if (typeof options.data === 'function') {
-      options.data = options.data(params);
+      options.data = options.data(self.$element, params);
```
kevin-brown: This should be `options.data.call(self.$element, params)` so the `this` is set instead of the first parameter.

kevin-brown:
Right now there's an issue in how you are calling the methods with the `$element` included, where now it is added as the first parameter instead of being used as the `this` reference.

SilverFire:
@kevin-brown :anguished: oh, sorry, that's my fault. Now it seems to be OK?

SilverFire:
:up:

kevin-brown:
Just a note that _I'm not ignoring this_, I'm actually planning on pulling in the changes soon (likely this weekend).

I almost applied this last weekend but ended up burning most of my time on a regression test that just wouldn't fail.

SilverFire:
:up:

PR #3415, opened 2015-05-27
Allow immediate typing when tabbing to multiple select.

### LINKED ISSUE #3331: Tabbing to a multiselect doesn't automatically open the dropdown and allow typing.
In 3.5.2 I could tab to through a form and the multiselect fields would automatically be in focus with the cursor in the input and the dropdown open. Now I have to press the spacebar or click it with my mouse to achieve the same. I don't know if this is intentional, but it wasn't very intuitive to me.

rywall:
Fixes #3331.

I'd be willing to make this an option if someone can make the case that it shouldn't be default behaviour.

@philipwalton let me know if this works for you.

### DIFF
--- a/src/js/select2/selection/multiple.js
+++ b/src/js/select2/selection/multiple.js
@@ -25,7 +25,11 @@ define([
     var self = this;
 
     MultipleSelection.__super__.bind.apply(this, arguments);
-
+    
+    this.$selection.on('focus', function (evt) {
+      self.$search.trigger('focus');
+    });
+    
     this.$selection.on('click', function (evt) {
       self.trigger('toggle', {
         originalEvent: evt


kevin-brown:
If this is moved into [the `Search` decorator for the selection](https://github.com/select2/select2/blob/bd6148e39f753ed2880f674a90b42ffe446b893a/src/js/select2/selection/search.js) I might entertain the idea of having it as the default behaviour for multiple selects.

Right now this isn't in that decorator, but it depends on that decorator, and will trigger an error (no `self.$search`).

stretch4x4:
@rywall are you going to have time to look at this?
If not I might take a look because this would be useful!

kevin-brown:
The modified version of this patch was committed in https://github.com/select2/select2/commit/02cca7baa7b78e73cdcf393172ee3a54be387167. And the focus war that I discovered (from the patch) was fixed in https://github.com/select2/select2/commit/79cdcc0956e242c1ce642bbaa93e538c54f4be01.

PR #3434, opened 2015-06-04
Support call back function when AJAX request render has been completed

wilsont:
Support call back function when AJAX request render has been completed

Example
```
ajax: {
    url: "/api/search/SearchTips",
    dataType: 'json',
    delay: 250,
    templateRenderCompleted: function ()
    {
    //console.log("on complete");
    }
}
```

### DIFF
--- a/src/js/select2/data/ajax.js
+++ b/src/js/select2/data/ajax.js
@@ -38,6 +38,11 @@ define([
   AjaxAdapter.prototype.processResults = function (results) {
     return results;
   };
+  
+  //Run when the AJAX data render completed
+  AjaxAdapter.prototype.complete = function () {
+              return;
+  };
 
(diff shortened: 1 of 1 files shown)


kevin-brown:
I'm going to have to reject this, as it isn't clear what benefit this has over just hooking into the `ajax.processResults` method.

wilsont:
The reason for doing that is to allow us to have a call back after the "UI" part has been appended to the DOM. 

In my use case I wanna call another plugin to highlight my search result (http://bartaz.github.io/sandbox.js/jquery.highlight.html)

so I will need to do something like this
```
  templateRenderCompleted: function ()
                    {
                        //console.log("on complete");
                        $(".select2-results").highlight(window.currentkeyword);
                    },
```

Please advise if there were better way to do. Thanks!

Issue #3325, opened 2015-05-07
Token seperators don't fire select events


stretch4x4:
In version 4.0.0 when adding new tags using token separators (eg , " ") the change event is being fired but not the select2:select event.
Adding a new tag using enter instead does however fire this event so I think this is a bug.
An example can be found here: https://jsfiddle.net/u0xja2wx/

kevin-brown:
This is because the tokenizer is triggering the `select` event directly instead of using Select2's internal event system.

https://github.com/select2/select2/blob/ac1680d81dcc60009c5d0fe0dd52e70bc065c140/src/js/select2/data/tokenizer.js#L25

stretch4x4:
Cool thanks for the heads up I will see if I can come up with a fix in the next few days

ncolomer:
:+1:

PR #3453, opened 2015-06-10
#3325 adding select event to tokenizer

stretch4x4:
#3325 adding select event to tokenizer


kevin-brown:
I need to write up some tests for the tokenizer so I can ensure that this doesn't regress in the future (and that this doesn't break anything right now), but in theory it looks like it should be fine.

kevin-brown:
Assigning myself to this so I don't forget to review it.

Issue #3332, opened 2015-05-11
Focus is lost when the change event is triggered


stretch4x4:
If a change event is triggered while a select2 has focus, the focus is lost meaning if a user was typing while the backend sent an update their typing will be lost and they will have to select the input again.
Example: https://jsfiddle.net/u0xja2wx/2/
If you enter a new tag and hit enter focus will be lost, comment out the change event and you can continue adding tags after you hit enter.
I believe this can be resolved in the Search.update function by checking if the search box has focus and restoring it at the end of the function if it did.

```js
var had_focus = this.$search.is(':focus');
// ...
if (had_focus) {
  this.$search.focus();
}
```

chanchalkariwala:
Can you be more specific? 
This change would be in Search.prototype.update?

stretch4x4:
Gday,




Yeah that is the general idea.




I submitted a pull request which should show you what I did:

https://github.com/stretch4x4/select2/commit/852715eccdc3f29435cab203e9a852e0e6706e2c




Hope that helps

chanchalkariwala:
Yep, that worked! Thanks!

PR #3455, opened 2015-06-10
#3332 Focus is lost when the change event is triggered

stretch4x4:


src/js/select2/selection/search.js
```diff
@@ -100,6 +100,7 @@ define([
   };
 
   Search.prototype.update = function (decorated, data) {
+    var had_focus = this.$search.is(":focus");
```
kevin-brown: This should be `hadFocus` instead of `had_focus`, trying to avoid inconsistent styles.

kevin-brown:
I left one inline comment and still need to try this out across browsers. Is there any chance you might be able to write up a browser test for this so we can ensure that this doesn't regress in a future version?

It should just require you to...

1. Create a selection container with the search decorator
2. Append it to the fixture
3. Focus the search
4. Force the selection to update
5. Confirm that the search is still focused

src/js/select2/selection/search.js
kevin-brown: Also, I wouldn't mind comparing `this.$search[0] == document.activeElement` to avoid the Sizzle requirement.

stretch4x4:
Ah python habits getting me again.
Yeah I should be able to replace the .is("focus") with your statement
I will have to have another crack at the test, my original attempts didn't work, I was struggling to find an example that was similar enough to copy.
Will give you a yell if I can't get it working

stretch4x4:
@kevin-brown I have had a crack at adding the testing but I don't think I am decorating the search correctly because it doesn't seem to be on the page during the test.. Any suggestions as to where I am going wrong?
```javascript
module('Selection containers - Focus events');

var MultipleSelection = require('select2/selection/multiple');
var Search = require('select2/selection/search');

var $ = require('jquery');
var Options = require('select2/options');
var Utils = require('select2/utils');

var CustomSelection = Utils.Decorate(MultipleSelection, Search);

var options = new Options();

test('updating selector does not shift the focus', function (assert) {
  var $container = $('#qunit-fixture .event-container');
  var container = new MockContainer();

  var selection = new CustomSelection($('#qunit-fixture select'), options);

  var $selection = selection.render();
  selection.bind(container, $container);

  $container.append($selection);

  var $search = $("select2-search__field")

  assert.ok($search.is(":focus"), "The search didn't have focus");
  $selection.trigger('change');
  assert.ok($search.is(":focus"), "The search didn't have focus");
});
```

kevin-brown:
The line where you get the search field based on the class

```js
var $search = $("select2-search__field")
```

It's missing the leading `.` required for jQuery to look it up by the class. Instead it is currently looking for the search field as a custom element.

stretch4x4:
Doh I meant to change that back..
Sadly that isn't the issue, that doesn't work either.

kevin-brown:
This patch landed in https://github.com/select2/select2/commit/698fe7b9e187e182f679aa679eb8b0ecb64a846b and the test landed in https://github.com/select2/select2/commit/88503d2c67dc7f4fb9395a17f17edfe4948cf738, closing this off.

PR #3464, opened 2015-06-10
ContainerCss: Remove magic `:all:` correctly. Fixes #3446.

### LINKED ISSUE #3446: containerCssClass :all: vs. :all
I’ve found a small inconsistency I can’t explain:

https://github.com/select2/select2/blob/master/src/js/select2/compat/containerCss.js#L24-L25

```javascript
if (containerCssClass.indexOf(':all:') !== -1) {
      containerCssClass = containerCssClass.replace(':all', '');
```

Shouldn’t it be `.replace(':all:', '')`? Currently a stray “:” remains in the class list. For example, if I configure

```
containerCssClass: ':all: hello-world'
```

and the `select` element has `my-class-a my-class-b`, the container classes are

```
select2-selection select2-selection--multiple my-class-a my-class-b : hello-world
```

The colon isn’t harmful AFAICS but it’s superfluous, isn’t it? ;)

Thanks.

molily:
I don’t know how to adjust the tests in a meaningful way to cover this. I tried make assertions on `$container.attr('class')` or `$container.attr('class').split(/\s/).length`, but there are two classes set in selection/base.js (`select2-selection`) and selection/single.js (`select2-selection--single`) which a unit test for `ContainerCSS` should not know about.

### DIFF
--- a/src/js/select2/compat/containerCss.js
+++ b/src/js/select2/compat/containerCss.js
@@ -22,7 +22,7 @@ define([
     containerCssAdapter = containerCssAdapter || _containerAdapter;
 
     if (containerCssClass.indexOf(':all:') !== -1) {
-      containerCssClass = containerCssClass.replace(':all', '');
+      containerCssClass = containerCssClass.replace(':all:', '');
 
       var _cssAdapter = containerCssAdapter;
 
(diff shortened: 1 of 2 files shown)


kevin-brown:
This is small enough that it doesn't need a test.

But it would be awesome if you fixed [the same issue in `dropdownCssClass`](https://github.com/select2/select2/blob/44cd938dce3bef6a854e783ee7cf00f4b9626217/src/js/select2/compat/dropdownCss.js#L25).

molily:
Sure! Missed that one.

kevin-brown:
Thanks for making the adjustment!

PR #3473, opened 2015-06-12
Upgraded dev dependency grunt-sass to "^1.0.0".

kahwee:
I cannot build select2 using iojs, there were nan errors during `npm install`.

```
../node_modules/nan/nan.h:339:13: error: no member named 'New' in 'v8::String'
    return  _NAN_ERROR(v8::Exception::Error, errmsg);
```

I tried upgrading dev dependency grunt-sass to "^1.0.0" and it worked.

Also remove explicit node-sass dependency since it is included as part of grunt-sass.


kevin-brown:
Thanks! Looks like there aren't any breaking changes that we need to watch out for.

PR #3477, opened 2015-06-15
Add Japanese translation

wiraqutra:
I've translated select2 4.0.0 into Japanese. (ja.js)


kevin-brown:
Thanks for the translations!

PR #3492, opened 2015-06-18
Replace tabs with spaces for consistency

pquentin:
Tabs remain in dist/js/select2.js and dist/js/select2.full.js, but as I understand those files are generated and should not be in commits modifying src/.


kevin-brown:
[Tests passed](https://travis-ci.org/select2/select2/builds/67328607), looks good. Thanks for cleaning this up!

PR #3495, opened 2015-06-18
Apply select2('option') calls on all elements

### LINKED ISSUE #3413: [4.0] destroy broken on selections with length > 1
Only the first select of the list is destroyed (similar issue #2187 but for v4.0).

Here is a test case (destroy after 3 seconds) :
http://codepen.io/mh-nichts/pen/JdXbdm

Tested on Windows 7 : Firefox 37, Chrome 42, IE 11.

pquentin:
Fixes #3413 

The idea of the patch is simple: as with `'object'` options, we iterate over `this`.

I don't know how to test destroy(), so... I did not test it. I might give it a try with some guidance. Should dist/ changes be part of the pull request? This is not mentioned in CONTRIBUTING.md.

### DIFF
--- a/src/js/jquery.select2.js
+++ b/src/js/jquery.select2.js
@@ -24,18 +24,21 @@ define([
 
         return this;
       } else if (typeof options === 'string') {
-        var instance = this.data('select2');
+        var ret;
+        this.each(function () {
+          var instance = $(this).data('select2');
 
-        if (instance == null && window.console && console.error) {
-          console.error(
-            'The select2(\'' + options + '\') method was called on an ' +
-            'element that is not using Select2.'
-          );
-        }
+          if (instance == null && window.console && console.error) {
+            console.error(
+              'The select2(\'' + options + '\') method was called on an ' +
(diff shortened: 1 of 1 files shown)


pquentin:
The tests pass, but I could have broken another string option. However, 'destroy' seems to be the only documented one. If this is indeed the only option in 4.0, then `thisMethods` should go away.

kevin-brown:
> However, 'destroy' seems to be the only documented one.

`select2('open')`, `select2('close')`, and `select2('destroy')` are [officially supported and documented](https://select2.github.io/examples.html#programmatic).

`select2('val')` (deprecated) and `select2('data')` (partially deprecated) are also on the list, but those are not documented. Interestingly, [they are partially tested](https://github.com/select2/select2/blob/master/tests/integration/select2-methods.js).

pquentin:
Thank you @kevin-brown for your answer! However, I'm not going to work on a fixed version of this pull request. I actually changed the way I use select2 to not use 'destroy'. I hope I still helped the project a little bit.

PR #3506, opened 2015-06-21
Fixes #3471: Changes ALT + UP to close the dropdown instead of opening it

### LINKED ISSUE #3471: Alt + Up is not bound correctly
Right now we are binding both ALT + UP and ALT + DOWN to open the dropdown, but ALT + UP is actually supposed to close the dropdown.

RyanDale:


kevin-brown:
It looks like I manually committed this in https://github.com/select2/select2/commit/d2346cc33186c2a00fa2dad29e8e559c42bfea00.

PR #3507, opened 2015-06-21
Fixes #3503: Removes redundant 'open' event trigger

### LINKED ISSUE #3503: select2:open event firing twice
Hello. I think it is not normal, that `select2:open` event firing twice when I open the `select` element.
Please check out fiddle: http://jsfiddle.net/qav390mp/

RyanDale:
The ```query``` event will fire the ```open``` event if the dropdown has not already been opened. In cases where the ```query``` event has already been triggered it makes an explicit trigger of the ```open``` event unnecessary.

### DIFF
--- a/src/js/select2/core.js
+++ b/src/js/select2/core.js
@@ -395,8 +395,6 @@ define([
     }
 
     this.trigger('query', {});
-
-    this.trigger('open');
   };
 
   Select2.prototype.close = function () {


kevin-brown:
Confirmed this, [it's being triggered](https://github.com/select2/select2/blob/989bd0264928564b02d66f926618f1b5c57f0831/src/js/select2/core.js#L276-L278).

Issue #3531, opened 2015-06-29
AJAX <option> doesn't have text inside


syndbg:
When I have an AJAX form and select a dropdown option, the created option element has a value, but no text inside.

Example: 
```
<select class="select2-input select2-hidden-accessible" data-url="/admin/sources/1285/ajax_platform" data-ajax="true" id="source_platform_id" name="source[platform_id]" tabindex="-1" aria-hidden="true">
    <option selected="selected" value="5">YouTube</option>
    <option value="45"></option>
</select>
```

In the above there's a value but no text which we actually get in *data.name*.

kevin-brown:
[You should be using `text`, not `name`.](https://select2.github.io/announcements-4.0.html#changed-id)

This documentation for this is being tracked at https://github.com/select2/select2/issues/920.

syndbg:
Thing is, that the "Loading remote data" in https://select2.github.io/examples.html has the same issue.

In sake of sanity, the example should at least provide a proper way to utilize select2 AJAX.

PR #3532, opened 2015-06-29
[#3531] Fix AJAX <option> doesn't have text inside

syndbg:
Addresses #3531 

Sets `option.innerHTML` to `data.name` which we get by selecting a dropdown option.

The final result is:

```
<select class="select2-input select2-hidden-accessible" data-url="/admin/sources/1285/ajax_platform" data-ajax="true" id="source_platform_id" name="source[platform_id]" tabindex="-1" aria-hidden="true">
    <option selected="selected" value="5">YouTube</option>
    <option value="45">Twitch.tv</option>
</select>
```

### DIFF
--- a/src/js/select2/data/select.js
+++ b/src/js/select2/data/select.js
@@ -174,6 +174,10 @@ define([
       option.value = data.id;
     }
 
+    if (data.name) {
+      option.innerHTML = data.name;
+    }
+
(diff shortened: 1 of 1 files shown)


kevin-brown:
Closing per my comments in https://github.com/select2/select2/issues/3531.

PR #3538, opened 2015-06-30
Stable/3.5

NickMaev:
If the ajax response data has HTML tags they will be excluded.


kevin-brown:
It sounds like you were trying to create a ticket, not a pull request.

PR #3540, opened 2015-06-30
Prevent HTML tags in title

NickMaev:
If the ajax response data has HTML tags they will be excluded.

### DIFF
--- a/src/js/select2/selection/single.js
+++ b/src/js/select2/selection/single.js
@@ -86,7 +86,7 @@ define([
     var formatted = this.display(selection, $rendered);
 
     $rendered.empty().append(formatted);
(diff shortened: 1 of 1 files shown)


kevin-brown:
Do you want to add a quick unit test for this? Also, does this affect the multiple select as well?

lagunawebdesign:
+1
Should be option **renderTitle: false**  or:
```
templateResult: _mainTemplates, 
templateSelection: _mainTemplates,
...
function _mainTemplates(data){
...
data.title = data.text; // title fix (no html)
...
}

```
```
for temp solution I use:
selectInput.select2({..}).on('change', function(){
			    // temp
			    $(".select2-selection__rendered").removeAttr('title');
			})
```

PR #3544, opened 2015-07-02
#3306: Position clear button before the inline content of `.select2-selection__rendered`

### LINKED ISSUE #3306: Clear button appears under long text when using allowClear
I am using Firefox version 31.5 (limited due to corporate policy).

If the text of the selected item is wider than the input field, the little x seems to be behind the text and cannot be clicked to clear the contents.  It simply drops the selection box down as if I had clicked anywhere on the selection.

I do have a width set.

This works fine in IE 11.

fk:
This should fix #3306, however I only tested in Google Chrome v43.0.2357.130.

### DIFF
--- a/src/scss/_single.scss
+++ b/src/scss/_single.scss
@@ -18,6 +18,10 @@
     text-overflow: ellipsis;
     white-space: nowrap;
   }
+
+  .select2-selection__clear {
+    position: relative;
+  }
 }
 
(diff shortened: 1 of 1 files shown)


SPodjasek:
Confirmed to work on Firefox 40.0.3

rywall:
@kevin-brown Can this be merged please? :-)

PR #3547, opened 2015-07-02
fix: modify announcement 4.0 doc Option params

mouse-lin:


kevin-brown:
Thanks, I completely missed this.

Issue #3548, opened 2015-07-02
Include AJAX parameters as query string parameters by default


dnapierata:
The AjaxAdapter currently only includes `q` as a default parameter to be sent to the server.
It is this block of code in [ajax.js](https://github.com/select2/select2/blob/master/src/js/select2/data/ajax.js) around line 18 which is responsible.
```  
AjaxAdapter.prototype._applyDefaults = function (options) {
    var defaults = {
      data: function (params) {
        return {
          q: params.term
        };
      },
      transport: function (params, success, failure) {
        var $request = $.ajax(params);

        $request.then(success);
        $request.fail(failure);

        return $request;
      }
    };

    return $.extend({}, defaults, options, true);
  };
```

I understand that `q` must be submitted for sake of backwards compatibility but I think it would be very beneficial to submit all `params` along with `q` to avoid users from always overriding the `ajax.data` property of select2 for built in functionality.

Ex: A select where a user wants pagination with infinite scroll (I'm not sure any other params are included as of now other than `term` and `page`)

```
elem.select2({
                ajax : {
                    url : 'doSomething'
                    data : function(params) {
                        return {
                            q : params.term, // search term
                            page : params.page //page
                        };
                    },
                },
            });
```
This seems unnecessary for functionality that is already built into select2 by default to always include a `data` callback that returns the arguments passed.

Here is a proposed solution for [ajax.js](https://github.com/select2/select2/blob/master/src/js/select2/data/ajax.js)
```
AjaxAdapter.prototype._applyDefaults = function (options) {
    var defaults = {
      data: function (params) {
        return $.extend(params, {
          q: params.term
        });
      },
      transport: function (params, success, failure) {
        var $request = $.ajax(params);

        $request.then(success);
        $request.fail(failure);

        return $request;
      }
    };

    return $.extend({}, defaults, options, true);
  };
```

kevin-brown:
> Here is a proposed solution for ajax.js

Feel free to make a pull request for this, I do prefer the idea of just passing all parameters by default.

BenJenkinson:
:+1:

PR #3552, opened 2015-07-06
Return all ajax parameters by default

dnapierata:
This commit resolves issue #3548


kevin-brown:
Just a note that I haven't forgotten this pull request. It's currently queued up to be merged, I just need to write some tests that cover the AJAX adapter (and this PR as a result).

PR #3565, opened 2015-07-09
use merged data instead of existingData for ArrayAdapter

rockshandy:
When options already exist in the targeted select element, new data passed in for use with the Array adapter will not get merged with any existing data.

### DIFF
--- a/src/js/select2/data/array.js
+++ b/src/js/select2/data/array.js
@@ -54,7 +54,7 @@ define([
         var existingData = this.item($existingOption);
         var newData = $.extend(true, {}, existingData, item);
 
-        var $newOption = this.option(existingData);
+        var $newOption = this.option(newData);
 
         $existingOption.replaceWith($newOption);
 
--- a/tests/data/array-tests.js
+++ b/tests/data/array-tests.js
@@ -21,6 +21,21 @@ var arrayOptions = new Options({
   ]
 });
 
+var extraOptions = new Options ({
+  data: [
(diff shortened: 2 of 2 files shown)


kevin-brown:
Can you add [a test](https://github.com/select2/select2/blob/34224604880875a47e5f6e4b5cc3bfe2e01f7b4b/tests/data/array-tests.js) for this? It sounds like this is something that we don't want to lose in the future.

rockshandy:
I was tempted to write a test, but didn't have a chance to look into adding one yet! I should be able to do something this weekend though.

rockshandy:
I have a simple test ready, passes after a recompile. Looking at some past pull requests... should I be recompiling myself, or should I just include the bug fix with the new test?

kevin-brown:
> should I be recompiling myself

It's _preferred_ if you don't recompile (or at least don't include the `dist` changes). The files will be recompiled before the tests are run, and we periodically recompile the master branch.

rockshandy:
I wrote a test and pushed that up as well!

kreintjes:
Experiencing this issue too. Now I am using a hack to remove and store the data from the options, init the select2 and then restore the data to the options. This works, but isn't very nice.

Thanks @rockshandy for the fix and @kevin-brown for select2 in general!

kevin-brown:
Just did a quick sweep of this and it appears to be a reasonable change. Thanks!

PR #3580, opened 2015-07-16
Array-based data adapter supports pagination via 'pageSize' option.

bjdodson:
After a fair amount of searching, it looks like a lot of people are interested in supporting infinite scrolling in conjunction with the array-based adapter for large data sets. I've added support for pagination in the Select adapter. It respects a 'pageSize' argument, defaulting to the old behavior if not found.


kevin-brown:
I like the idea, and I almost went through with doing something similar to this.

Just a few notes

- This needs tests
- You may have better luck generating the array of `matches` and then slicing that you match the page.

I think this might also have issues for cases like nested representations (results that are grouped) as `self.matches` will filter those down and then return the root element.

bjdodson:
Hi Kevin,

I made the first suggested change and track the full set of results and use slice() on subsequent queries. I was concerned that the first query for a single letter would take too long here, but this doesn't appear to be the case.  Code is cleaner this way and subsequent paging is perfectly fast (I can't ever see the "Loading more..." div)

Overall the behavior is quite nice for reasonably sized lists.  My current test has 2,000 items and searching is very smooth, even with 1 character. For much longer lists, we may need an adapter that can do sub-sequence matching (eg, K > Ke > Kev ...) and smart caching when deleting a term-- although adjusting the minimum query length can go a long way.

Frankly my use-case as an address book matcher puts me in the low thousands of entries and the performance is reasonable for now.

layjump:
Why not create a new adapter inherit from ArrayAdapter to achieve this? for example:
```javascript
define([
'./array',
'../utils',
'jquery'
], function (ArrayAdapter, Utils, $) {
  function LazyAdapter ($element, options) {
    this.lazyOptions = $.extend({}, options.get('lazy'), true);
    this.lazyOptions.pageSize = this.lazyOptions.pageSize || 30;
    this.lazyOptions.data = this.lazyOptions.data || [];

    ArrayAdapter.__super__.constructor.call(this, $element, options);
  }

  Utils.Extend(LazyAdapter, ArrayAdapter);

  LazyAdapter.prototype._genPaginatedData = function (data, params) {
    var options = this.lazyOptions;
    var page = params.page || 1;
    var term = params.term || '';
    var pageSize = this.lazyOptions.pageSize;
    var _items = [];

    try {
       _items = data.filter(function(item){
        return (item.text.indexOf(term) >= 0);
      });
    } catch (e) {
      // TODO: handle exceptions here
    }

    return {
      items: _items.slice( (page-1)*pageSize, page*pageSize ),
      totalCount: _items.length
    };
  };

  LazyAdapter.prototype._formatPaginatedResult = function (data, params) {
    params.page = params.page || 1;
    return {
      results: data.items,
      pagination: {
        more: (params.page * this.lazyOptions.pageSize) < data.totalCount
      }
    };
  };

  LazyAdapter.prototype.query = function (params, callback) {
    var data = this._genPaginatedData(this.lazyOptions.data, params);
    var results = this._formatPaginatedResult(data, params);

    callback(results);
  };

  return LazyAdapter;
});
```

siccovansas:
I would love to see this feature in Select2 4.0. I use code quite [similar to this](http://stackoverflow.com/a/25466453/1266303) which gives really nice pagination for large local datasets in Select2 3.5. I'm having quite some trouble to get this functionality working in 4.0 though.

kevin-brown:

PR #3591, opened 2015-07-20
Fix Error if createTag returns null

nobodypb:


kevin-brown:
I'm assuming the issue is related to the fact that the `callback` was being called when it was `null`, so Select2 tried to select a `null` data object? Sounds like a bug.

If you can fix the jshint error (you need braces with your `if`) and possibly add a test case for this, it should be good to go for 4.0.1.

nobodypb:
Yeah, that was exactly my issue which happened when two or more separators were entered without any term. Thats very confusing if your selector is space.

Issue #3559, opened 2015-07-08
$dropdownContainer memory leak ?


drekinov:
Hello we use select2 v4.0 with marionettejs.
select2 is called inside behaviors and on view clean up .select2('destroy') is called

however $dropdownContainer.on('mousedown') event listener is not `off` and GC -

Proposed fix: 

```js
Dropdown.prototype.destroy = function () {
    // Detach mousedown and prevent memory leak
    this.$dropdownContainer.off('mousedown');
    // Remove the dropdown from the DOM
    this.$dropdown.remove();
  };
```

if event is not `off` then heap snapshots show yellow background reference to events['mousedown'].
After some trial and error that leads to `AttachBody.prototype.bind`

kevin-brown:
`$dropdownContainer` is a DOM object generated and managed by Select2. It's the container that gets appended to the `<body>` or right after the selection container, depending on what option you chose.

We were working off of the assumption that it would be removed from the DOM (and eventually garbage collected) when `Dropdown.destroy` was called, but as you can tell that doesn't actually happen. This is because the container itself is never removed from the DOM, but the dropdown within it is.

It's a bug, `AttachBody.prototype.destroy` should be calling `$dropdownContainer.remove()`.

PR #3603, opened 2015-07-22
$dropdownContainer memory leak 

sesteel:
So, again.... This problem was affecting production code for us; I needed this fix due to a pretty substantial memory leak my team was looking into. If this is being provided elsewhere, please ignore. Otherwise, here is a small patch per the suggestion by kevin-brown in the #3559 issue.


package.json
```diff
@@ -30,7 +30,7 @@
   ],
   "license": "MIT",
   "main": "dist/js/select2.js",
-  "version": "4.0.0",
+  "version": "4.0.1",
```
kevin-brown: We can do this at a later time, this shouldn't be included in the pull request.

kevin-brown:
This is probably correct (the changes look accurate), but you changed the package version and then recompiled the files. You can safely include only the `src` changes, and we can handle bumping the versions and recompiling the files at a later time.

PR #3605, opened 2015-07-23
Multiple select, search field: Removes padding on mobile Safari.

molily:
On iOS Safari (tested with iOS 7.1 and 8.4 on iOS simulator), inputs have a large default padding. With an inline width starting at 1.5em, the first characters entered are cut off when there is already a selection. See screenshots.

“X” entered:
x-entered-and-cut-off

Styles:
ua-styles
applied-padding

Not sure if disabling the padding completely is the right solution, but it worked for us. This might affect the overall select height, but this way it’s at least consistent across browsers.

### DIFF
--- a/src/scss/_multiple.scss
+++ b/src/scss/_multiple.scss
@@ -26,6 +26,7 @@
     border: none;
     font-size: 100%;
     margin-top: 5px;
+    padding: 0;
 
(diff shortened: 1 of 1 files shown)


CUnknown:
+1

PR #3629, opened 2015-07-30
Array data paging v4.0.0 

robertio4:
Propongo este cambio debido a la necesidad de paginar grandes array's de datos en la versión 4.0.0. Por lo tanto he añadido soporte a través del método 'query'.

I propose this change because of the need to page through large array 's data in version 4.0.0. So I added support through the ' query ' method.

### DIFF
--- a/dist/js/select2.full.js
+++ b/dist/js/select2.full.js
@@ -3060,6 +3060,20 @@ S2.define('select2/data/select',[
   SelectAdapter.prototype.query = function (params, callback) {
     var data = [];
     var self = this;
+    var more = false;
+
+    var pageSize = this.options.get('pageSize');
+    var page = params.page ? params.page : 1;
+    var seekTo = (page - 1) * pageSize;
+
+    if (pageSize && page > 1 && this.cachedQueryTerm === params.term) {
+      var end = seekTo + pageSize;
+      callback({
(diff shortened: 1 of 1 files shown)


kevin-brown:
Let's keep this to a single pull request, https://github.com/select2/select2/pull/3580.

robertio4:
Sorry, is my first time. How I can make commits to that thread ?

kevin-brown:
Unfortunately you cannot make commits to a branch (or pull request) that is controlled by someone else. You are generally left with two options

- Make a comment on the commit ([which you did](https://github.com/bjdodson/select2/commit/b046d56bfc9b038651c23c9315ede8f4468029a3#commitcomment-12452132))
- Make a comment on the pull request, at the line where it happened

Either way, you are forced to wait for the commit or pull request creator to adjust the commit based on your recommendations.

If the pull request and/or comment goes unanswered for a while (more than two weeks), then you should _definitely_ [cherry pick the commits from their repository](http://stackoverflow.com/q/5120038/359284) and then [add a new commit with the recommended change](https://git-scm.com/book/en/v2/Git-Basics-Recording-Changes-to-the-Repository). This will keep their original commits in tact (with their authorship, so you aren't just taking code) and will allow you to take credit for the changes you discovered were necessary.

PR #3636, opened 2015-08-02
Selections can be removed when Select2 is disabled

goodkiller:
Allow to disable "close" button when option is disabled. Simply works.


kevin-brown:
I was just able to confirm the issue related to this, but I don't agree that removing the icons is the best way to guard against this. The check should really be happening [in the remove handler](https://github.com/select2/select2/blob/14832a8ca4980b58d4930b1471ba9ab4cb72da6e/src/js/select2/selection/multiple.js#L35-L46), which should always be ignored if the component is disabled.

PR #3637, opened 2015-08-03
fix: modify announcement 4.0 doc Tags example

wobbie:


kevin-brown:
I'm assuming this has something to do with jQuery requiring valid JSON objects in the attribute?

PR #3651, opened 2015-08-05
Extended search (ex. 'hwo' will match 'hello world')

vsevolodp:
It try to match input string within beginning of words in data.text respecting order of words.
Word breakers are '()'"_-' and space as defined in function replaceWordDelimiters

Sample match: 'sotma' == '**so**me-**t**ext to **ma**tch'


vsevolodp:
You can play with it here http://jsfiddle.net/zxLvwdh5/5/

vsevolodp:
Made new PR. Closing this one

PR #3654, opened 2015-08-06
Fix IE 8-11 not showing Chinese characters

wdietz:
Fix issue in IE 8-11 that prevents typing characters that require multiple key presses to produce a single glyph.


kevin-brown:
I have no idea why this line of code was left behind. And I also have no idea why it might fix the IE issue.

But I'm fine with removing it in any case.

PR #3656, opened 2015-08-06
Fix not checking for empty

Uxio0:
If $element.data('select2') is not null, but is "", script breaks


pedrofurtado:
@Uxio0 Could you send a JSFiddle example where this scenario occurs?

PR #3677, opened 2015-08-14
[bugfix]fix IE11 autofocus bug

ForeverPx:
The input's placeholder will always trigger the 'input' event , so the select2 is autofocus and auto search when it initial , and whatever you do , it always trigger.

This patch is to fix it . When in IE ,  use keydown event to replace input event .


wdietz:
Feature detection for events.
http://stackoverflow.com/questions/2877393/detecting-support-for-a-given-javascript-event#answers

PR #3678, opened 2015-08-14
Clickable OptGroup

xibman:
When you click on  optgroup it select or unselect children options


kevin-brown:
I'd rather keep this as a third party feature. [An adapter actually exists](https://github.com/bnjmnhndrsn/select2-optgroup-select) already.

PR #3687, opened 2015-08-19
!- maybe means != ?

taf2:
Definitely a syntax error perhaps != was the intention?


kevin-brown:
Yup, that wasn't intentional. Thanks!

PR #3691, opened 2015-08-20
Fix improper grammar in Dutch i18n file

Velocity-:
Grammar is incorrect when maximum is 1, 'kunnen' should be 'kan' in that case.


Velocity-:
Line too long, closing/fixing/rerequesting.

PR #3692, opened 2015-08-20
Fix improper grammar in Dutch i18n file

Velocity-:
Grammar is incorrect when maximum is 1, 'kunnen' should be 'kan' in that case.


kevin-brown:
Thanks!

Velocity-:
No problem, glad I could contribute :+1:

Issue #3695, opened 2015-08-20
Tags are broken in 3.5.3


kevin-brown:
Noticed this right after the release last night.

grabcode:
+1

Noodlewitt:
Also having problems with this in 4.0 - 
here's a fiddle:
http://jsfiddle.net/Noodlewitt/k1gLagmg/1/

kevin-brown:
@Noodlewitt 4.0 and 3.5.x are completely different beats (different code bases as well).

If you're having an issue in 4.0.0, it's preferred if you create a new ticket about it.

tzellman:
I believe this will fix the broken tags. #3705

PR #3705, opened 2015-08-26
Fix broken tags, #3695

tzellman:
The local variable defined will overwrite the tags function, since the var will get hoisted to the functional scope


kevin-brown:
Thanks for the fix!

tzellman:
No problem! Thanks for giving us select2.

Issue #3710, opened 2015-08-26
Problem with auto-scrolling the last active element on the list


lstypka:
There is problem with auto-scrolling highlighted element for list with infinite scroll (select2-more-results). The problem occurs when I try to select last element which is not disabled, but below there are disabled items.
Registered behavior:
![2015-08-26 17h10_55](https://cloud.githubusercontent.com/assets/1587788/9498061/435604f8-4c18-11e5-86e4-b2eea185b9de.gif)

kevin-brown:
This is directly related to https://github.com/select2/select2/issues/1879 and isn't something I have plans to fix in the 3.5.x branch.

The highlight handling is _pretty nasty_ there and it's handled in a different way in 4.0.0.

PR #3711, opened 2015-08-26
#3710 Fixed auto scrolling problem

lstypka:
My change concerns to ensureHighlightVisible method. Original version added offset for 'get more details' when last element was highlighted. This method doesn't take into account case where selected last active element is not the last item on the list. In other words, below the selected element there are inactive / disabled elements. For such a case offset shouldn't be added. I've added additional checking whether the selected last item has a disabled sibling. If so, the offset will not be added.

Behavior before fix:
2015-08-26 17h10_55

Behavior after fix:
2015-08-26 16h49_24

### DIFF
--- a/select2.js
+++ b/select2.js
@@ -1682,7 +1682,7 @@ the specific language governing permissions and limitations under the Apache Lic
 
         // abstract
         ensureHighlightVisible: function () {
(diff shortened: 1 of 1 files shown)


kevin-brown:

PR #3717, opened 2015-08-28
Add select2.css to composer generated require.css

P0rnflake:
Hi,
I have just recognized that the provided stylesheet is not added to the require.css (generated by composer components-installer).

Is there a reason why this was not done?


kevin-brown:
Thanks!

> Is there a reason why this was not done?

Looks like we just never noticed it was an option.

PR #3723, opened 2015-08-31
Adding list of public events

Gavin-Paolucci-Kleinow:
Adding a list of the public events, both non-preventable and preventable, to the Options documentation page. I found that the only way I could get this information was in the source code, and thought it would be helpful to have in the documenation.

### DIFF
--- a/docs/_includes/options/events.html
+++ b/docs/_includes/options/events.html
@@ -32,6 +32,22 @@
     <a href="https://api.jquery.com/on/"><code>.on</code> method</a> provided
     by jQuery.
   </p>
+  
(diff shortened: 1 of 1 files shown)


kevin-brown:
I definitely agree.

Unfortunately we are in the process of redoing that page of the documentation (see https://github.com/select2/select2/pull/3724), so I can't merge this pull request. Feel free to create a new pull request to `docs-faq` though.

Gavin-Paolucci-Kleinow:
Thanks. Sorry about. I've made pull request #3734 to docs-faq.

PR #3734, opened 2015-09-03
Adding Preventable/Non-Preventable Events List

Gavin-Paolucci-Kleinow:
Adding the list of Preventable and Non-Preventable events to the new options documentation.


kevin-brown:
Handled this in https://github.com/select2/select2/pull/3724.

PR #3754, opened 2015-09-09
remove the href property from the select2-search-choice-close link

RustyToms:
For some reason the URL is actually changing to '#' in the Ember.js framework version 1.13. Since this is unnecessary because the anchor tag is not actually being used as a link, perhaps it would be best to remove the href attribute altogether.


kevin-brown:

PR #3757, opened 2015-09-10
Use apply when calling methods

### LINKED ISSUE #3756: Arguments translated to array and passed as single argument
When manually triggering an event in select 2, the arguments for the event itself are translated to a single argument, meaning the event is not correctly triggered.

For example:

    $element.select2('trigger', 'select', { data: item });

gets passed to the trigger function as:

    name: ['select', { data: item }], args: undefined

instead of:

    name: 'select', args: { data: item }

The fix for this is to replace line 38 in jquery.select2.js from:

    var ret = instance[options](args);

with:

    var ret = instance[options].apply(instance, args);

christophano:
Method called using apply to avoid translating multiple args into single arg.
Fixes #3756

### DIFF
--- a/src/js/jquery.select2.js
+++ b/src/js/jquery.select2.js
@@ -35,7 +35,7 @@ define([
 
         var args = Array.prototype.slice.call(arguments, 1);
 
-        var ret = instance[options](args);
+        var ret = instance[options].apply(instance, args);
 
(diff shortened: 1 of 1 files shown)


kevin-brown:
Thanks for the fix.

PR #3779, opened 2015-09-17
Add select2.min.js to scripts list

Restless-ET:


kevin-brown:
Is there any particular reason why this is done?

Typically only the uncompiled assets are included here so they can be integrated into build systems down the line. Minifying already minifed assets has a tendency to cause really strange glitches.

Restless-ET:
When I fetch this as a bower asset I get the minified version, so I was just trying to mimic the same behaviour when fetching as a composer component.

I can use bower so it's not a limitation for me.

If you believe this doesn't make sense feel free to close the PR. No hard feelings. :)

kongr45gpen:
BTW, is there any reason that the full select2.js file is not included with composer?

grappler:
http://stackoverflow.com/a/21266210/850651

kevin-brown:
> BTW, is there any reason that the full select2.js file is not included with composer?

For consistency with other distribution platforms.

PR #3782, opened 2015-09-18
Correct Typo in noResults

nopash:
I corrected a typo in noResult.

### DIFF
--- a/src/js/select2/i18n/th.js
+++ b/src/js/select2/i18n/th.js
@@ -24,7 +24,7 @@ define(function () {
       return message;
     },
     noResults: function () {
-      return 'ม่พบข้อมูล';
+      return 'ไม่พบข้อมูล';
     },
     searching: function () {
       return 'กำลังค้นข้อมูล…';


kevin-brown:
What's the difference between the two? I only ask because we've had this translation for a while, and the Google Translate doesn't sound too bad (the new one doesn't either).

nopash:
Hi

"ไม่" is the correct word meaning "no" or "not" in Thai

"ม่" doesn't have any meaning.

Best regards,

Nopachat Kalayanapan
T: +66 (0) 8 4717 9799
W: www.nopash.com

On Sep 18, 2015, at 8:59 PM, Kevin Brown <notifications@github.com> wrote:

What's the difference between the two? I only ask because we've had this
translation for a while, and the Google Translate doesn't sound too bad
(the new one doesn't either).

—
Reply to this email directly or view it on GitHub
<https://github.com/select2/select2/pull/3782#issuecomment-141458328>.

PR #3793, opened 2015-09-24
Using zepto js

futurist:
Using zepto js in stead of jQuery, add $.data, removeData function, and since Zepto don't support `:not` selector, has to change all `$("div:not(.page)")` into `$("div").not(".page")`

Also, Zepto don't support array value as `.val()` for  multiple select element, has to change function related to `.val()`

That's main issues, and select2 worked with no error, please review pull request. But some function missing, like place holder in single mode select, and multiple mode select with some strange behavior.

I know it's not easy, but want to try.


futurist:
OH, sorry, too many commits pulled, but only want to commit the last d79bdcd commit.

kevin-brown:
This needs to be made against the `stable/3.5` branch as `master` contains the code for 4.0.0.

Issue #3809, opened 2015-10-03
X (clear) button not clickable if select input width is smaller than option text


syndbg:
Reproducible in: http://jsfiddle.net/jEADR/2169/

Scenario:

* Select `Alabama` from input
* Click the X (clear) button
* Verify it works. Good!
* Now select `Long` from input
* Try clicking the X (clear) button
* Verify it does not work.

kevin-brown:
This is a duplicate of https://github.com/select2/select2/issues/3306.

PR #3811, opened 2015-10-03
Fix X(clear) button being unclickable

syndbg:
Addresses #3809.

The solution is to add `z-index` starting from the parent `select2-container` till the `select2-selection__clear` DOM element.

I also linked http://philipwalton.com/articles/what-no-one-told-you-about-z-index/ in the CSS as a comment since people are probably going to wonder why I set opacity when I want to just modify a `z-index`.

### DIFF
--- a/src/scss/_single.scss
+++ b/src/scss/_single.scss
@@ -1,3 +1,5 @@
+@import "mixins/span_z_index";
+
 .select2-selection--single {
   box-sizing: border-box;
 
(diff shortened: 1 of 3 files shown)


syndbg:
@ivaynberg  ping.

kevin-brown:
How well does this compare to https://github.com/select2/select2/pull/3544? I'd rather personally avoid overriding the z-index if possible, it tends to cause other unrelated issues in the long run.

syndbg:
In my opinion it's just a fool-proof approach that won't break. I prefer `z-index`-es.

kevin-brown:
We decided to go with https://github.com/select2/select2/pull/3544 after a short discussion about how neither pull request fully fixes the issue, but they both have the same side effect.

PR #3821, opened 2015-10-09
Speak results in dropdown

### LINKED ISSUE #3735: Screen readers read "blank" or are silent when changing options in the dropdown
I tested the basic example on the examples page with JAWS 16 and NVDA 2015.2 screen readers, and in both cases, the options are not read when expanded and navigating through the options with the up and down arrow keys. JAWS is silent when changing options, and NVDA reads "blank".

Since we already use select2 in production and need to make this accessible, I'd be willing to help test ARIA related modifications.

binary-koan:
Addresses #3735

As of select2 4.0, screen readers give no indication of which result in the combo box is highlighted. Previously, when the highlighted result was changed, the screen reader would read out the text of the item. This patch restores that functionality by setting up aria-autocomplete on the search box and changing its active descendant when moving up/down through the results. This means:

- When a character is typed into the search box, screen readers now read out the first result
- When the up/down arrow key is used to navigate results, screen now should read the currently highlighted result

This has been tested in JAWS and NVDA, but not exhaustively (ie. only with multi-select boxes) so there may be cases where it doesn't work. Also, I'm not sure if it's a good idea to have a separate "results:speak" event, but it made sense at the time for reasons I've now forgotten ... let me know if there's a better place to change the active descendant.

See what you think and if there are any problems just let me know!

Jono

### DIFF
--- a/src/js/select2/selection/search.js
+++ b/src/js/select2/selection/search.js
@@ -37,6 +37,7 @@ define([
 
     container.on('close', function () {
       self.$search.val('');
+      self.$search.attr('aria-activedescendant', null);
       self.$search.trigger('focus');
     });
 
@@ -54,7 +55,7 @@ define([
       self.$search.trigger('focus');
     });
 
-    container.on('results:speak', function (params) {
+    container.on('results:focus', function (params) {
       self.$search.attr('aria-activedescendant', params.id);
     });
 
--- a/src/js/select2/selection/search.js
+++ b/src/js/select2/selection/search.js
@@ -12,7 +12,7 @@ define([
       '<li class="select2-search select2-search--inline">' +
         '<input class="select2-search__field" type="search" tabindex="-1"' +
         ' autocomplete="off" autocorrect="off" autocapitalize="off"' +
-        ' spellcheck="false" role="textbox" />' +
+        ' spellcheck="false" role="textbox" aria-autocomplete="list" />' +
       '</li>'
     );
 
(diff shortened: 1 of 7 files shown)


src/js/select2/selection/search.js
```diff
@@ -54,6 +54,10 @@ define([
       self.$search.trigger('focus');
     });
 
+    container.on('results:speak', function (params) {
```
kevin-brown: Instead of using a custom event here, you can hook into `results:focus` and use `params.element.id` to get the `id` of the element.

kevin-brown:
Sorry about the delay in reviewing this, I've left one in-line comment that hopefully can spark a bit of discussion.

The `results:focus` event is only triggered for options that are selectable, which is why the standard messages (usually errors/warnings) do not trigger it. `aria-activedescendent` is supposed to point to the active element, the one that is currently in focus, so I'm not sure how to best handle error messages.

Ideally we could also get some tests going alongside of this to ensure that we don't lose support for reading out options in the future.

binary-koan:
No problem, I haven't had much time to think about it this week anyway ...

Good point about `aria-activedescendant` - it would semantically make much more sense if it was just set to actually focused results in `results:focus`. There would need to be a way to unset it when there are no results showing though (ie. when you type more and the autocomplete disappears).

How about making errors `aria-live` and/or `role=alert` since they basically act like alert messages?

And yeah, I was thinking it would be hard to write tests but actually it should be straightforward just to check that the attributes are set properly. I'll see what I can do :)

By the way, @joedolson do you and the Wordpress team have any suggestions on this?

binary-koan:
Sorry this has taken me a while, but I've updated with `aria-live` for messages and added a couple of initial tests. Not sure how to test if the `activedescendant` changes when you select something (ie. I can't figure out how to add results to a mock selection then select one), so any help on that would be appreciated!

PR #2654, opened 2014-09-01
Debounce closing on Android browser (take 2)

Soviut:
Second attempt at cleaning up this PR https://github.com/ivaynberg/select2/pull/2653

Issue #2061, opened 2014-01-20
Search auto closes on Android


khawkins98:
Quick note: We're using a very modified version of Select2, so I don't think this should go upstream, but I'm sharing it for posterity.

Our problem was with the Android browser, it appeared to be closing our Select2 search in a way that was different from past issues. 

Symptom:
- Using latest Select2 3.4.5 (and recent versions 3.4.x)
- Browser is native Android (reports were from Android 4.x; Chrome is fine)
- User taps search field
- Select2 interface shows
- 10 or 20 ms later, the Android keyboard shows, select 2 closes
- User can type on keyboard and search works fine

A user workaround was to long tap on the search field, and Select2 works properly.

We worked around this in the code by using a timer, in the style of:

At the start of select2.js:
```
    var openTime = new Date().getTime();
```

At the end of opening: function
```
    openTime = new Date().getTime();
```

At the start of close: function
```
            // if it's been open for less than one second, don't exit
            var tempCounter = new Date().getTime();
            if (openTime > tempCounter - 1000) {
                // console.log('short: ' + (openTime - tempCounter));
                return;
            }
```

The combined effect is to see when select2 has opened, and if the user or the browser tries to close it less than 1 second after opening, prevent it.

A bit of a hack, but for us it makes the operation of select2 more predictable.

edu2004eu:
One year and a few versions later, this still has no explanation to what can cause this.

A few more notes:
- it seems that if you tap on the select in the lower half, it doesn't repro (at least for me) -- this is consistent
- it seems that if I comment out https://github.com/select2/select2/blob/6633cda9061d9fc65075dd140c63f9a6adbcbd72/select2.js#L1422 the issue is fixed, so that might be the call to the `close` method that gets called involuntarily

kevin-brown:
Fun fact: This appears to be fixed in 4.0.

I'd still love to know exactly why it's happening with the mask, but my money is still on a delayed `click` event that is being triggered on the mask.

Dmitry9999:
[...]

PR #3840, opened 2015-10-16
fixed bug on android where dropdown closes immediately after opening.

mandric:
Issue: https://github.com/select2/select2/issues/2061

It looks like this fix never got into 3.5, resubmitting https://github.com/select2/select2/pull/2654.


mandric:
Closing this as it's more of a hack than a fix.  Upgraded to select 4 instead where this issue is fixed.

Issue #3431, opened 2015-06-02
Error when preventing 'select2:opening' event


deje1011:
First of all, thank you for this plugin! I did not completely cotton up to the new API, yet but all in all it works great and is very flexible. 

However I get an error when I try to prevent the dropdown from opening or closing, 
```
uiSelectElement.one('select2:opening', function(e){
    e.preventDefault();
});
``` 
Error : Uncaught TypeError: Cannot set property 'prevented' of undefined

```
  Select2.prototype.trigger = function (name, args) {
    var actualTrigger = Select2.__super__.trigger;
    var preTriggerMap = {
      'open': 'opening',
      'close': 'closing',
      'select': 'selecting',
      'unselect': 'unselecting'
    };

    if (name in preTriggerMap) {
      var preTriggerName = preTriggerMap[name];
      var preTriggerArgs = {
        prevented: false,
        name: name,
        args: args
      };

      actualTrigger.call(this, preTriggerName, preTriggerArgs);

      if (preTriggerArgs.prevented) {
        args.prevented = true; // this is where the error occurs 

        return;
      }
    }

    actualTrigger.call(this, name, args);
  };
```
Line 5161  (I´m using version 4.0)

It seems like there are no arguments passed when triggering the open / close events. 

Thanks in advance, 
Jesse

kevin-brown:
Part of the issue here is that we do not provide in a blank object (`{}`) if one wasn't passed in originally.

althaus:
@kevin-brown Is it possible to release all the current fixes as 4.0.1 and move the open issues to 4.0.2?

MusikAnimal:
For anyone still on Select 4.0.0, you could alternatively use CSS to disable opening/closing Select2:
```
.select2-selection {
  pointer-events: none;
}

.select2-selection__choice {
  pointer-events: visible;
}
```
All browsers and IE10+

MusikAnimal:
For anyone still on Select 4.0.0, you could alternatively use CSS to disable opening/closing Select2:
```
.select2-selection {
  border-radius: 4px !important;
  pointer-events: none;
}

.select2-selection__choice__remove {
  pointer-events: visible;
}

.select2-dropdown--below {
  display: none;
}
```
Super hacky but it works, all browsers and IE11+

PR #3851, opened 2015-10-19
Fixes an issue of args === undefined and script terminates

yallups:
I am returning `false` on `select2:opening` and I get this error for `args` being `undefined`. it terminates execution and breaks other things.

in this patch the functionality remains without breaking execution.

I tried being explicit in the code convention but this could also be a small footprint with
A:
```javascript
args || (args = {});
```
B:
```javascript
args === undefined && (args = {});
```

### DIFF
--- a/src/js/select2/core.js
+++ b/src/js/select2/core.js
@@ -357,6 +357,10 @@ define([
       'select': 'selecting',
       'unselect': 'unselecting'
     };
+    
+    if (args === undefined) {
+      args = {};
+    }
 
     if (name in preTriggerMap) {
       var preTriggerName = preTriggerMap[name];


yallups:
I have this work around at the moment
{% gist ea797dceaa2c18898554 %}

yallups:
Just found the issue that this fixes. #3431
I know that @kevin-brown wants to make sure to always pass an empty object as an argument but I have a feeling that this will keep coming back. If the function is dependent on having an object with a property then we should make sure we have one and it shouldn't be the responsibility of the caller. (IMO ;))

kevin-brown:
> I tried being explicit in the code convention

I'd much rather have the code be explicit in this case, so the current way is fine.

> I know that @kevin-brown wants to make sure to always pass an empty object as an argument but I have a feeling that this will keep coming back.

The main reason for framing this as a "let's fix the missing object" issue instead of a "let's prevent this going further" was because I'd rather favor explicitly passing in no extra parameters when raising an event than having that handled implicitly. That's within the core codebase though, which is something _we control_, and it's not something we can expect for third-party plugins and adapters.

So yeah, this PR seems reasonable.

yallups:
Thanks for the merge! 
I see your point about being explicit in the method call though.

PR #3853, opened 2015-10-20
Stop bubbling up events when removing a choice

### LINKED ISSUE #3209: [multiselect] Disable showing dropdown when select2-selection__choice__remove is clicked

piotr-cz:
Fixes #3209 

I've tried also changing the original click event listener to exclude `.select2-selection__choice__remove` child, but I didn't figure how.

Anyway preventing event to bubble up to main container seems to be cleaner solution.

We are using Select2 as a filter in SPA when changing options triggers re-rendering view with filtered items. When options is removed, options container opens up and when view re-renders options container jumps out to upper left corner and detaches itself from main container.

By the way: what about a new release?

### DIFF
--- a/src/js/select2/selection/multiple.js
+++ b/src/js/select2/selection/multiple.js
@@ -46,6 +46,9 @@ define([
 
         var data = $selection.data('data');
 
+        // Stop bubbling up to selection
+        evt.stopPropagation();
+
         self.trigger('unselect', {
           originalEvent: evt,
           data: data


kevin-brown:
With 4.0.0 we're trying our best to _avoid_ breaking backwards compatibility in the future, and we actually designed how events propagate to allow for others to pick and choose how they want Select2 to work. This was one of those places, which is why `originalEvent` is being passed into the `select2:unselect` event.

> Anyway preventing event to bubble up to main container seems to be cleaner solution.

I definitely agree, and luckily you can do this without modifying the Select2 core as seen in https://github.com/select2/select2/issues/3209#issuecomment-149663474.

> By the way: what about a new release?

Working on it, I've been pretty busy in the last month with non-FOSS things.

PR #3870, opened 2015-10-26
Translation

larsholm:
Translated "The results could not be loaded." to danish: "Resultaterne kunne ikke indlæses."


kevin-brown:
Thanks for the pull request!

I've merged this in manually as https://github.com/select2/select2/commit/ef508db1615f93f4cb3e0760a8e3ffd2ecee9a74, I forgot to mention earlier that translation changes should go in the `src/select2/i18n/` folder instead of `dist/i18n`, so I did that for you.

PR #3875, opened 2015-10-28
Optimize `addSelectedChoice`

zero2lin:
When I set the value with over 100 choices, it works slowly especially on ie8.
So I change the doms loop appending into one time appending in the function `addSelectedChoice`.


zero2lin:
#2848 Fix performance issue with setVal on MultiSelect

PR #3883, opened 2015-10-29
Fixed bug with deselecting items in grouped multiple select when data ids are not strings

### LINKED ISSUE #3882: Cannot deselect items in grouped multiple select when data ids are not strings
On a multi-select, if the ID's of data items are a number, then it is impossible to deselect. I believe [this is the offending code](https://github.com/select2/select2/blob/8ad8f200ba8c9429d636453b8ee3bcf593e8c87a/src/js/select2/data/select.js#L93) in which the `id` is represented internally as a string, but a number in `data.id`.

Jivings:
Fixes https://github.com/select2/select2/issues/3882


kevin-brown:
I think this was fixed in the latest master, because the issue previously was that we were not setting the normalized data object on the `<option>`.

In any case, a failing test case would be useful so we can try to prevent this from regressing in the future.

Jivings:
@kevin-brown Here's my failing [codepen on v4.0.0][1], I'll try and get around to writing a proper test.

[1]: http://codepen.io/Jivings/pen/RWJPVE

kevin-brown:
This has been corrected in one of the previous versions as we are now forcing ids to be strings wherever we find them.

Issue #3287, opened 2015-04-26
browserify errors if jquery.mousewheel is not in node_modules


leftmostcat:
When including dist/js/select2.js, browserify fails upon reaching [line 5328](https://github.com/select2/select2/blob/master/dist/js/select2.js#L5328) due to jquery.mousewheel not being present in node_modules/ for select2.

kevin-brown:
Still need to do some testing here to see if we can just change the word `require` (local variable) to something else, as it doesn't look like Browserify is looking at the syntax tree at all.

Svjard:
+1

samholmes:
+1

Why doesn't select2 just include jquery-mousewheel as a dependency in the package.json file??

kevin-brown:
> Why doesn't select2 just include jquery-mousewheel as a dependency in the package.json file??

jquery-mousewheel is an optional dependency, which is surprisingly difficult to do with browserify. Unlike most module loaders, browserify does not allow modules to be loaded on the fly.

samholmes:
If it's optional because jquery-mousewheel is a shim, then it must be decided if Select2 wants to handle the shim or let the developer handle the shim. A package shouldn't have an *optional* dep. The way I see it, there are two options:

1. Let the developer add the jquery-mousewheel package to his project, and pass this off to Select2 under some kind of configuration. This option is basically saying that Select2 isn't concerning itself with shims/polyfills.
2. Require the jquery-mousewheel package as a dependency in Select2 package.json. This is saying that Select2 uses the jquery-mousewheel package as a shim/polyfill when it's needed, and so it is required to be bundled with it. Select2 can still be smart on when it uses the jquery-mousewheel plugin (using feature detection and whatnot).

The first one is better if you want to keep Select2 more lean and focused (unix-philosophy-like), but the second option is certainly easier on the developer.

jorrit:
I think it should be an explicit dependency too.

jquery.select2.js contains:

```javascript
define([
  'jquery',
  'require',

  './select2/core',
  './select2/defaults'
], function ($, require, Select2, Defaults) {
  // Force jQuery.mousewheel to be loaded if it hasn't already
  require('jquery.mousewheel');
```
The requirejs docs explicitly state on http://requirejs.org/docs/errors.html#notloaded:

> In particular, the following will not work:
> //THIS WILL FAIL
> define(['require'], function (require) {
>     var namedModule = require('name');
> });
[...]

PR #3887, opened 2015-10-30
Add npm dependencies to package.json

krzystof:
I had an error doing `require('select2');` with browserify.
Adding these dependencies to package.json resolves it automatically.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,6 +41,10 @@
       "lib": "dist"
     }
   },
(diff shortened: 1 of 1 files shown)


kevin-brown:
This is a duplicate of https://github.com/select2/select2/issues/3287. `almond` shouldn't be required at all, we're inlining that in all of the files.

PR #3888, opened 2015-10-30
fixed issue with non-scrollable results list

fsw:
In a scenario when one page of ajax results fits into results area, it becomes non-scrollable and "loading more results.." is displayed at the bottom, is unusable and there is no way to load more results:

http://jsfiddle.net/fz7q66pe/3/

This seems to be an edge-case but we had our results being displayed in a large modal and in a grid so this became a problem on large screens.

This fix triggers scroll event after results are appended and it will check if "load more" is already visible and take proper action.


alexweissman:
@fsw I realize that you submitted this PR over 2 years ago, but do you still stand by it as a solution?  I'm worried that manually triggering the `scroll` event could have other, unintended side effects.

alexweissman:
That's exactly what I was thinking!

fsw:
I guess this looks better now. I tested this manually and works fine on my fiddle.
Is there anything more I can do here? Should I commit build dist in this PR as well?

alexweissman:
Nope, no need to (and actually should not) commit any `dist/` code.  I will run the build process and make the changes in `dist/` when this PR is merged.

A qunit test for this edge case would be helpful though, if possible!

diakonos:
Buuuuump. 🤜🏼🤛🏼

I just had to fix this in our local version of Select2. I was about to open an issue to begin the process of getting this properly fixed for Select2 as a whole, but then I found this PR!

Any reason why this never got merged?

I checked out the changes, and I'm actually not sure if it will solve all cases. In my case, I needed to trigger the `loadMoreIfLoadingMoreVisible` function for both `results:append` as well as `results:all` because the `append` event isn't triggered for the first render, and in our case we needed it checked at this point as well.

PR #3903, opened 2015-11-05
Autocomplete should be false to stop the chrome autofill

patmooney:
I'm using Select-2 for a location auto-complete. Chrome likes to whip out its auto-fill box when I search for London ( part of my address ). The auto-fill box covers the select-2 suggestion list.

I tested my change in accordance with this answer on stack overflow -> http://stackoverflow.com/a/29582380

And it worked for me ( Version 43.0.2357.130 (64-bit) )


kevin-brown:
As much as I hate how Chrome is doing this, the proper fix should happen _in Chrome_. Here's the related Chromium ticket.

https://code.google.com/p/chromium/issues/detail?id=468153

PR #3904, opened 2015-11-05
Corrected few Hebrew translations

MeiRct:


kevin-brown:
Changes to translations should be made to `src/i18n/he.js`, where they will later be compiled into the minified distributed format.

PR #3911, opened 2015-11-07
Corrected Hebrew translations

MeiRct:


kevin-brown:
Let me start off by saying thanks for contributing updated translations to Select2. This is a community-driven project, and we're always glad to see solutions from those who use the plugin.

I personally don't know Hebrew well enough to be able to judge this change, and unfortunately this isn't a change where I can do a small amount of research to confirm the wording that is being used. The normal solution is for a [small explanation of the changes](https://github.com/select2/select2/pull/3692) to be included, so we (as the maintainers) can understand why the change is being made, and others can verify the accuracy.

MeiRct:
@kevin-brown Here are the explanations:

About  `errorLoading`:
Before the correction, the translation was: "התוצאות לא נטענו בהלכה".
The word "בהלכה" is wrong and should be spelled "כהלכה", but besides that, this translations means "The results were loaded improperly", while my corrected translation means "Error while loading results", which is what needed.

About `inputTooLong`, `inputTooShort`, `maximumSelected`:
In Hebrew, it's not acceptable to have the digit 1 for representing a single object. You should use the actual word "one", which is "אחד" in Hebrew, together with the singular noun (e.g. "one item" / "one character", instead of "1 item" / "1 character"). 
**These 3 parts needed a critical fix anyway**, because they were addind the English letter 's' to the end of the Hebrew string.

About `loadingMore`:
The translation was "טען תוצאות נוספות" which means "Load more results", as an instruction, probably a typo. 
I changed "טען" to "טוען" so now it means "loading more results", which is the correct form.

kevin-brown:
I fixed a few whitespace issues (nothing major) and squashed the commits into https://github.com/select2/select2/commit/7974fd166294f1ea8653a8ced5ddb506dd23d532, which is now sitting on the master branch and will be included in the upcoming 4.0.1 release.

MeiRct:
@kevin-brown Actually, if mentioning whitespaces clean up, can I ask if there's a specific reason why you didn't use if shorthands in this file ("ternary operator", `condition ? true : false`)? Code readability perhaps?
I used the traditional `if { } else { }` in my commit only in order to follow your original code pattern, but naturally, I always use the shorthand.

kevin-brown:
> Code readability perhaps?

That's the main reason. For many languages, there tends to be multiple levels of nesting necessary to determine the right suffixes, and doing it with a ternary operator tends to create a mess. Compare [`select2_locale_pl.js`](https://github.com/select2/select2/blob/9d44e06dc153d70e0811a91219448323153a9908/select2_locale_pl.js) vs [`select2/i18n/pl.js`](https://github.com/select2/select2/blob/d0ce218c0bf596630bd7779d4767c8d33af9e1e7/src/js/select2/i18n/pl.js).

It was also done for consistency reasons, as the large majority of the code base favors the traditional `if { } else { }` syntax over the ternary operator.

MeiRct:
I just found out that the minified version turns the conditions and declelations into a very compressed form, including a use of ternary. 
I didn't know that the fetching can do that (I prefer not to use the verb "compiling" - not so correct). Is it done with Bower only?

By the way, I'll be grateful if you can provide fetching instructions in short. I was exposed to Bower and other modern JS tools and package managers only recently, and I wasn't sure how to fetch a minified version of Select2 by myself, since I saw both *composer.json* and *bower.json* files in the package root dir.

PR #3924, opened 2015-11-12
clearing a user tag causes wrong selection in single mode

portrayer:
when clearing a user tag,

``` this.$element.val(this.placeholder.id).trigger('change');```
made the <select> selecting null, but then removing the tag option made the select default choose the first option and select2 in wrong state.
```

### DIFF
--- a/src/js/select2/selection/allowClear.js
+++ b/src/js/select2/selection/allowClear.js
@@ -60,6 +60,7 @@ define([
       }
     }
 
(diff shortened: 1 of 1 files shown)


kevin-brown:
For anyone who is running into this issue, the suggestion is to create a blank "placeholder" option that will be selected. The reason why it's selecting the first option is because there is no blank option to reset back to for the placeholder.

PR #3936, opened 2015-11-16
Croatian used for Serbian

Mrkisha:
This was copy paste Croatian to Serbian. Now its proper Serbian.

### DIFF
--- a/src/js/select2/i18n/sr.js
+++ b/src/js/select2/i18n/sr.js
@@ -17,7 +17,7 @@ define(function () {
     inputTooLong: function (args) {
       var overChars = args.input.length - args.maximum;
 
-      var message = 'Obrišite ' + overChars + ' simbol';
+      var message = 'Обришите ' + overChars + ' симбол';
 
       message += ending(overChars, '', 'a', 'a');
 
@@ -26,27 +26,27 @@ define(function () {
     inputTooShort: function (args) {
       var remainingChars = args.minimum - args.input.length;
 
-      var message = 'Ukucajte bar još ' + remainingChars + ' simbol';
+      var message = 'Укуцајте бар још ' + remainingChars + ' симбол';
 
       message += ending(remainingChars, '', 'a', 'a');
 
       return message;
     },
     loadingMore: function () {
-      return 'Preuzimanje još rezultata…';
+      return 'Преузимање још резултата...';
     },
     maximumSelected: function (args) {
-      var message = 'Možete izabrati samo ' + args.maximum + ' stavk';
+      var message = 'Можете изабрати само ' + args.maximum + ' ставк';
 
-      message += ending(args.maximum, 'u', 'e', 'i');
+      message += ending(args.maximum, 'у', 'е', 'и');
 
       return message;
     },
     noResults: function () {
-      return 'Ništa nije pronađeno';
+      return 'Ништа није пронађено';
     },
     searching: function () {
-      return 'Pretraga…';
+      return 'Претрага...';
     }
   };
 });


kevin-brown:
This translation was originally added by @limonte in https://github.com/select2/select2/pull/2027.

I personally do not know the differences between the languages, and Google Translate is pulling up semi-valid translations for both of the languages when these strings are used. We do have translations for Croatian set up (`select2/i18n/hr`), should we maybe migrate some of these back?

limonte:
Both cyrillic and latin alpabets are correct for Serbian. Cyrillic Alphabet is official alphabet in Serbia, but latin is used more nowadays. 

As to me this PR have now much sense. Although typography is bad, @Mrkisha you replaced `…` (ellipsis) with `...` (three dots) which isn't correct.

Mrkisha:
As for elipsis, I get same thing when I copy elipsis `…` and when I type 3 dots `...`. 

As for language, Serbian can be written with latin (croatiаn alphabet) but its not Serbian alphabet. Serbian alphabet is called Azbuka (Азбука). The Constitution of R. Serbia, article 10. says Serbian language is written with Cyrillic alphabet (Aзбука).
`

Mrkisha:
Blic and B92 are foreign owned! And for real, Blic? Geez.
http://www.rts.rs/ and http://www.politika.rs/ are well respected medias in Serbia. Also official website of the parliament http://www.parlament.gov.rs/ and official website of government http://www.srbija.gov.rs/. Just to mention few.

I'm not going into history of why and how croatian latin was introduced to Serbia.

limonte:
This is your personal opinion, I respect it. But objectivity isn't on your side:

http://www.alexa.com/topsites/countries/RS

My examples are on 5th and 6th places. Your are 52nd and 56th. 

---

Offtopic: I'll explain you @Mrkisha  why I personally hate, as Russian citizen, cyrillic alphabet, which is pretty same in Serbian and Russian languages. 

Arabic numerals and so math came to Europe widely in 15th century. So Newton and other heroes of science could make revolutions one after another. 

On the other side, due to cyrillic alphabet and related to it [Цифирь](http://pandia.ru/text/77/352/images/image002_10.jpg) Russia got Arabic numerals and math only in 17th century.

So these guys [Ćirilo i Metodije](https://sh.wikipedia.org/wiki/Kiril_i_Metod) beside introducing cyrillic alphabet to us, took at least 200 years of math and science from Russia.

kevin-brown:
So, now that we've established that there are two variants of the Serbian language (Cyrillic and Latin), we can figure out what the best course of action to take is.

Select2 uses the ISO 639-1 language code to name the language file. In cases where there are variants for a language (`pt-PT` and `pt-BR`, for example), we keep the main language file as the dominant language (`pt-PT` becomes `pt`) when needed, or separate the two (`zh-CN` and `zh-TW`) when there is no dominant language.

In this case, the ISO language code for Serbian is `sr`. The IANA Language Subtag Registry defines subtags for script variants (Latin is `Latn` and Cyrillic is `Cyrl`) which we can use to differentiate between the two. So the existing translation would be `sr-Latn` and the new translation would be `sr-Cyrl`. For backwards compatibility, I'd rather keep `sr-Latn` as `sr`, just so people don't have to change that on a minor update (even if it might not be the dominant or official script, backwards compatibility is important).

-----

So, if these changes were added as a new language, `sr-Cyrl`, I think it would be the best compromise of script variants and backwards compatibility.

limonte:
@kevin-brown :+1:

Mrkisha:
I could not disagree more! Who are we to decide which alphabet Serbia is using for Serbian language? Serbian Constitution article 10 states that Serbian Cyrillic is Serbian Alphabet. Those who what transliterated version of Serbian can use Croatian translation (same language, different alphabet and they are compatible 1 to 1).

kevin-brown:
We have two sides looking to [paint the words on the shed](https://en.wikipedia.org/wiki/Parkinson%27s_law_of_triviality) with the Cyrillic alphabet and the Latin alphabet. The One True Alphabet™ is not something that we're going to argue on this tracker.

> Who are we to decide which alphabet Serbia is using for Serbian language?

We aren't. Well, we are because we have to keep `select2/i18n/sr` as the Latin alphabet **for backwards compatibility reasons**. It's not something I'm going to budge on, because it's a noticeable breaking change and not a bug.

I'm _perfectly fine_ with a `select2/i18n/sr-Cyrl` existing for those who need Serbian translations with the Cyrillic alphabet.

> same language, different alphabet and they are compatible 1 to 1

I'm going to have to disagree with this part, simply going based off the fact that there is enough of a difference to have them called different languages. And Wikipedia [demonstrates a (small) difference](https://en.wikipedia.org/wiki/Comparison_of_standard_Bosnian,_Croatian_and_Serbian#Language_samples).

limonte:
@kevin-brown I believe that @Mrkisha won't change his mind and so PR, so I made #3943 as discussed here.

btw. @Mrkisha you forgot to replace latin `a` with cyrilic `а` in lines 22 and 31 :)

Mrkisha:
So drupal released today their new major version 8. And lets visit their translations page and lets choose Serbian https://localize.drupal.org/translate/languages/sr/translate?project=drupal&status=1&release=454938. What alphabet is used for translation?

Mrkisha:
@kevin-brown using wikipedia to prove me wrong is pathetic. Anyone can write anything on wikipedia. Second since you mentioned the language differences, what language do you speak? American, English, Schotish, Welsh, Southafrican, Australian...? Seriously, you do not even know what dialect is, not to mention anything else... Please do not embarrass yourself any more.

limonte:
@Mrkisha for which language are you native speaker - Serbian or Croatian?

kevin-brown:
I've closed this off now that https://github.com/select2/select2/issues/3943 has been created.

> We have two sides looking to [paint the words](https://en.wikipedia.org/wiki/Parkinson%27s_law_of_triviality) on the shed with the Cyrillic alphabet and the Latin alphabet. The One True Alphabet™ is **not something that we're going to argue on this tracker**.

PR #3943, opened 2015-11-19
Add sr-Cyrl language.

limonte:
As discussed in #3936 

>
>Select2 uses the ISO 639-1 language code to name the language file. In cases where there are variants for a language (pt-PT and pt-BR, for example), we keep the main language file as the dominant language (pt-PT becomes pt) when needed, or separate the two (zh-CN and zh-TW) when there is no dominant language.
>
>In this case, the ISO language code for Serbian is sr. The IANA Language Subtag Registry defines subtags for script variants (Latin is Latn and Cyrillic is Cyrl) which we can use to differentiate between the two. So the existing translation would be sr-Latn and the new translation would be sr-Cyrl. For backwards compatibility, I'd rather keep sr-Latn as sr, just so people don't have to change that on a minor update (even if it might not be the dominant or official script, backwards compatibility is important).

### DIFF
--- a/src/js/select2/i18n/hr.js
+++ b/src/js/select2/i18n/hr.js
@@ -15,6 +15,9 @@ define(function () {
   }
 
   return {
+    errorLoading: function () {
+      return 'Preuzimanje nije uspjelo.';
+    },
     inputTooLong: function (args) {
       var overChars = args.input.length - args.maximum;
 
--- a/src/js/select2/i18n/sr-Cyrl.js
+++ b/src/js/select2/i18n/sr-Cyrl.js
@@ -14,6 +14,9 @@ define(function () {
   }
 
   return {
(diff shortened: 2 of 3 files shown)


kevin-brown:
I'm just going to pull this comment on in

> Side note: We are also missing [the `errorLoading` translation](https://github.com/select2/select2/blob/7e2ade09652f66f3a70787f6f0e2918d171a6dcb/src/js/select2/i18n/en.js#L4-L6) for these files. Is anyone interested in adding it?

Right now this will default back to the English translation if there are any issues when loading the results.

limonte:
@kevin-brown I've added `errorLoading` translations to `sr`, `sr-Cyrl` and `hr` languages.

kevin-brown:
Thanks!

PR #3946, opened 2015-11-19
Add the decorator parameter to the handlers for the attach and detach of the dropdown

manuel-alvarez-alvarez:
Otherwise the container attribute is undefined and the listeners stay in the DOM.


kevin-brown:
Didn't notice this before, and it hasn't been triggering any errors because we never use any properties on objects attached to `container`, we only use properties on it (which will return `null`).

Thanks for the pull request.

Issue #3158, opened 2015-03-15
Version Select2 4.0.0 setting a title of selects


afattahi54:
I have updated to Select2 4.0.0-rc.2 and find that it will set the title of select element. 
Please see http://jsfiddle.net/h0mqofwq/1/
How can we disable it ?!

Knoxvillekm:
set options: placeholder and allowClear
http://jsfiddle.net/Knoxvillekm/h0mqofwq/2/

Why when I set allowClear: true, I have selected option after init?

afattahi54:
@Knoxvillekm you are asking something else, please ask a new question with new topic

Hatamikhah:
yes, I want to disable this feature in Select2 4.0.0-rc.2 for specific reasons. Could you help me in this issue?

afattahi54:
Manually removed by:

    $(".select2-selection__rendered").attr("title","");

jwynveen:
This issue pertains to v3.5.2 as well and affects the use of select2 with jquery validate. Because select2 adds the title attribute, jquery validation uses that title as the validation message. It should be an option to disable this feature of automatically adding title attributes for the cases like this where it conflicts with other plugins.

Fiddle showing the issue. Click submit on form to see validation messages. https://jsfiddle.net/jwynveen/rn2vgnsj/2/

Daijobou:
Maybe that helps? https://stackoverflow.com/a/46327202

pedrofurtado:
We don't have immediate plans to provide this. We are focused now to fix UI bugs. But if you open a PR with unit tests, I will be glad to review and approve if everything is ok.

PR #3988, opened 2015-12-04
Add on option to disable title rendering on selected element

nicolasqos:
There is no way to prevent rendering title on selected element. 

This is particularly harmful when we want to create our own title attribute on select2 or when we do not want title at all. 

See #3158


nicolasqos:
:+1:

code-chris:
Can this be merged soon? It would be very helpful.

dciesielkiewicz:
In fact, it's exactly what I need. Can this be merged?

Vinze:
I'm also having trouble with the title attribute being added. I'm loading some json content in my original select field which greatly helps with formatting the results in select2, but the downside is that the title attribute shows the original content:

img

Why am I loading a json string in the original select? Because of this:

```js
function formatSelection(option) {
   var a = JSON.parse(option.text);

   return $('<span>' + a.name + '<br><small>' + a.street + ', ' + a.city + '</small></span>');
}
```

eayin2:
This fixed my issue. Why hasn't this been merged? I reckon for these four lines you don't need to add tests :-)

In current select2 versions there's a second occurance of $selection.prop('title', ..);
so that also has to be just wrapped with the if-condition
https://github.com/select2/select2/blob/062c6c3af5f0f39794c34c0a343a3857e587cc97/src/js/select2/selection/multiple.js#L96

alexweissman:
Hey folks, I'm just taking a look at this issue but I've never encountered this before in my own usage of Select2.  What exactly is the issue, and how can I replicate it?

eayin2:
@alexweissman Hey

Actually this issue can be avoided by setting also `title: "something"` in the ajax call for each select option.

For example when using django-autocomplete-light you would add to their select2() ajax call
`value.title: "something` in here
https://github.com/yourlabs/django-autocomplete-light/blob/master/src/dal_select2/static/autocomplete_light/select2.js#L91-L92

Another example from Stackoverflow https://stackoverflow.com/a/36361433:
```
var data = [{
  id: 0,
  text: '<div style="color:green">enhancement</div>',
  html: '<div style="color:green">enhancement</div><div><b>Select2</b> supports custom themes using the theme option so you can style Select2 to match the rest of your application.</div>',
  title: 'enchancement'
}, {
  id: 1,
  text: '<div style="color:red">bug</div>',
  html: '<div style="color:red">bug</div><div><small>This is some small text on a new line</small></div>',
  title: 'bug'
}];

$("select").select2({
  data: data,
  escapeMarkup: function(markup) {
    return markup;
  },
  templateResult: function(data) {
    return data.html;
  },
  templateSelection: function(data) {
    return data.text;
  }
})
```

When you omit the title, then you will see the html code.

alexweissman:
So a couple of things:

1. This seems like a very specific feature.  Could it be generalized (perhaps as part of the templating system)?  In general, we should prefer more generic control features over zillions of boolean configuration options.
2. This does not implement the feature for multi-select.

VictorioBerra:
multi-select user here, I am using the [jQuery DataTable's Editor's Select2 plugin](https://editor.datatables.net/plug-ins/field-type/editor.select2). This is really killing me right now.

While I do aggree with the statement about generalization, I also think thatselect2 should not be making assumptions about how I want to display my data. Especially doing it in a way that there is no off button. I have to hack around it by tapping into events and stripping out tags by hand.

I think that there should be a way to easily disable this feature off for an instance of select 2.

kevin-brown:

DavidMoraisFerreira:
@kevin-brown Hi, if someone were to invest some time into cleaning up this PR, would you consider reviewing and (hopefully) merging in the changes?

PR #3993, opened 2015-12-04
Added Armenian Localization

mhamlet:


kevin-brown:
Thanks for submitting a new translation to Select2!

There are a few build failures that are going to need to be corrected before this translation can be accepted.

https://travis-ci.org/select2/select2/builds/94978262

PR #3997, opened 2015-12-07
Fix #1456 Add an option to trigger submit on enter

### LINKED ISSUE #1456: Triggering submit with enter?
I'm running s2 on an input field that grabs data via ajax.

For the life of me I can't figure out how to have the form submit on enter keypress, when a input value is picked. Using various js method didn't seem to work either, I assume because of how s2 handles the input.

What am I missing?

PMickael:
``` javascript
$(".select2").select2({
        enterTriggerSubmit: true,
});
```

### DIFF
--- a/src/js/select2/core.js
+++ b/src/js/select2/core.js
@@ -321,7 +321,9 @@ define([
           evt.preventDefault();
         }
       } else {
-        if (key === KEYS.ENTER || key === KEYS.SPACE ||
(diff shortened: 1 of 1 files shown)


kevin-brown:
Shouldn't the `submit` event be triggered on the `<form>` instead of the `<select>`?

There are plans to trigger these key-related events on the `<select>` element in future versions, so that developers can manually implement this how they want.

PMickael:
It's better to treat directly with the select who will trigger is parent form submit like any form elements.

Issue #3420, opened 2015-05-28
createSearchChoicePosition removed in select2 4.0 withour replacement


kreintjes:
I know the createSearchChoice option is replaced by the createTag option in select2 4.0. However, the old createSearchChoicePosition option seems to be completely removed without replacement (createTagPosition does not work).

Am I correct and if so, any plans on restoring this option?

kevin-brown:
This is exposed as [the internal `insertTag` method](https://github.com/select2/select2/blob/bd6148e39f753ed2880f674a90b42ffe446b893a/src/js/select2/data/tags.js#L100-L102) on the tags decorator, but it is not yet exposed through the options when initialzing Select2.

kreintjes:
Thanks for the quick reply. So for now it is not possible to override it yet, but it will be in the (near) future?

PR #4008, opened 2015-12-09
Search tag position option

stretch4x4:
Fixing issue #3420
Adding an option to move the location in the search list of the typed tag to the bottom.
We needed this as the users were hitting enter and saving their tag instead of the matched item.


kevin-brown:
I was planning on adding this as an option at some point (`insertTag` perhaps), similar to how `createTag` and `sorter` work. As a result, I'm hesitant to accept this pull request because I'd rather support a more generic option that inserts the tag into the results instead of a specific option that controls the location of it.

stretch4x4:
That sounds fair..
I will have to try to take a look at createTag and sorter to see if I can figure out how that would work..
It might be pushing my abilities to try to do that myself..
But would like to help fix this issue so that next time we upgrade I don't break anything.

stretch4x4:
Figured out what you meant and that was much simpler than expected so have updated to that approach :)

kevin-brown:
I've manually committed this in https://github.com/select2/select2/commit/e842797aaf9beb0327a106b92ae5f412ee0b455f, which is almost the exact same commit as https://github.com/stretch4x4/select2/commit/7276620fde0bd7c7123a9b7b7f7a128b5e715df0. I also built out a (very basic) test for this in https://github.com/select2/select2/commit/db5e947664aca0e9b84db437f1009bd28a03b86b.

stretch4x4:
Cheers

Issue #3354, opened 2015-05-13
Hitting backspace on multi-select converts element to text


pjclas:
Perhaps this is intended behavior but it seems odd to me.  If you use the backspace key to delete an element in a multi-select box, that element doesn't go away like it would if you clicked the 'x'.  Instead it is converted to text with an additional space at the end that you have to backspace through to get rid of.  Is this intended?  Would be nice to have an option to disable it if that is the case.  Fiddle providede below:

http://jsfiddle.net/y7fo61ms/

kevin-brown:
> Is this intended?

Yup, totally intentional. https://github.com/select2/select2/commit/395e06aff52059f654aff47f03a26cad8955a092

> Would be nice to have an option to disable it if that is the case.

I'm not sure how I feel about having another option for this, I'd like the get the current set documented first. But I can tell you that you can completely erase the search box in [`Search.searchRemoveChoice`](https://github.com/select2/select2/blob/dd9af6893450bbb054ca79546a16bb18198f3e88/src/js/select2/selection/search.js#L127-L135) and it will do what you are looking for.

pjclas:
Hmm, ok.  I guess I'll have to see if I can wrap my head around this
behavior.  Thanks for the info.

smajl:
Wow, I was just surprised by this behavior too. Not what I was used to from 3.5.x. Well, it can be usefull, but there is also a nasty bug, which is already discussed here https://github.com/select2/select2/issues/3297

hackeron:
How to get the old behaviour? - the new behaviour is very undesirable and users are complaining why can they cannot backspace to remove an item and why does it turn into text. Is there any workaround in the meantime?

stonexer:
Can't agree with this behaviour. Just vote to get the old one.

kevin-brown:
Reopening this while we decide how best to handle this.

Zane-XY:
this is really odd,  my tag contains spaces, after I hit backspace, it breaks down to more small tags automatically, and it seems I can never delete all of them.

heebj:
+1

ldaguise:
I also vote for having an option to get back the old behaviour.
Thanks

giangt:
+1
Thanks.

gtrman:
+1

rnt363:
[...]

PR #4009, opened 2015-12-09
Tag backspace behaviour option

stretch4x4:
Fix for #3354
This patch allows for configuration of this behaviour, you can have the existing default behaviour to insert the text, have nothing inserted or specify something else on the item to insert eg the id


stretch4x4:
@kevin-brown Maybe this would also be a good candidate for the option being the function to call..

pleek91:
Is there documentation for this new option? Thanks

stretch4x4:
Not at this stage no. Other than creating a demo for the example page I am not really sure what documentation is required. Happy to write some if someone points me at an example of what is required.
But I think @kevin-brown is still looking at if there is a better solution anyway.

pleek91:
I'm just looking to have backspace delete the tag rather than replacing it with the tag's text.

src/js/select2/selection/search.js
```diff
@@ -194,11 +194,23 @@ define([
...
 
-    this.$search.val(item.text);
+    if (
+      this.options.options.searchRemoveAttribute ||
+      typeof this.options.options.searchRemoveAttribute === 'string'
+    ) {
+      if (item[this.options.options.searchRemoveAttribute]) {
+        searchValue = item[this.options.options.searchRemoveAttribute];
+      } else {
+        searchValue = '';
```
stretch4x4: Looking at this again I should probably update this to set searchValue to the value of searchRemoveAttribute.. This would make it clearer to use an empty string to do nothing and also allow backspace to always insert a given string (not that I have any idea why you would want that)

stretch4x4:
In that case @pleek91 you would need to copy in the above code / super that function with my version.
Then set `searchRemoveAttribute = ''` I believe that should do what you need

pleek91:
@stretch4x4 if its not to much to ask, could you create a jsfiddle for this? I'm not quite following. Thanks

stretch4x4:
See if this makes more sense.. It won't actually work though because it doesn't have the core change.
I am not very good with the superclassing which is the only way you could make the fiddle work.
https://jsfiddle.net/wq5gt85o/
But if you copy the edited version of `searchRemoveChoice` over your locally copy it should work.
If you use a cdn version of the library you will have to wait till it is merged in or learn how to overide it locally

pleek91:
Thanks, that ultimately put me on the right track. Ended up just commenting out line 2000 of select2.js. Thanks for your help @stretch4x4

stretch4x4:
@kevin-brown have you had a chance to think about this any further?
Just did some testing on the current RC and realised I had lost this from my local version.

Issue #4052, opened 2015-12-30
Not able to select element with id 0


AnirudhaGohokar:
Am not able to select element which has an id set as 0.
Though am able to select elements with id > 0
Here's a [jsfiddle](http://jsfiddle.net/mdb13b5d/) for reference. 

I think its being being used to show "Searching" element which is by default disabled. Can this be changed?

kevin-brown:
This has to do with your id being `0` and not `"0"` (notice the quote difference). Converting the id to a string fixes the issue.

https://jsfiddle.net/mdb13b5d/1/

This has to do with how Select2 compares ids in non-local results, and the fact that a standard `<select>` expects all ids to be strings.

AnirudhaGohokar:
Thanks Kevin..
This has resolved the issue.

kevin-brown:
I'm going to keep this open as a documentation issue, as this question has come up in the past.

PR #4076, opened 2016-01-11
Fix bug with creation option item without "value" attribute if data.id is 0 (number)

blomnik:
Fix `data.id` condition expression


kevin-brown:
This is related to https://github.com/select2/select2/issues/4052, which is a common issue. Considering how many places we expect that ids should be strings (like in a standard `<select>`), I'm not sold on us special casing this.

PR #4084, opened 2016-01-14
Added array example when using data attribute

fmpwizard:
`tokenSeparators` expects an array of characters when using javascript to initialize select2. When using the data-\* attribute `data-token-separators` , you don't write an array, you simply write the characters one after the other, in quotes.

### DIFF
--- a/docs/_includes/options/core/data-attributes.html
+++ b/docs/_includes/options/core/data-attributes.html
@@ -34,6 +34,30 @@ $("select").select2({
 });
 </pre>
 
+
+  <h3>
+    How should <code>array</code> options be written?
+  </h3>
+
+  <p>
+    This means that if you declare your <code>&lt;select&gt;</code> tag as...
+  </p>
+
+<pre class="prettyprint">
+&lt;select data-tags="true" data-token-separators=",;" &gt;&lt;/select&gt;
+</pre>
+
+  <p>
+    Will be interpreted the same as initializing Select2 as...
+  </p>
+
+<pre class="prettyprint linenums">
+$("select").select2({
+  tags: "true",
+  tokenSeparators: "[',', ';']"
+});
+</pre>
+
   <h3>
     Are options with nested configurations supported?
   </h3>


kevin-brown:
Any particular reason why this works? I would normally expect jQuery to pick up the array (written as a string in `data-token-separators`) and convert it to a JS array.

In theory,

```html
<select data-token-separators=";,"></select>
```

Should be equivalent to

```js
$('select').select2({
  tokenSeparators: ';,'
});
```

Which isn't exact an array (though it is array-like).

fmpwizard:
Hi @kevin-brown , thank for looking into it, I tried several array like syntaxes and select2 kept just ignoring my separators and would use anything that was inside quotes, like, if I used: `data-token-separator="[',', ';']"` , then typing a space triggered a new tag work, instead of just the command and semicolon. As it took me a while to get this working, I thought it would be good to have an example on the docs.

Let me know if there is anything else I can do.

Thank you.

kevin-brown:
It's because you are using single quotes instead of double quotes in your array. JSON requires double quotes, so jQuery isn't parsing it as valid JSON.

[Related Stack Overflow question](http://stackoverflow.com/q/6071618/359284#6071653)

fmpwizard:
Hi Kevin, sorry for the delay, you are right, if I use single quotes around the array syntax, it works, unfortunately that doesn't work for me because I'm using Scala to generate the html and their parser forces me to use double quotes on any attribute.

Would you accept the PR for documentation if I add both syntaxes, with  a note that if your server side language forces double quotes, this is the alternative syntax that also works?

Thanks

kevin-brown:
> Would you accept the PR for documentation if I add both syntaxes, with a note that if your server side language forces double quotes, this is the alternative syntax that also works?

There's a problem with this: technically the alternative syntax is wrong. It's not just that it's wrong in the "it technically produces the equivalent data" sense, it only works because strings are indexable similar to how arrays are. Select2 expects an array for the `tokenSeparators` option, and **passing a string is not officially supported and may break in the future**.

It works, yes. Only if you need to provide an array of characters, but not if you need to provide an array of strings (or other data, like numbers). So I'm not comfortable with us recommending it as an alternative syntax for arrays when it doesn't actually produce an array.

I would be comfortable with documenting the official way for providing an array (using JSON) in the attribute.

PR #4118, opened 2016-01-23
Add missing diacritics to diacritics.js

mlpo:
I noticed that diacritics are missing in “diacritics.js”, it is “œ” and “Œ” (especially used in French « cœur » :heart:)

This patch is pretty trivial.


mlpo:
Well, actually, wait before merging. Other diacritics are missing and may be added. I take care to do that.

Issue #3618, opened 2015-07-28
Error when form name with [] array


nurmanhabib:
For example I created a form select using [] array.
```
<select name="kunjungan[temp-0][details][1][details]" class="form-control select2">
    <option value="somevalue">Some Value</option>
    <option value="othervalue">OtherValue</option>
</select>
```
Then I initialization select()
```
$('.select2').select()
```

then when I observed on the web browser console, an error occurs.
```
SyntaxError: invalid range in character class
```
I tried to eliminate the name without the array. The results are normal.
```
<select name="kunjungan_temp-0_details_1_details" class="form-control select2">
    <option value="somevalue">Some Value</option>
    <option value="othervalue">OtherValue</option>
</select>
```

kevin-brown:
Can you expand upon the issue you are facing by adding a few more pieces of information to your bug report?

- What steps can be used to reproduce the issue?
- What is the bug and what is the expected outcome?
- What browser(s) and Operating System have you tested with?
- Does the bug happen consistently across all tested browsers?
- What version of jQuery are you using? And what version of Select2?
- Are you using Select2 with other plugins?

The [contributing guide][cg] contains a useful section with a few questions which can help us track down the bug and speed up the process of finding a fix for it.  The creation of [an isolated test case][test-case] would also be useful.

[cg]: https://github.com/select2/select2/blob/master/CONTRIBUTING.md#reporting-bugs-with-select2
[test-case]: http://css-tricks.com/6263-reduced-test-cases/

nurmanhabib:
Like the example I reported above, the actual plugin select2() is running normally.
```
<select name="kunjungan[temp-0][details][1][details]" class="form-control select2">
```
Bug:
select2 selector can not seem parsing dashes in "temp-0", because I've just tried without dashes `kunjungan[temp0][details][1][details]` running normally

I've tried with Mozilla Firefox 39.0

I also use some plugins, such as TinyMCE, typeahead, and others.

jQuery v2.1.4
Select2 4.0.0

Max-Might:
I get the same error with Select2 4.0.1 and jQuery 2.1.4
The error happens always when selecting a value from a drop-down.

Tested with browsers Firefox 42 and Opera 34.

Max-Might:
> select2 selector can not seem parsing dashes in "temp-0"

Yeah, this seems to be the problem.
IMO this issue is quite critical because it prevents people from using select2.

semmais:
[...]

PR #4123, opened 2016-01-25
Remove square brackets when generating id

bnjmnhndrsn:
This fixes [issue 3618](https://github.com/select2/select2/issues/3618) by removing any square bracket when generating an id using the `name` attribute of the select element. The presence of square brackets, when used for namespacing events, seems to have been wreaking havoc with the regular expressions jQuery uses to match event names. This caused two sorts of errors:
 1.  If the name attribute contained inadvertently contained a malformed regular expression, it would cause errors. [Trying opening and closing the dropdown on thus fiddle to see error](https://jsfiddle.net/ogg1o102/1/)
 2. If the name attribute contained any sort of square bracket, it would cause event handlers to not be removed properly. [See fiddle for a general example](https://jsfiddle.net/vtojt4wo/1/).

While escaping the `name` attribute using the [guidelines here](https://learn.jquery.com/using-jquery-core/faq/how-do-i-select-an-element-by-an-id-that-has-characters-used-in-css-notation/) seems to fix the first case, it doesn't seem to fix the problem of event handlers not getting removed. [See this fiddle as another general example](https://jsfiddle.net/vtojt4wo/4/). For this reason, I've altered `_generateId` to remove any square brackets from the id entirely. 

As this is a small change, but one which fixes bugs in a different place, I wasn't sure where in the test suite, tests should be added. Any guidance on that and I would be happy to add them.

### DIFF
--- a/src/js/select2/core.js
+++ b/src/js/select2/core.js
@@ -94,7 +94,8 @@ define([
     if ($element.attr('id') != null) {
       id = $element.attr('id');
     } else if ($element.attr('name') != null) {
-      id = $element.attr('name') + '-' + Utils.generateChars(2);
+      id = $element.attr('name').replace(/\[|\]/g, '')
+      id =  id + '-' + Utils.generateChars(2);
     } else {
       id = Utils.generateChars(4);
     }
--- a/src/js/select2/core.js
+++ b/src/js/select2/core.js
@@ -94,7 +94,7 @@ define([
     if ($element.attr('id') != null) {
       id = $element.attr('id');
     } else if ($element.attr('name') != null) {
-      id = $element.attr('name').replace(/\[|\]/g, '')
+      id = $element.attr('name').replace(/\[|\]/g, '');
       id =  id + '-' + Utils.generateChars(2);
     } else {
       id = Utils.generateChars(4);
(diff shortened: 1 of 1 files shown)


src/js/select2/core.js
```diff
@@ -94,7 +94,8 @@ define([
     if ($element.attr('id') != null) {
       id = $element.attr('id');
     } else if ($element.attr('name') != null) {
-      id = $element.attr('name') + '-' + Utils.generateChars(2);
+      id = $element.attr('name').replace(/\[|\]/g, '');
```
kevin-brown: jQuery [recommends](https://learn.jquery.com/using-jquery-core/faq/how-do-i-select-an-element-by-an-id-that-has-characters-used-in-css-notation/) using the following regex

```js
/(:|\.|\[|\]|,)/g
```

As it will handle a couple of other characters that can also be valid selectors.

kevin-brown: We also might want to apply this to the `id` case above, as it also has the same issue (though less common).

bnjmnhndrsn:
Amended to use the recommended regex and apply the logic to ids generated from `id` also.

kevin-brown:
Looks like this should be set to be pulled in for 4.0.2. I'll merge this in once I create some tests, but until then this looks fine.

PR #4127, opened 2016-01-27
Replaced protocol-relative URLs with HTTPS protocol

drmonkeyninja:
Use of protocol-relative URLs is now seen as an anti-pattern as it leaves the door open to attacks. As the CDNs serve over HTTPS it should be recommended that people use HTTPS rather than protocol-relative URLs. See Paul Irish's post on [Protocol-relative URLs](http://www.paulirish.com/2010/the-protocol-relative-url/) for more details.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -113,7 +113,7 @@ slug: home
     <p>
       Select2 is hosted on both the
       <a href="https://cdnjs.com/libraries/select2">cdnjs</a> and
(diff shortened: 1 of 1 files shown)


kevin-brown:
This will be merged in for 4.0.2, I have it milestoned so I can do a quick sweep of other areas in the documentation to make sure this is consistent.

drmonkeyninja:
:+1:

PR #4130, opened 2016-01-28
Update defaults.js

bensampaio:
Extend options recursively so that if there are default options for ajax they are applied.


kevin-brown:
I can review this patch for 4.0.2 and get it merged, but I can't confirm that this is actually an issue unless we can get a failing test case in place that can be fixed by this change. If you would be willing to create a test case ([here's a basic example](https://github.com/select2/select2/blob/97f7eea9ceefd1ec86737e57a65b2551cc40d504/tests/options/translation-tests.js)), we can speed up that process and ensure that it will definitely land for 4.0.2.

bensampaio:
I created a test, where shall I place it? Also how can I run it? I never used QUnit and I'm also not so familiar with the select2 folder structure yet, so some hints on where to place the test and how to run it would be super helpful. Thanks!

kevin-brown:
> I created a test, where shall I place it?

In the `tests` folder, under the `options` directory. I'm thinking in a new `ajax.js` file, given that your test is probably specific to the `ajax` options (but can likely be generalized). You will also need to add a reference to the file in `unit.html`, so it will be added to the unit tests that we run.

> Also how can I run it?

Once you [get your local environment set up](https://groups.google.com/d/msg/select2/IFHylwxz1Hw/4GzSXlwnCAAJ), you can just run the tests using `grunt` by calling `grunt test`. You can also just load up the pages (`unit.html`) directly in your browser and they _should_ work, assuming you have compiled your changes.

bensampaio:
Thanks! It's done, I added the test and confirmed that it only passes when my changes are present.

PR #4138, opened 2016-02-01
Rtl patch

moshfeu:


kevin-brown:
How is this any different from the RTL support that comes built into Select2?

https://select2.github.io/examples.html#rtl

moshfeu:
I don'e understand how I missed this..

PR #4139, opened 2016-02-01
Added Greek (el) translation

masterakado:
Translated English (en) language file to Greek (el)


mutil:
This looks good and handles pluralization.
I can confirm that translations are correct.
Can you please merge?

PR #4146, opened 2016-02-03
Adding Greek and Slovenian translations.

rogercoder:


kevin-brown:
Thanks for your contribution to Select2, @rogercoder.

We have another pull request that is open which adds translations for the Greek language at https://github.com/select2/select2/pull/4139. I, unfortunately, am not familiar enough with the Greek language to be able to properly judge which pull request should be accepted, so I'm not sure which one I should look into merging.

Most of the translations have minor differences (I'm not sure which one is more accepted) and it appears that your translations do not actually handle pluralization.

rogercoder:
>> it appears that your translations do not actually handle pluralization.

Yep, in the process we typically use with our translation service we just
get things translated as "item(s)" rather than trying to satisfy the
sometimes tricky logic of plurality used in different languages.

If it makes things easier for you I can submit a different pull request
with only the Slovenian translations.

-Roger

PR #4155, opened 2016-02-05
reuse result of previous calculation

ZuBB:


protorp:
please, write how use it

PR #4193, opened 2016-02-16
Redirect the focus of original input to Select2

fabdouglas:
Redirect the focus of original input to the `focus` event of Select2.
this way, we get back the accessible label for feature of previous Select2 versions.

The focus feature worked well in my previous version (3.4.8) without addition pieces of code
It's clearly a regression.
Le commit introducing the regression is there : https://github.com/select2/select2/commit/db7cd1b373f9a2a743a68ce32cc59dc8fc899eb8, and is due to the `input` becoming hidden instead of out of screen. An hidden input cannot get focus.

This is the second version of #2588
Tested on 3.5 head.

### DIFF
--- a/select2.js
+++ b/select2.js
@@ -2297,9 +2297,10 @@ the specific language governing permissions and limitations under the Apache Lic
             this.search.attr("id", this.focusser.attr('id') + '_search');
 
             this.search.prev()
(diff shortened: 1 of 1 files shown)


kevin-brown:

Issue #4191, opened 2016-02-15
Ajax call on empty input


blaaat:
I want to suggest the following change in data/ajax.js line 91. 
`if (this.ajaxOptions.delay && params.term !== '') {
`

params.term is undefined when empty. So I would like to change this to: 
`if (this.ajaxOptions.delay && params.term !== undefined && params.term !== '') {
`

So an empty search field loads initial results fast as soon as an user opens the select2 box without addition delay. The params.term !== '' seems to have this intention as well.

WilsonChiang:
Shouldn't actually check on an empty string I believe.

See:
https://github.com/select2/select2/pull/4202

PR #4202, opened 2016-02-22
Fix delay not being applied to empty search term

WilsonChiang:
This pull request includes a

- [x] Bug fix

There is an issue where if the search term is rapidly removed (By holding down the backspace key for example), an incorrect request will be sent at the end containing the first letter of the search term. This is due to the delay being ignored if an empty input is submitted. The input is originally `undefined` so I have changed it to that, which also fixes this bug.

![Demonstration of bug]
(http://i.imgur.com/MDimCHj.png)

### DIFF
--- a/src/js/select2/data/ajax.js
+++ b/src/js/select2/data/ajax.js
@@ -88,7 +88,7 @@ define([
       self._request = $request;
     }
 
-    if (this.ajaxOptions.delay && params.term !== '') {
+    if (this.ajaxOptions.delay && params.term !== undefined) {
       if (this._queryTimeout) {
         window.clearTimeout(this._queryTimeout);
       }


smajl:
Related: https://github.com/select2/select2/issues/4191
Suggests: `if (this.ajaxOptions.delay && params.term !== undefined && params.term !== '') {`

src/js/select2/data/ajax.js
```diff
@@ -88,7 +88,7 @@ define([
       self._request = $request;
     }
 
-    if (this.ajaxOptions.delay && params.term !== '') {
+    if (this.ajaxOptions.delay && params.term !== undefined) {
```
kevin-brown: This should include `''` as well as `undefined`.

WilsonChiang: What's the use case for `''`? The input box is initialized as `undefined`, so avoiding the delay on initiation seems to be the desire here. Avoiding it on empty string causes the issue I mentioned originally.

WilsonChiang: @kevin-brown Any thoughts? :)

kevin-brown: I believe the goal of checking for `''` was to handle the case when someone completely empties the search box (deletes everything). Assuming that _is_ when this case was previously triggered, it's something we're going to need to maintain.

PR #4212, opened 2016-02-25
Fix #3320 Prevent select2:open when clearing selection

### LINKED ISSUE #3320: Prevent select2:open when clearing selection
Prevent catching the click (and disallow opening of the select or triggering the `select2::open`) when the clear indicator is pressed to clear the selection when `allowClear` is `true`.

This behavior worked fine in 3.5.2 - not sure if there is a reason for this. 

Use case: When `allowClear` is `true` and one clicks the selection clear indicator -- the value gets cleared but it also triggers the dropdown to open again when it is not necessary.

marsantony:
This pull request includes a
- [x] Bug fix

- Prevent select2:open when clearing selection

https://github.com/select2/select2/issues/3320

### DIFF
--- a/src/js/select2/selection/allowClear.js
+++ b/src/js/select2/selection/allowClear.js
@@ -62,7 +62,6 @@ define([
 
     this.$element.val(this.placeholder.id).trigger('change');
 
-    this.trigger('toggle', {});
   };
 
   AllowClear.prototype._handleKeyboardClear = function (_, evt, container) {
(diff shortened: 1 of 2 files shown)


comwes:
Dear, 
As a company, we are having complains from our customers regarding this issue.
We have tested and it is fixing the reported issue.
Can you please accept this pull request, I think it might be helpful not only for us, but for many other users.
KR,
Gerry

Issue #4203, opened 2016-02-23
Select2 looses focus when enclosed in a label tag


stretch4x4:
## Steps to reproduce the issue

1. Create a select2 inside a label tag (see fiddle: https://jsfiddle.net/2kt7z8wu/)
2. When you click on the select2 dropdown, the search gets focus before loosing it as the hidden select gets the focus.

## Expected behavior and actual behavior

When I click on a select I expect it to retain focus.
We found this because we have replaced standard selects in our django admin screens which are wrapped in labels and they don't retain focus correctly. Without a lot of extra work we can't change the html so would prefer to fix select2.

## Environment

Browsers

- [x] Google Chrome
- [x] Mozilla Firefox
- [x] Internet Explorer

Operating System

- [x] Windows
- [ ] Mac OS X
- [ ] Linux
- [ ] Mobile

Libraries

- jQuery version: 4.0.1
- Select2 version: 2.2.0

## Isolating the problem

- [ ] This bug happens [on the examples page](https://select2.github.io/examples.html)
- [x] The bug happens consistently across all tested browsers
- [x] This bug happens when using Select2 without other pluigns
- [x] I can reproduce this bug in [a jsfiddle](https://jsfiddle.net/2kt7z8wu/)

Delagen:
refer to: #4187

stretch4x4:
That is a solution to the problem, but I think it does need to be integrated into select2. Because you don't have to do that on a normal select to make it work correctly.
Now that I have an idea on how it could be solved I can take a look at where in select2 this might be fixable and see if I can come up with a pull request..

PR #4235, opened 2016-03-07
Fixing focus issue with wrapping labels

stretch4x4:


stretch4x4:
Not sure if this is the best place for this code but it does resolve the issue I am seeing with labels in #4203

kevin-brown:
Now that we've determined that the issue is that a `click` event is being generated on the label, why is the `click` event being generated when Select2 is opened?

This seems like it _might be_ a band-aid over a larger issue.

stretch4x4:
Actually that is a really good point, I assumed it was because they were nested but now that I think about it that doesn't make a lot of sense..
I will see if I can trace the click and find some more info.

stretch4x4:
Turns out it is being generated for a standard one: https://jsfiddle.net/2kt7z8wu/3/
I think the issue is that the same behaviour happens and the hidden select gets focus..

Issue #3479, opened 2015-06-15
Usability issue in infinite scrolling


pranav2410:
i have select2 with infinite scrolling,
the problem is whenever select2 fetches more results from remote data-source the highlight is shifted to the first element in the list instead of highlighting the next element in list
for repo : on your examples page 
1) try Loading remote data example
2) enter some keyword
3) use down arrow key to load more data and go to last element so that it will load more data
4) the focus goes to first element of drop down after loading new data

pranav2410:
any update on this?

AlexanderLeonov:
I ran into the same problem. We currently use version 3.5 and we have plans to migrate to version 4, but in order to do that it must support "keyboard-only" model well, so this behavior "as is" will not work for us.

So I did some investigation and it appears to me that the problem is with setClasses function. It does not follow "separation of concerns" principle: instead of just setting classes on newly loaded items like its name suggests it does also change highlighted item (lines after $options.each() cycle). This leads me to three points:
- Changing highlighted item is definitely a separate concert and it should not happen here, regardless of considering anything else;
- "results:append" event handler (through setClasses() call or by any other means) should not change highlighted item in dropdown at all. It should still call setClasses() though to update "aria-selected" on any newly loaded items if they've already been selected before current data loading;
- For performance reasons it should probably go through only newly loaded items, not all of them, cause previously loaded data had already been processed on previous data loads and user actions. Even if personally I didn't see any performance problems yet it doesn't mean that it should not be optimized.

So my conclusion is that setClasses() definitely should be rewritten.

Kevin - what do you think? I can prepare pull request to fix it if it works for you.

kevin-brown:
> Kevin - what do you think? I can prepare pull request to fix it if it works for you.

If you can think of a better way to handle `setClasses`, I'm all for it. I'll gladly review any pull requests that are sent in.

PR #4238, opened 2016-03-07
Fixed infinite scrolling issue using down key.

AlexanderLeonov:
This pull request includes a

- [x] Bug fix
- [ ] New feature
- [ ] Translation

- setClasses has been split into two functions - focusFirstItem and setClasses itself;
- calls to focusFirstItem have been added after all but one calls to setClasses (not added in "results:append" event handler) - I just don't feel I've explored this code well enough so I preserved previous behavior as much as I could.

This is related to issue https://github.com/select2/select2/issues/3479.

### DIFF
--- a/src/js/select2/results.js
+++ b/src/js/select2/results.js
@@ -97,6 +97,25 @@ define([
     return sorter(data);
   };
 
+  Results.prototype.focusFirstItem = function () {
+
+    var $options = this.$results
+      .find('.select2-results__option[aria-selected]');
+
+    var $selected = $options.filter('[aria-selected=true]');
+
+    // Check if there are any selected options
+    if ($selected.length > 0) {
+      // If there are selected options, highlight the first
+      $selected.first().trigger('mouseenter');
+    } else {
+      // If there are no selected options, highlight the first option
+      // in the dropdown
+      $options.first().trigger('mouseenter');
+    }
+
+  };
+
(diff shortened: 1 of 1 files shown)


kevin-brown:
This is set to land for 4.0.3, but before it can land I'm aiming to get some tests in place so we don't have to fix this in the future. If you're interested in writing up those tests, that would definitely speed up the process.

AlexanderLeonov:
Sure, I'm totally up to it. I just need short guide from you on what you think is right to test here, cause I don't see whole picture in the project, what is tested, what is not, what needs to be tested, what doesn't.

kevin-brown:
> cause I don't see whole picture in the project, what is tested, what is not, what needs to be tested, what doesn't.

_Ideally_, everything is tested. In reality, I only have a small amount of time and I write tests for things which I know were previously an issue.

Unfortunately we don't currently have any tests against the `select2/results` module, so there is going to be a bit of overhead in getting that set up.

> I just need short guide from you on what you think is right to test here

At the very least, the first test I would write would be for the `results:all` event. It would check that the `mouseenter` event was properly triggered and that the highlight class was present.

The second test (more relevant to infinite scrolling) would be to trigger the `results:append` event and check that the `mouseenter` event _wasn't_ triggered, and that the highlight class hasn't moved.

PR #4240, opened 2016-03-08
Fixed preventing override scrolling event while using Select2 with ajax request

justiceform:
This pull request includes a

- [x] Bug fix

Problem story
  I'm using **ajax Select2** with Bootstrap3 Modal with scrolling modal. so when select2's dropdown opened, Bootstrap modal can't scroll.

- change this to self 
-
-

PS. this commit from @mahadoang's idea, all credit to him

### DIFF
--- a/src/js/select2/dropdown/attachBody.js
+++ b/src/js/select2/dropdown/attachBody.js
@@ -98,7 +98,7 @@ define([
 
     $watchers.on(scrollEvent, function (ev) {
       var position = $(this).data('select2-scroll-position');
-      $(this).scrollTop(position.y);
+      $(self).scrollTop(position.y);
     });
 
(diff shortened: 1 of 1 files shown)


src/js/select2/dropdown/attachBody.js
```diff
@@ -98,7 +98,7 @@ define([
 
     $watchers.on(scrollEvent, function (ev) {
       var position = $(this).data('select2-scroll-position');
-      $(this).scrollTop(position.y);
+      $(self).scrollTop(position.y);
```
kevin-brown: `self` is a reference to the `AttachBody` object, not a jQuery/DOM element.

kevin-brown:
See my in-line comment on why this doesn't fix whatever issue you are seeing.

Please [submit a ticket](https://github.com/select2/select2/issues/new) (instead of a pull request) about this issue, assuming one doesn't already exist.

PR #4249, opened 2016-03-10
Fix #4248

### LINKED ISSUE #4248: Adding a selected option should update container text

1. Instanciate select2 on `<select><option value="x">y</option></select>`
2. Append `<option value="bar" selected>foo</option>` to non-multiple select

(Or use add-another popup in django admin).

Expected: value updates to "bar", container text updates to "foo"
Actual: value updates to "bar" but container text is still "y".

## Environment

Browsers

- [ ] Google Chrome
- [ ] Mozilla Firefox
- [ ] Internet Explorer

Operating System

- [ ] Linux

Libraries

- jQuery version: vendorized from select2 master
- Select2 version: master

## Isolating the problem

A test case is provided in a PR.

jpic:
This pull request includes a

- [x] Bug fix

- Added test case
- Observe subtree for added options and sync the container text

Issue: #4248

### DIFF
--- a/tests/integration/select2-methods.js
+++ b/tests/integration/select2-methods.js
@@ -1,3 +1,4 @@
+/* globals start, stop, setTimeout */
 module('select2(data)');
 
 var $ = require('jquery');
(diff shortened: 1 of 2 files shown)


jpic:
The new test fails as expected, still trying to fix it, but there might be a design decision to make so I'm not really sure. Should _syncAttributes be extended to support this ? should a new _sync method be added ? should we reuse the same observer and enable subtree: true on it for performance or make another observer ?

jpic:
Ready for review B)

tests/integration/select2-methods.js
```diff
@@ -1,3 +1,4 @@
+/* globals start, stop, setTimeout */
```
jpic: Sorry, but apparently we don't have [assert.async()](https://api.qunitjs.com/async/) in our version of QUnit !

kevin-brown:
I've mentioned it on other channels (IRC, specifically) that this pull request has issues because it bypasses the formatting of selections. This is because instead of triggering `selection:update` with the new selection, it just sets the selection directly. This also only works on single selects, partially because that's the easiest to design for, but it results in this being inconsistent.

As a result, I'm finally going to formally reject this pull request. Tonight I'll take a shot at doing this in a way that works with the Select2 event system instead of bypassing it.

jpic:
Perhaps you could reuse the test case though ?

PR #4251, opened 2016-03-11
Meteor package

rgnevashev:
This pull request includes a

- [ ] Bug fix
- [ ] New feature
- [ ] Translation

-
-
-

### DIFF
--- a/.versions
+++ b/.versions
@@ -1,0 +1,4 @@
+jquery@1.11.4
+meteor@1.1.10
+rgnevashev:select2@4.0.2
+underscore@1.0.4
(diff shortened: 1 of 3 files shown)


kevin-brown:
Thank you for contributing to Select2!

If there are no changes required for Select2 to work with Meteor, we would rather not release another package just for Meteor. Note that we do it for Bower and a few other managers because of historical reasons.

If there are changes required to make Select2 work with Meteor, we recommend creating a separate third-party package that contains that code and links to Select2. You can see a few examples of that in the [integrations section of the README](https://github.com/select2/select2/blob/6844be777b0fe9af1d2b9a8f6051b7eb7d532247/README.md#integrations).

PR #4253, opened 2016-03-14
Documentation Update: Select2 with a Bootstrap Modal

rohankapoorcom:
This pull request includes a

- [ ] Bug fix
- [ ] New feature
- [ ] Translation

- Update documentation to include how to use Select2 inside a Bootstrap Modal.

### DIFF
--- a/docs/_includes/options/dropdown/placement.html
+++ b/docs/_includes/options/dropdown/placement.html
@@ -25,6 +25,21 @@ $('select').select2({
     I&apos;m using a Bootstrap modal and I can&apos;t use the search box
   </h3>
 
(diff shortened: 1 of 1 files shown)


719media:
I'm not sure exactly what issue you are trying to solve here, but I assume it's how the select2 becomes "unresponsive" when in a bootstrap modal.

I solved it by simply changing the dropdownparent select2 option
```
$newProductModal = $('#modal_id');
$newProductModal.on('shown.bs.modal', function () {
    //use dropdownParent, as this is in a modal
    $(this).find('select[name="style_id"]').select2({
        dropdownParent : $newProductModal
    });
});
```

Issue #3303, opened 2015-05-01
Dropdown body is positioned incorrectly when dropdownParent isn't statically positioned


nchen63:
Steps to reproduce:
1. navigate to http://jsfiddle.net/Leppndoo/
2. click on dropdown

issue: the dropdown body does not appear below the control. 

The control is located at top:100px;left:100px but the dropdown body is located at top:228px;left:200px

nchen63:
```<body>
    <div style="height:500px; width:500px; overflow:auto; display:block; background-color:red"> 
            <span class="asdf" style="height:3000px; display:block; left:100px;top:100px;position:relative">
                <select class="js-example-basic-single">
                      <option value="AL">Alabama</option>
                      <option value="WY">Wyoming</option>
                </select>
            </span>
    </div>
</body>
```

```
$(document).ready(function () {
    var parentElement = $(".asdf");
    $(".js-example-basic-single").select2({
                dropdownParent: parentElement
            });
    
});
```

MilosR:
The problem is because one of, or more, parent containers have position other than static (relative, absolute), and in script dropdown top & left values are calculated with regard to document body.

Line 124 of attachBody.js
    var offset = this.$container.offset();

kevin-brown:
_This is probably why Select2 3.5 was special casing static positioning..._

If we can find a way that works for determining the offset which doesn't require special casing, that'd be awesome. But I might be willing to review a pull request that does handle special casing, as long as we can get tests alongside of it.

wesgarland:
I had the same issue, but MilosR's fix did not work for me.  It moved the dropdown position to the top left of the parent container, which is not correct.  I am trying to insert my stuff pretty deep into the document, to play nice with a bunch of existing code.

Unfortunately, I am not a jQuery guy so my fix is clunky, but it appears to work for my case, on Firefox-latest and Chrome-latest.  I appended this to the end of AttachBody.prototype._positionDropDown, right after the call to this.$dropdownContainer.css:

    var rect = this.$dropdownParent.getBoundingClientRect();
    this.$dropdown[0].style.left = -rect.left + "px";
    this.$dropdown[0].style.top = -rect.top + "px";

That's line 4095 of select2.full.js for version 4.0.0.

afattahi54:
May be this help ;)
Removing the scroll bar of parent div, cause the div drop down shown in different
![Uploading 12.png…]()
![Uploading 13.png…]()
![Uploading 14.png…]()

[...]

PR #4267, opened 2016-03-23
Fixing bug with non-static parents for dropdowns

rcousart:
This pull request includes a

- [X] Bug fix

- Added `parentOffset` to the computation of the `css.top` when direction is above

This fix adds the offset to the case where the dropdown direction is above as seen and resolved in #3303


kevin-brown:
I can test this out, but it would be useful if we could also get a unit test that covers this (similar to the [existing ones](https://github.com/select2/select2/blob/6844be777b0fe9af1d2b9a8f6051b7eb7d532247/tests/dropdown/positioning-tests.js)) so we can ensure that this bug doesn't come back in the future.

rcousart:
@kevin-brown good idea, will add that shortly

rcousart:
I believe this logic triggers only when there is a class in the `span` block inside of the select container that specifies the direction of the dropdown. I added the class `select2-dropdown--below` and `select2-dropdown--above` and ran the current test that checks for offset. 

Can anyone take a look and see if this is actually doing what I want it to? I would love some advice on how to test this better to ensure this doesn't break in future changes.

lvlalbert:
I'm having a problem when the dropdown is above and not below. The offsets do not get reset when the dropdown closes or something along those lines. Does this fix that? I tried this by replacing the lines in select2.js but it didn't work. Here is an image detailing my issue. 
https://i.imgur.com/8BaP5ND.jpg
Any ideas?
Thanks. This is my first time posting on GitHub.

kevin-brown:
I have committed https://github.com/select2/select2/pull/4267/commits/edee884abe4246871afb1731250f1aa9fc0dc845 as https://github.com/select2/select2/commit/ecdc41663508f0a5ea242fe249aa5a9a3a741ba6. I did not include the tests, because after playing around with them it hit me that they do not actually fail without the change. I did modify the existing tests in https://github.com/select2/select2/commit/98f054fc18d4eabe104153cbab0bc6447456d310 to check that the classes are automatically being applied to the dropdown.

Thanks for the initial pull request to fix this issue. You should see it fixed in the upcoming 4.0.3 release.

Loadmaster:
I had the same problem, where the drop-down was being positioned too far down (too large a `top` value) on the page when the Select2 element was contained within a table `<td>` element.

In my situation, the problem was that the (non-`static`) parent's position was being added to the drop-down's `top` offset, when it should _not_ have been.

See #4382 for details and my suggested code fix.

PR #4271, opened 2016-03-28
Added the document.ready call to the multiple example section. Necess…

kurtaking:
This pull request includes a BUG FIX on:
https://select2.github.io/examples.html

I noticed on the examples page that the the basic example includes the document.ready() function necessary to make it work. However, on the "multiple" example, it does not. When I first came to select2 to use the tool, I was not very familiar with Javascript and it took me way too long to realize that was the one line I was missing. I made the changes and submitted a pull request because I think it is necessary to help users who may not be as familiar with JavaScript. 

Before: 

<script type="text/javascript">
$(".js-example-basic-multiple").select2();
</script>

After:

<script type="text/javascript">
$(document).ready(function() {
$(".js-example-basic-multiple").select2();
});
</script>

This was not related to a specific ticket. It is something I personally noticed and wanting to change.

### DIFF
--- a/docs/_includes/examples/basics.html
+++ b/docs/_includes/examples/basics.html
@@ -50,7 +50,9 @@ $(document).ready(function() {
 
 {% highlight html linenos %}
 <script type="text/javascript">
-$(".js-example-basic-multiple").select2();
+$(document).ready(function() {
+  $(".js-example-basic-multiple").select2();
+});
 </script>
 
 <select class="js-example-basic-multiple" multiple="multiple">


alexweissman:
Actually, when I look at other jQuery plugins' documentation, they _do_ seem to include the `$(document).ready` wrapper or something similar.  Take a look at:

- https://mottie.github.io/tablesorter/docs/#Getting-Started
- http://eonasdan.github.io/bootstrap-datetimepicker/#minimum-setup

Maybe we should include this after all.

kurtaking:
@alexweissman Now that I have a few years of Javascript experience under my belt, I laugh at the fact that I needed help with this. However, I still agree and think it would be good to have for people who are starting out.

alexweissman:
Haha yep, sometimes it's harder to see where you came from, once you're on the other side.

I've updated the documentation: https://select2.org/getting-started/basic-usage

PR #4291, opened 2016-04-07
fix for formatResults not appending to container

sdsalamida:
This pull request includes a

- [ ] Bug fix

- move node.append(label) outside of the formattedResult undefined evaluation

Proposed fix for issue where results generated in user-defined formatResult function are not appearing in the select2 results.

Scenario: formatResult is being used with container parameter. A result is appended to the container but does not appear in the results.
Solution: node.append needs to be called regardless of formatResult return value


kevin-brown:
We are no longer accepting pull requests for Select2 3.x. We have no intentions of releasing a Select2 3.5.5, and it is recommended to update to Select2 4.0 when you have the chance.

This pull request would revert a change added in https://github.com/select2/select2/pull/1733.

Issue #4068, opened 2016-01-06
Up/Down arrow keys not working when JAWS-17 is open


ShawnXu:
Here is a screencast http://screencast.com/t/ROFpa3ZhGyd recorded in Chrome with JAWS 17 open on https://select2.github.io/examples.html

Our QA reported that when JAWS is on, the Up/Down arrow keys can fail to navigate.  Pressing the Down arrow key or TAB key sometimes results in "document end" according to JAWS read, which may have something to do with the fact that select2's option list is appended to the bottom of the page. Not sure if there could a a fix on select2's side. 

I'm also bring this issue up to JAWS support since it's possible that it's their software intercepting and interrupting those key strokes.

Thanks,
Shawn

ShawnXu:
Hi, I wonder if select2 dev had the bandwidth to look at this issue yet?

kevin-brown:
As with most accessibility issues, we don't have the tools necessary to be able to quickly debug them. So any debugging comes during the short periods of time where we have access to JAWS (or other screen readers, when they are properly configured), assuming we don't have other things that take higher priority.

My _guess_ is that JAWS is intercepting the keys before they reach Select2, but I have no idea what can be done to prevent that (on Select2's side, at least).

krtek4:
Hi There,

Currently having the same issue with Jaws 17 on Windows 10 with IE 11.

What I remarked is that even with Jaws 17, the first example on the example page https://select2.github.io/examples.html#single does not work at all with the keyboard if it was selected using "tab".

Steps to reproduce :

1. Put the focus on the "real select"
2. Press Tab so the focus goes to the select2 version
3. Press Up / Down

Nothing is selected and the page just scroll around. Arrow key selection seems to work fine withtout Jaws when the select2 is opened using the mouse, but fails with Jaws, so this seems to be a "compound" issue.

vlad-saling:
Hey, we are also experiencing this issue. Symptom and steps to reproduce match. I'll talk to our devs if we can give Select2 team a hand with this. Although I am not exactly sure where to begin.

alexweissman:
Has anyone been able to confirm if v4.0.4 resolves this issue?  #4427 fixed a major focus issue.

Issue #3472, opened 2015-06-12
Arrow keys do not change value of select


CUnknown:
According to https://github.com/select2/select2/issues/3279 the feature of opening the select2 search dropdown is not implemented in the newer version according to technical difficulties, but also because it does not reflect the default behaviour of a native select.

Though, following this native select behaviour, the user should be able to browse through the dropdown items by using the arrow keys on the keyboard. This does not seem to work for the select2 4.0 implementation. Is there a reason for this?

AlexBuitrago:
+1000

rain01:
yes please

sidharthc:
Is there a solution for this?

uzlov:
this.on('keypress', function (evt) {
      var key = evt.which;

      if (self.isOpen()) {
        if (key === KEYS.ENTER) {
          self.trigger('results:select');

          evt.preventDefault();
        } else if ((key === KEYS.SPACE && evt.ctrlKey)) {
          self.trigger('results:toggle');

          evt.preventDefault();
        } else if (key === KEYS.UP) {
          self.trigger('results:previous');

          evt.preventDefault();
        } else if (key === KEYS.DOWN) {
          self.trigger('results:next');

          evt.preventDefault();
        } else if (key === KEYS.ESC || key === KEYS.TAB) {
          self.close();

          evt.preventDefault();
        }
      } else {
        if (key === KEYS.ENTER || key === KEYS.SPACE ||
            ((key === KEYS.DOWN || key === KEYS.UP) && evt.altKey)) {
          self.open();

          evt.preventDefault();
        }
        if (key === KEYS.DOWN) {
          this.$element.find('option:selected').next().attr('selected', 'selected').trigger('change');
          evt.preventDefault();
        }
        if (key === KEYS.UP) {
          this.$element.find('option:selected').prev().attr('selected', 'selected').trigger('change');
          evt.preventDefault();
        }
      }
    });

rain01:
@uzlov what is this? where do you put it?

sidharthc:
@uzlov seems good.

uzlov:
@rain01
it's a patch for select2
please, find function this.on('keypress', function (evt) {
and replace with my code

alvarotrigo:
+1

CUnknown:
@uzlov could you submit a PR?

alvarotrigo:
[...]

PR #4293, opened 2016-04-07
Allow the use of arrow keys when select2 is closed

krtek4:
This pull request includes a
- [ ] Bug fix
- [x] New feature
- [ ] Translation

- Allow the use of arrow keys when select2 is closed
- Set a 'aria-live' attribute so that SR read the new item

This is related to #3472 and #4068.

This commit add the possibilities to cycle through the
possible options even when select2 is closed.

I also added a 'aria-live' attribute on change so that
a screen reader will read the newly selected item upon
change.

### DIFF
--- a/src/js/select2/core.js
+++ b/src/js/select2/core.js
@@ -329,7 +329,9 @@ define([
           evt.preventDefault();
         } else if (key === KEYS.DOWN || key === KEYS.UP) {
             var options = this.$element.find('option:selected');
-            options = key === KEYS.DOWN ? options.prevAll(":enabled") : options.prevAll(":enabled");
+            options = key === KEYS.DOWN ? 
+                options.nextAll(':enabled') :
+                options.prevAll(':enabled');
 
             var val = options.first.val();
             if (undefined !== val) {
--- a/src/js/select2/core.js
+++ b/src/js/select2/core.js
@@ -327,6 +327,16 @@ define([
           self.open();
 
           evt.preventDefault();
+        } else if (key === KEYS.DOWN || key === KEYS.UP) {
+            var options = this.$element.find('option:selected');
+            options = key === KEYS.DOWN ? options.prevAll(":enabled") : options.prevAll(":enabled");
+
+            var val = options.first.val();
+            if (undefined !== val) {
+                this.$element.val(val);
+                this.$element.trigger('change');
+            }
+            evt.preventDefault();
         }
       }
     });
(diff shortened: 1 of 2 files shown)


Lashas83:
Please fix quotes and formatting to pass the build

krtek4:
Thanks for the update !

The tests are now green :)

src/js/select2/core.js
```diff
@@ -327,6 +327,18 @@ define([
           self.open();
 
           evt.preventDefault();
+        } else if (key === KEYS.DOWN || key === KEYS.UP) {
+            var options = this.$element.find('option:selected');
+            options = key === KEYS.DOWN ? 
+                options.nextAll(':enabled') :
+                options.prevAll(':enabled');
+
+            var val = options.first.val();
```
fdintino: I believe this should be `options.first().val()`

alitbutt: You might not like to change the selected multiple items.
That would make line 330:
} else if ((key === KEYS.DOWN || key === KEYS.UP) && (!(self.options.get('multiple'))) ) {

jibranb:
Hello. Are the up / down arrow keys supposed to work by default at this point? Mine are not and I'm not finding an API option to enable them. Using latest (v4.0.5). Sorry if this has been answered, but I'm not finding a solution. If I should post this somewhere else please let me know. Thanks.

alexweissman:
@jibranb it looks like this PR was never merged.  So, no, this feature is not yet available.

Since it is a new feature, and not a bug fix or translation, I'll add it to the 4.1 (minor version) milestone.

jibranb:
@alexweissman thanks for the update. I did a quick search of the source (v4.0.5) it appears there is code to handle arrow keys. But it's not firing due to a requirement to press the alt key (see /dist/js/select.js [line 5395](https://github.com/select2/select2/blob/master/dist/js/select2.js#L5395)). I've removed that requirement for now. Not crazy about editing the core, but it's working for now.

Is there a way to override the 'keypress' event ([line 5368](https://github.com/select2/select2/blob/master/dist/js/select2.js#L5368))?

```
        if (key === KEYS.ENTER || key === KEYS.SPACE ||
            (key === KEYS.DOWN **&& evt.altKey**)) {
          self.open();

          evt.preventDefault();
        }
```
Thanks.

alexweissman:
@jibranb ah, thanks for catching that!  I didn't realize that this was already implemented.  Any idea where this was discussed originally?

I'm going to link to the [source code](https://github.com/select2/select2/blob/062c6c3af5f0f39794c34c0a343a3857e587cc97/src/js/select2/core.js#L346-L351) instead of the distribution version that you linked to, for future reference.

Seems like this should also be documented.

Issue #3616, opened 2015-07-27
ajax.templateSelection() is called twice for every selection event


renehamburger:
This can be reproduced with the ajax example given at https://select2.github.io/examples.html and a breakpoint in `formatRepoSelection()`.

francescomeloni:
someone can suggest a workaround for this bug?

pjryan126:
My suggested workaround:

```
var bool = true; // boolean switch to handle select2 bug
function templateSelection(selection) {
     bool = !bool;
     if (bool == true) {
          // format and return formatted selection here
     }
}
```

smoothlikejazz:
Love this Library, I also ran into this; So in revisiting and looking through the code there is an event called "selection:update"; the listener is registered twice, 
Once in  SingleSelection.prototype.bind and again in BaseSelection.prototype.bind
```
container.on('selection:update', function (params) {
      self.update(params.data);
});
```
Now since its registered in the Base prototype method it should be safe to remove the second binding from the SingleSelection.prototype.bind method. 

In testing by removing the second binding it will only fire once as you would expect in the ajax case. I am not sure of the Ramifications of doing this though, maybe author can chime in.

kevin-brown:
@smoothlikejazz That should only need to be registered once, and I don't believe the ordering matters.

Feel free to submit a pull request which drops it from the SingleSelection.

smoothlikejazz:
Okay great thanks @kevin-brown will do

joanebrown:
@smoothlikejazz  Hi! I still have this problem with the Select2 4.0.3. Your PR is not merged? In my case, the `templateSelection` its called four times...

*BTW, who knows what's the diferrence between "select2.full" and "select2"?*

Regards!!!

joanebrown:
@kevin-brown There is any date for the 4.1.0 release? There are some workaround for this problem until the new release isnt yet ready?

CodyBrouwers:
I am getting `templateSelection` called four times as well. Removing this listener brings it down to twice at least but is it going to get merged soon?

Angelo8828:
+1 here. Experiencing the same problem.

Angelo8828:
As a workaround for this bug. My work is like this.

`
// DECLARE A DELIMITER VARIABLE
var delimiter = 1;

$('select').select2({

........

templateResult: function (data) {
delimiter = 1;
....... // FORMATTING OF SEARCH QUERY RESULTS
},
templateSelection: function (data) {

[...]

PR #4306, opened 2016-04-13
Removing the double event binding registration of "selection:update" …

smoothlikejazz:
This pull request includes a
- [x] Bug fix
- [ ] New feature
- [ ] Translation

- Removing the double registration of the listener binding on "selection:update" within SingleSelection.prototype.bind, as it already exists in BaseSelection.prototype.

https://github.com/select2/select2/issues/3616


alexweissman:

Issue #1699, opened 2013-09-08
chrome incorrect html5 validation tooltip position issue 


RusAlex:
``` css
position: absolute;
```

in .select2-offscreen now  broke google-chrome html5 built-in validation messages.

Chrome is trying to attach tooltip to the original "select" tag. But it's position now at the top left window corner. And tooltip is in incorrect place.

PR #4311, opened 2016-04-15
HTML5 required validation

fabdouglas:
This pull request includes a bug fix for #128, #3065, #3069 and #1699.
The CSS will place the feedback validation exactly at the right place
The JS will ensure the required of original input (and also hidden) is moved to the created select2 input. This one is not hidden, but sized to 1px, corresponds to the input styled by the fixed CSS, and will transfert the focus to the Select2 required input.
Tested on IE11, FF and Chrome.

### DIFF
--- a/select2.css
+++ b/select2.css
@@ -649,7 +649,7 @@ html[dir="rtl"] .select2-container-multi .select2-search-choice-close {
 
 .select2-offscreen, .select2-offscreen:focus {
     clip: rect(0 0 0 0) !important;
-    width: 1px !important;
+    width: 100% !important;
     height: 1px !important;
     border: 0 !important;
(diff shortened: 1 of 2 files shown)


fabdouglas:
Also support now single and multiple selection.
For multiple, "required" implies : at least one choice

fabdouglas:
For #3065 the jsfiddle : https://jsfiddle.net/u0ztnxyj/

kevin-brown:
The 3.5.x branch of Select2 is no longer being maintained, and as a result we are closing all pull requests made towards `stable/3.5.x`. We highly recommend upgrading to the 4.0 branch of Select2 if possible.

PR #4324, opened 2016-04-25
Ability to hide currently selected value in dropdown as default option

davilious:
This pull request includes a
- [ ] Bug fix
- [x] New feature
- [ ] Translation

- Adds a default option to be able to hide those elements that have been selected
- Prevents navigation by getting proper options, those with `[aria-selected=false]` when `hideOptionsWhenSelected: true` or  `[aria-selected]` for the default option `hideOptionsWhenSelected: false`

This issue was discussed in several issues, main issue listed below: 
- https://github.com/select2/select2/issues/1538

I hope it helps.

### DIFF
--- a/src/js/select2/defaults.js
+++ b/src/js/select2/defaults.js
@@ -375,7 +375,8 @@ define([
         return selection.text;
       },
       theme: 'default',
-      width: 'resolve'
+      width: 'resolve',
+      hideOptionsWhenSelected: false
     };
   };
 
--- a/src/js/select2/results.js
+++ b/src/js/select2/results.js
@@ -237,6 +237,13 @@ define([
 
     var id = container.id + '-results';
 
+    var getOptions = function() {
+      if (self.options.get('hideOptionsWhenSelected')) {
+        return self.$results.find('[aria-selected=false]');
+      }
+      return self.$results.find('[aria-selected]');
+    };
+
     this.$results.attr('id', id);
 
     container.on('results:all', function (params) {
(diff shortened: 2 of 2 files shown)


zatuliveter:
Waiting for the feature!
Thanks!

guill3m:
Still waiting for this feature. @kevin-brown are there any plans to merge this PR? Thank you.

techno808:
Any news on this, it would be really useful.

kevin-brown:
This is basically a JavaScript equivalent for the CSS snippet included in this comment: https://github.com/select2/select2/issues/1538#issuecomment-170757783

I'd rather see this done in CSS personally, especially because it doesn't require adding an additional option.

davilious:
Thanks for your response @kevin-brown. I don't think it is enough by hiding those elements selected with CSS because the element is still there and it is selectable, for example, when moving up/down using key arrows. 

Lemme elaborate it. We need to hide those selected options on the fly, as long as they are selected after triggering `results:previous` event, the CSS solution is good when the Dom is just loaded but not after.

Below you can see an example where there are some options, one of them is selected by default and it works but, try it to reproduce the ussue by following the steps:

[Select2 Issue#4324 demo](http://codepen.io/davilious/pen/d8bff22a34783527925d8df94091c9b6/)

1) Type "joanito" and select "joanito";
2) Type "jo" and move until the very last item "joan";
3) You will see that it doesn't flow well, going from "jonshua" to nothing and then to "joan";

It happens because "joanito" is hidden using CSS but Select2 still treats it as an existing element.

Does it make any sense?  What do you think?

zatuliveter:
Completely agree with @davilious comment.

davilious:
Hey @kevin-brown, do you think this pull request could be re-open considering [my arguments above](https://github.com/select2/select2/pull/4324#issuecomment-283018478)? Thanks in advance.

Issue #3359, opened 2015-05-14
Using tabs as tokenSeparators for tags 


doctorallen:
I've tried so many different things to try and get select2 to allow for the tab key as a token separator to start a new tag, to no avail. Does anyone know how to do this?

mofman:
+1

Kaldrax:
+1
Right now Tab even deletes the input which really sucks.

Hemanth089:
+1

superKalo:
+1
Would be very useful!

doctorallen:
I hate to mention this here, but for those of you that need this feature, you may want to look at [selectize](https://brianreavis.github.io/selectize.js/).

althaus:
+1

grantgeorge:
+1

simsketch:
+1

sampatbadhe:
:+1:

sanosay:
(Y)

andreasba:
Hello there,

I really like select2 to thanks a lot for the great work @kevin-brown !

Now I am a bit confused about the status of this issue - is it closed and resolved or just closed? It would be great to have a solution to this.

Thanks
Andreas

rycwilson:
+1

griesi007:
+1

gbeauvoir:
+1
Add also ENTER key press

bbotella:
+1

Atyz:
+1

kcivey:
:+1:

anhtuank7c:
:+1:

cireficc:
👍

knightq:
👍 Any new?

davilious:
I sent a pull request a few weeks ago, I don't know yet whether or not it will be included in future versions. For those of you interested, here the [PR](https://github.com/select2/select2/pull/4325)

carldanley:
:+1:

LuciusCaesar:
+1

bkroeker:
+1

DiesIrae:
And Enter, pleeease! :-)

jaketoolson:
Just found this, hoping its still relevant.

jransegnola:
+1 to this.

Was just trying to find my own way around this.

jeremy-hunter:
kevin-brown's solution worked for me:
https://github.com/select2/select2/pull/4325#issuecomment-229234605
selectOnClose creates the tag on both tab and enter

[...]

PR #4325, opened 2016-04-25
Uses tabs for creating tags as a default feature

davilious:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

- Allows the use of tabs to create tags by triggering the proper events, `results:select`

https://github.com/select2/select2/issues/3359#issuecomment-214272255

I believe previous versions of Select2 added it as a default feature without the need of setting it via options. 

I run the tests and everything seems to be fine.

I hope it helps.

### DIFF
--- a/src/js/select2/core.js
+++ b/src/js/select2/core.js
@@ -299,16 +299,15 @@ define([
       var key = evt.which;
 
       if (self.isOpen()) {
-        if (key === KEYS.ESC || key === KEYS.TAB ||
-            (key === KEYS.UP && evt.altKey)) {
+        if (key === KEYS.ESC || (key === KEYS.UP && evt.altKey)) {
           self.close();
 
           evt.preventDefault();
-        } else if (key === KEYS.ENTER) {
+        } else if (key === KEYS.ENTER || key === KEYS.TAB) {
           self.trigger('results:select', {});
 
           evt.preventDefault();
-        } else if ((key === KEYS.SPACE && evt.ctrlKey)) {
+        } else if (key === KEYS.SPACE && evt.ctrlKey) {
           self.trigger('results:toggle', {});
 
           evt.preventDefault();


Jan-Kruse:
Works perfectly smooth, thanks a lot :balloon::tada::confetti_ball:

drush:
+1

mlorant:
@davilious While the PR is not merged/released, do you have a built version of select2 with this patch or is there any way to do the same by overriding events handlers in my own code?

davilious:
Hey @mlorant, what I did was to include the change in the uncompressed version **leaving a comment** to fix it when the merge happens. [Here a fiddle](https://jsfiddle.net/bs3wctwg/) where you will find the context but make sure you have the **same version (4.0.2)** and **replace it** when it is merged or new versions of this change take place.

carldanley:
Any update on having this PR added to the next release?

kevin-brown:
Just a note: This won't be landing as a default feature because it's a breaking change.

My standard recommendation is to look into `selectOnClose` to see if it's what you are actually looking for.

drush:
@kevin-brown selectOnClose addresses this for my use case, thanks for referencing that. I hadn't seen it before.

davilious:
@kevin-brown thx for the reference, I wasn't aware there was an option already which works perfectly fine. Closing this PR.

davilious:
Just for the records. The `selectOnClose` option is not enough for my needs because you can't leave the field blank, it is a little annoying indeed. Below a demo, try to delete the tag set.

[Select2 - issue#4325 demo](http://codepen.io/davilious/pen/542c85c91008201f995b686be16de559?editors=1010)

However, I understand my pull-request might be a little drastic, people may want to use the "Tab" key to close the dropdown instead of selecting one option.

SatanEnglish:
Same Issue here even using the clear for the dropdown selects the first item

rob101:
+1

PR #4329, opened 2016-04-27
Add dist i18n js for es-MX

jtapia:
This pull request includes a

- [ ] Bug fix
- [ ] New feature
- [x] Translation

- Add es-MX js file


simi:
Is there any difference compared to https://github.com/select2/select2/blob/master/src/js/select2/i18n/es.js? I wasn't able to find any.

kevin-brown:
This appears to be the same as `es.js`, which is what will be used if `es-MX` is specified as the language. Select2 will build a fallback chain, and that is the second one (out of three) in the chain.

jtapia:
I think that approach is not the correct one, I mean, we have the same languague 'spanish' but we have different idioms, so maybe you should add those different files for each spanish languaue country

jtapia:
@simi on this case we use the same words, but maybe on Colombia they don't, so we should include all of those files

simi:
That's ok to add new translation, but I wasn't able to find any difference between current es.js and your submited es-MX.js. It is useless to have same file twice in repository and thanks to fallbacks, es-MX language will use es.js translation file.

jtapia:
I started looking into an issue inside `spree` admin, `http://localhost:3000/javascripts/select2_locale_es-MX.js` and at the end it took me here on `select2`, and that's because you're matching any `es-` locale to `:es` and on this case is wrong because `es-MX` is the correct locale, not `:es`

PR #4337, opened 2016-05-02
Added the right single quotation mark to diactritics.

charettes:
This pull request includes a
- Bug fix

- Added the `RIGHT SINGLE QUOTATION MARK` character to diacritics

It should be simplified to a typewriter apostrophe: https://en.wikipedia.org/wiki/Apostrophe#Unicode

I could also add all the apostrophe variants but I feel this should be the most common one as [this is the preferred character to use for apostrophe according to the Unicode standard](http://unicode.org/Public/UNIDATA/NamesList.txt).


alexweissman:
Failing some jshint checks:

```
   src/js/[secure]/diacritics.js

    844 |    '\u2019': "'"

                          ^ Strings must use singlequote.

   src/js/[secure]/i18n/lt.js

      2 |  // rules from http://www.unicode.org/cldr/charts/latest/supplemental/language_plural_rules.html#lt

                                                                                                            ^ Line is too long.

      6 |    } else if ((count % 10 >= 2 && count % 10 <= 9) && (count % 100 < 11 || count % 100 > 19)) {

                                                                                                        ^ Line is too long.

>> 3 errors in 92 files

Warning: Task "jshint:code" failed. Use --force to continue.
```

charettes:
@alexweissman just pushed an adjusted commit.

alexweissman:
Oh, awesome, thanks!

PR #4339, opened 2016-05-03
reference the optgroup in the option's data-object

murb:
To prevent parent()-lookups on the element, reference the optgroup directly in the option’s data-object when iterating the select elements children.

This pull request includes a

- [ x ] New feature

- the item function now registers the `optgroup` as an attribute of the option, which  
- the label is added as a data-attribute to the listed option (maybe an unwanted extra?)
- test is added.

### DIFF
--- a/src/js/select2/data/select.js
+++ b/src/js/select2/data/select.js
@@ -197,7 +197,7 @@ define([
     return $option;
   };
 
-  SelectAdapter.prototype.item = function ($option) {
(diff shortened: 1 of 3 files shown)


murb:
Closing this pull request; this is not the right approach to things (and didn't even work as expected after testing in real work). Doing a few parent() calls on the option-elements isn't that bad https://select2.github.io/options.html#templateSelection

Issue #2997, opened 2015-02-01
Firefox Mobile: Search Field (with and without ajax) does nothing


dominikzogg:
I can search within Chrome (Android) but it i try it within Firefox (Android) i get the search mask on click, be able to enter chars, but it sees, like they get completely ignored. Any idea?

Android 5.0.2, FF 35.0

kevin-brown:
I can reproduce this for the search input, only when entering searchable characters. If you hit "enter", it will select the highlighted input.

Interestingly, the back button doesn't try to remove the last option in a multiple select.

dominikzogg:
@kevin-brown cause i need to type (using ajax)

kevin-brown:
Still debugging this issue, hoping to track down exactly why it is happening before the final release of 4.0.

dominikzogg:
@kevin-brown cool, thx

dominikzogg:
@kevin-brown thank you

Issue #3417, opened 2015-05-27
In IE9 backspace in select menu does not refresh search results


cdaveb:
I'm using Select2 4.0.0. In Internet Explorer 9 (using Windows 7), if you type in the select menu it successfully limits the items to those matching the text, but if you then use backspace to delete characters, it doesn't refresh the search results. It will only refresh the search results if you exit the menu and reopen it, or type something new.

This behavior does not exist in the other browsers I've tested it in (such as Chrome Mac, IE8 or IE10).

Here are examples of the issue using the Select2 website examples page:
![step1](https://cloud.githubusercontent.com/assets/6133808/7848072/6c46b20a-047d-11e5-9a85-136ebd243a32.png)
![step2](https://cloud.githubusercontent.com/assets/6133808/7848073/71d94be2-047d-11e5-807a-1894326fcbb0.png)
![step3](https://cloud.githubusercontent.com/assets/6133808/7848078/75f0707a-047d-11e5-912a-11b43aed35fe.png)

kevin-brown:
I was able to reproduce this in IE 9 compatibility mode.

acetrand:
+1 Same here. Tested on Win7 IE9.

Note that Select 3.5.2 works.

acetrand:
I did some more investigation on this and the source for it seems to be that it listens for `keyup` or `ìnput` which isn't triggered in IE9 for backspace.

I haven't looked if Microsoft has this documented somewhere but I found a thread on a similar issue: http://stackoverflow.com/questions/6382389/oninput-in-ie9-doesnt-fire-when-we-hit-backspace-del-do-cut

theelectricwiz:
+1. Just tried removing the input event handling altogether (leaving in keyup) and it fixes the issue. What is the benefit of using the input event over the keyup event in this instance?

theelectricwiz:
For completeness, the input event is required to handle the special input characteristics of touchscreen devices, in particular firefox on these devices (#2997). The bug is only on IE9 (<IE9 doesn't support oninput and it is fixed in >IE9) (tested with https://jsfiddle.net/uxewb7pw/)

theelectricwiz:
In my second attempt, I have created a special case just for IE9 which fixes the problem https://github.com/theelectricwiz/select2/commit/c74f51be75458c8e21df58767ed350c241706f3a#diff-bb1b322a1846b8b86b3741d853a1f0ee

`src/js/select2/dropdown/search.js 37`
```javascript
[...]

PR #4340, opened 2016-05-03
Remove the input event handling

theelectricwiz:
This pull request includes a
- bugfix for #3417 

- Removing input event handling
- An incidental minor version bump in almond (0.3.1 -> 0.3.2)


kevin-brown:
Unfortunately the `input` event handling needs to stay. It was added to fix https://github.com/select2/select2/issues/2997.

theelectricwiz:
Ah. Ain't browser differences great. Thanks Kevin. Will have a think about it.

Issue #4014, opened 2015-12-11
Stack overflow is triggered when .data() is called in the options


Hacktivista:
I don't know if it's a Select2 problem, or a Parsley problem. I couldn't identify the error :(

Here is the simplest fiddle which shows the problem:
http://jsfiddle.net/jew1mg89/1/

The error is not present using the previous version of Select2 (3.x)

vin-zhou:
Hi guys, I faced the same problem these days. Any progress ? @kevin-brown @Hacktivista

Hacktivista:
@vin-zhou As workaround you can initialize parsley after loading select2.

Eventually somebody will be able to fix the problem. I couldn't :(

chesterx:
same here :(

kevin-brown:
Here's an updated version of the jsfiddle that works out of the box, without any changes, to reproduce the issue that you appear to be seeing.

http://jsfiddle.net/jew1mg89/5/

kevin-brown:
Tracking this down a little further, the issue is triggered within `select2/option`, which is what parses the configuration for Select2. Specifically in [`fromElement`](https://github.com/select2/select2/blob/09ecd3cdb96da5d62f498ac0f7f1bf0bb6e54621/src/js/select2/options.js#L26-L111), where we parse out any data from the `data-*` attributes on the element.

You can reproduce the error without Select2 by trying to clone the data for an element using [`$.extend`](https://api.jquery.com/jquery.extend/), but only with the [`deep`](https://api.jquery.com/jquery.extend/#jQuery-extend-deep-target-object1-objectN) option specified.

http://jsfiddle.net/jew1mg89/8/

I expected this to be an issue in Select2 as well, but apparently Select2 can be deep copied without issue.

http://jsfiddle.net/jew1mg89/9/

So that officially makes this a Parsely issue, triggered by Select2.

marcandre:
I feel that `select2` should not assume anything about `data()` that isn't his. Since `jquery`'s deep extend isn't safe on recursive structure, then it shouldn't be used on objects that aren't `select2`.

Can `select2` write it's own deep clone function that is recursion safe? I'll provide a PR if needed.

marcandre:
Note: I didn't check why `select2` wants to do a deep copy of `data()`. Maybe that's not actually needed, a shallow copy could suffice?

kevin-brown:
> Maybe that's not actually needed, a shallow copy could suffice?

[...]

PR #4346, opened 2016-05-06
Cache objects in Utils.__cache instead of using $.data (#4014)

NadeemAfana:
This pull request includes a
- [x] Bug fix
- [ ] New feature
- [ ] Translation

- Cache objects in `Utils.__cache` instead of using $.data 
- Updated several files to use the new functions to cache data in `Utils.__cache`
-  Removed all `$.data` code 

This is a bug fix for #4014


kevin-brown:
There are indentation changes in here that are going to prevent this from being merged in.

NadeemAfana:
I will fix the indentation.

kevin-brown:
- There are still some indentation issues
- This is going to need integration tests to make sure that we don't have this issue in the future

NadeemAfana:
Sorry for the indentation issues. I fixed the EOL to be UNIX format. My editor had switched this to Windows format.

Do you want me to add integration tests ?

alexweissman:
There's an integration test now...can this be merged in?

alexweissman:

oh-ren:
Please also consider the following discussion regarding `$myOriginalSelectElement.data('select2')`: https://github.com/select2/select2/issues/5071#issuecomment-340218988

chrisdeeming:
Sorry for posting here, but it feels relevant and I've tried to get support from the official channels with no answers yet.

Although there’s a fallback in `Utils.GetData` to get from the data attribute if the cache is empty, there’s nothing that actually sets the data attribute in the first place.

What I’m trying to do is access what was equivalent to:

```js
var instance = $element.data('select2');
```

But this no longer seems to be possible.

Is there a public way of accessing the `GetData` or `SetData` functions?

Is this not a sort of dangerous change to make in essentially a bug fix release in terms of breaking backwards compatibility?

alexweissman:
I'm here, just been busy with php[world] and traveling.

chrisdeeming:
To be fair, I've not exactly been waiting long for a reply (and it's a weekend!), but I didn't necessarily want to see this one slip into a stable release if there's something wrong.

alexweissman:
Yes, thanks @oh-ren and @chrisdeeming.  I didn't realize that these changes affected the public API.

But, it seems like this is meant to fix a bug (#4014).  Honestly, I don't really understand the problem that this is solving as I have never encountered this myself.  It seemed like a safe change, but apparently not.

Is there a way we can fix this bug (whatever it may be) without causing a breaking change?  Join me on my own chat server (https://chat.userfrosting.com/channel/select2) and we can discuss.

alexweissman:
@NadeemAfana any thoughts on this?  Seems like this is breaking a lot of other aspects of Select2.

NadeemAfana:
`$element.data('select2')` is clearly not public API; it is an internal implementation which people should avoid. Nonetheless, I will push a fix for this.

alexweissman:
Thanks!  Yes, unfortunately Select2 is so big and widely used, and had been poorly documented for so long, and tends to attract the greenest types of developers, that it's part of the public API whether we want it to be or not 😦 

Are there any use cases where people really _need_ to be able to access `data('select2')`?  Otherwise we can add something to the documentation and then deprecate it over time.

Also, is there any chance that this modification is what is causing #5133?

NadeemAfana:
I would definitely deprecate `.data('select2')`. It makes things confusing when people start using the internal components as they change over time. For example, updating `.data('select2').options.options.maximumSelectionLength` directly has no effect. According to [this comment,](https://github.com/select2/select2/issues/5071#issue-263367289)  it worked in the past, but this change has nothing to do with this PR.

> Also, is there any chance that this modification is what is causing #5133?

No, because I was able to reproduce the issue with `4.0.5`.

chrisdeeming:
We use the data stored there specifically to access the elements which are dynamically created by Select2, mostly as a shortcut.

They are, specifically `.select2-container` and `.select2-selection`.

We could just find them in the DOM relative to element we've instantiated Select2 on but if they're already cached in the Select2 object then, why not?

Point taken about invalid use cases but that's not something we'd do.

oh-ren:
I suggest to create a new method then, perhaps `$mySelect.select2('getInstance')`?
To reiterate *my* use case (there might be others of course) here: add a class to `$container` when the select is deemed `invalid` (for styling purposes).

Not sure how other (popular) jQuery plugins implement getting a reference / access to the internal instance. It's not something uncommon, or is it?
Anyway it provides a means to get things done that otherwise are impossible / the plugin doesn't cater for. Manually selecting based on classes (or manual DOM traversal) is like 10000 times worse than being able to access the relevant elements already constructed by Select2 itself.

(Off topic, but somewhat related: `$('#mySelect2').hasClass("select2-hidden-accessible")` is not that great, why not something like: `$('#mySelect2').select2('isInitialized')`?

@alexweissman Thanks for the invite, might check the channel out, but simply discussing here works best?

chrisdeeming:
@oh-ren I would indeed say that something similar to `$mySelect.select2('getInstance')` is generally the norm in other libraries.

That said, the data attribute approach is not uncommon either. As a compromise, if there was a built in method which could get certain things (but not set them) then that would work too. `$mySelect.select2('getInstance')` could just get the bits that are considered "safe" to expose.

alexweissman:
Keep in mind also, that modifications to the internals of Select2 should be possible with custom adapters/decorators.  

Can you provide examples of other jQuery plugins that have a `getInstance` method?  

Incidentally, this seems like a perfect example of how Javascript's lack of visibility scoping can be frustrating.  Perhaps we need a rewrite of Select2 into [Typescript](https://www.typescriptlang.org/docs/handbook/classes.html#public-private-and-protected-modifiers) more badly than we realize 😁 

At any rate, I've merged in @NadeemAfana's BC fix for release on `4.0.6-rc.1`, so this should solve the problem for the time being.

NadeemAfana:
The tests now run for both jQuery 1.x and 2.x. This allows us to see if there is any discrepancy in the future.

NadeemAfana:
Re-opening.

jamessaget:
Hello,

I have read through issue #5071 which lead me here. What I am trying to achieve is change a config option after initialisation and stumbled into the issue stated here https://stackoverflow.com/questions/37016655/select2-cannot-read-property-query-of-null regarding the error thrown within the dataAdapter.

```javascript
$('.country_select').on('select2:opening', (e) => {
   if($(e.target).data('select2').options.options.minimumResultsForSearch == 5) return
   $(e.target).data('select2').options.options.minimumResultsForSearch = 5;
   let options = $(e.target).data('select2').options.options;
   $(target).select2('destroy');
   $(target).select2(options);
})
```

Any suggestions would be appreciated

PR #4348, opened 2016-05-10
Stories/af/fix main aria roles properties

afercia:
This pull request includes a
- [x] Bug fix
- [ ] New feature
- [ ] Translation

- fixes ARIA roles for the selection element and the search field
- fixes ARIA roles for the results list
- fixes ARIA properties `aria-owns`, `aria-activedescendant`, and `aria-autocomplete`
- fixes ARIA role for the `aria-live` message

See #3735

As mentioned in #3735 making the dropdown options being announced by screen reader will require several steps. Will comment with more details on the issue.

### DIFF
--- a/src/js/select2/dropdown/search.js
+++ b/src/js/select2/dropdown/search.js
@@ -9,9 +9,9 @@ define([
 
     var $search = $(
       '<span class="select2-search select2-search--dropdown">' +
-        '<input class="select2-search__field" type="search" tabindex="-1"' +
+        '<input class="select2-search__field" type="text" tabindex="-1"' +
         ' autocomplete="off" autocorrect="off" autocapitalize="off"' +
-        ' spellcheck="false" role="textbox" />' +
+        ' spellcheck="false" role="combobox" aria-autocomplete="list" />' +
       '</span>'
     );
 
@@ -25,6 +25,7 @@ define([
 
   Search.prototype.bind = function (decorated, container, $container) {
     var self = this;
+    var resultsId = container.id + '-results';
 
     decorated.call(this, container, $container);
 
@@ -48,7 +49,7 @@ define([
 
     container.on('open', function () {
       self.$search.attr('tabindex', 0);
-
+      self.$search.attr('aria-owns', resultsId);
       self.$search.focus();
 
       window.setTimeout(function () {
@@ -58,7 +59,8 @@ define([
 
     container.on('close', function () {
       self.$search.attr('tabindex', -1);
-
+      self.$search.removeAttr('aria-activedescendant');
+      self.$search.removeAttr('aria-owns');
       self.$search.val('');
     });
 
@@ -72,6 +74,10 @@ define([
           self.$searchContainer.addClass('select2-search--hide');
         }
       }
+    });
+
+    container.on('results:focus', function (params) {
+      self.$search.attr('aria-activedescendant', params.data._resultId);
     });
   };
 
--- a/src/js/select2/results.js
+++ b/src/js/select2/results.js
@@ -37,7 +37,7 @@ define([
     this.hideLoading();
 
     var $message = $(
-      '<li role="option" aria-live="assertive"' +
+      '<li role="alert" aria-live="assertive"' +
       ' class="select2-results__option"></li>'
     );
 
--- a/src/js/select2/results.js
+++ b/src/js/select2/results.js
@@ -14,7 +14,7 @@ define([
 
   Results.prototype.render = function () {
     var $results = $(
-      '<ul class="select2-results__options" role="tree"></ul>'
+      '<ul class="select2-results__options" role="listbox"></ul>'
     );
 
     if (this.options.get('multiple')) {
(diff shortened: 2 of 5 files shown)


src/js/select2/dropdown/search.js
```diff
@@ -9,9 +9,9 @@ define([
 
     var $search = $(
       '<span class="select2-search select2-search--dropdown">' +
-        '<input class="select2-search__field" type="search" tabindex="-1"' +
+        '<input class="select2-search__field" type="text" tabindex="-1"' +
```
kevin-brown: The rest of the changes make sense, but I can't figure out why this change was needed. Can a search field not function as a combobox?

src/js/select2/results.js
```diff
@@ -14,7 +14,7 @@ define([
 
   Results.prototype.render = function () {
     var $results = $(
-      '<ul class="select2-results__options" role="tree"></ul>'
+      '<ul class="select2-results__options" role="listbox"></ul>'
```
kevin-brown: How well does this work with a nested [`group` role](https://www.w3.org/TR/wai-aria/roles#group)?

Part of the reason why `tree` was picked was because it supported nesting of options. From what we could tell, `listbox` only supports a single level (completely flat).

src/js/select2/dropdown/search.js
afercia: Well, the semantics of this input field is completely overridden by the applied ARIA role. Technologies that understand ARIA will get it is a "combobox" (also, currently has a role="textbox") so the type attribute is not so much relevant.
About browsers, the only difference with a text box I'm aware of is webkit special styling which Select2 resets with CSS

src/js/select2/results.js
afercia: A "tree" widget is something different, from the spec:
> A type of list that may contain sub-level nested groups _that can be collapsed and expanded_
The ARIA authoring practices describe the interaction screen readers should implement with ARIA widgets, I think using a Tree View is one of the main reasons the Select2 options are not announced. See version 1.0 and 1.1 (both working drafts):
https://www.w3.org/TR/wai-aria-practices/#TreeView
https://www.w3.org/TR/wai-aria-practices-1.1/#TreeView

On the other hand, the Combobox widget explicitly requires a listbox with options, see:
https://www.w3.org/TR/wai-aria-practices/#combobox
https://www.w3.org/TR/wai-aria-practices-1.1/#combobox

I'm afraid ARIA doesn't provide a way to emulate the semantics of `optgroup` or at least I wasn't able to find it. Nested groups would require further investigation and iteration. That's why I'm proposing to proceed with small steps and then iterate, if you agree :)

Testing the PR, only Firefox+NVDA announce the group "labels", IE11+JAWS skip the labels but do announce the options, Safari+VoiceOver announce just the "flat" ones.

kevin-brown:
I want to thank Paal Joachim Romdahl for reminding me about this pull request, not because I forgot about it (I've been discussing this with a few others for a while now), but because I never publicly commented about the plans for Select2 to handle this.

---

As has been mentioned a couple of times before, Select2 supports a wide variety of use cases, some of which are natively supported by screen readers and others not so much. Our goal with Select2 4.0.0 was to make Select2 natively accessible and to _do it right the first time_, so we didn't have to hack it in later. Part of that involved researching existing implementations, looking into any recent improvements to ARIA, and doing some testing of our own along the way. As we've clearly learned after the release of Select2 4.0.0, we didn't get it right the first time, we actually _got it very wrong_. We accept that, and hopefully we can get Select2 to a state where it is at least partially acceptable, so the most basic use cases work natively. From there, we hope to work to get the more advanced use cases supported.

In the next couple of weeks, we plan to work with others to get Select2 to be accessible under the most basic use case: a flat dropdown that is searchable. Since that appears to be the use case that most screen readers support.

michalkleiner:
Related PR #4881.

src/js/select2/results.js
fuzzbomb: > I'm afraid ARIA doesn't provide a way to emulate the semantics of optgroup
Also note that plain old HTML optgroups don't get conveyed to screen readers either.

afercia: For reference, in ARIA 1.2 `role=group` will be allowed as child of `role=listbox` to emulate `optgroup`. See https://github.com/w3c/aria/pull/782

michalkleiner:
Any further updates on this topic?

kevin-brown:
Alright, time to kick this again since it got stale.

We dropped the ball on this PR. We've continued to drop the ball throughout the years. This PR is still valid and fixes a valid problem. Even though it has a merge conflict now, it's one of the ones we're willing to correct the merge conflict on and manually apply the changes back once they're reviewed.

michalkleiner:
Bump again.

kevin-brown:
I've started to pull the changes here out into https://github.com/select2/select2/pull/5582 where I'm adding some tests and fixing the merge conflicts.

kevin-brown:
Closing this off in favor of https://github.com/select2/select2/pull/5582.

The only thing implemented here that is not implemented in that PR is the handling of dropdowns without search boxes. I am going to spin off a new ticket about it, but my initial thinking is that Select2 should behave like a `listbox` when it does not have a searchbox, and like a `combobox` when it does have a search box.

PR #4356, opened 2016-05-11
Fix AJAX data source error

### LINKED ISSUE #4355: AJAX data source "the results could not be loaded" error
## Prerequisites
1. Navigate to: https://select2.github.io/examples.html#data-ajax
2. Input text
3. Before the ajax results are loaded, input more text

When I follow those steps, I see...

"The results could not be loaded." is displayed (then, after a while, the results are populated)

I was expecting...

"Searching.." displayed as more text is inputted
## Environment

Browsers
- [x] Google Chrome
- [x] Mozilla Firefox
- [ ] Internet Explorer

Operating System
- [x] Windows
- [x] Mac OS X
- [x] Linux
- [ ] Mobile

Libraries
- jQuery version: v1.11.2
- Select2 version: master (v4.0.2#master)
## Isolating the problem
- [x] I can reproduce this bug in [a jsbin](https://jsbin.com/)

MichaelMackus:
This pull request includes a
- [x] Bug fix

- Check for existence of "status" in the $request object (instead of truthfulness)
- Check that $request.status === 0 (number) not '0' (string)

Fixes #4355

Verified in Firefox & Chromium

### DIFF
--- a/dist/js/select2.full.js
+++ b/dist/js/select2.full.js
@@ -3444,7 +3444,7 @@ S2.define('select2/data/ajax',[
       }, function () {
         // Attempt to detect if a request was aborted
         // Only works if the transport exposes a status property
-        if ($request.status && $request.status === '0') {
+        if ('status' in $request && $request.status === 0) {
           return;
         }
 
--- a/dist/js/select2.full.js
+++ b/dist/js/select2.full.js
@@ -3444,7 +3444,7 @@ S2.define('select2/data/ajax',[
       }, function () {
         // Attempt to detect if a request was aborted
         // Only works if the transport exposes a status property
-        if ('status' in $request && $request.status === 0) {
+        if ($request.status && $request.status === '0') {
           return;
         }
 
--- a/dist/js/select2.js
+++ b/dist/js/select2.js
@@ -3444,7 +3444,7 @@ S2.define('select2/data/ajax',[
       }, function () {
         // Attempt to detect if a request was aborted
         // Only works if the transport exposes a status property
-        if ($request.status && $request.status === '0') {
+        if ('status' in $request && $request.status === 0) {
           return;
         }
 
--- a/dist/js/select2.js
+++ b/dist/js/select2.js
@@ -3444,7 +3444,7 @@ S2.define('select2/data/ajax',[
       }, function () {
         // Attempt to detect if a request was aborted
         // Only works if the transport exposes a status property
-        if ('status' in $request && $request.status === 0) {
+        if ($request.status && $request.status === '0') {
           return;
         }
 
--- a/src/js/select2/data/ajax.js
+++ b/src/js/select2/data/ajax.js
@@ -82,7 +82,8 @@ define([
       }, function () {
         // Attempt to detect if a request was aborted
         // Only works if the transport exposes a status property
-        if ('status' in $request && $request.status === 0) {
+        if ('status' in $request &&
+            ($request.status === 0 || $request.status === '0')) {
           return;
         }
 
--- a/src/js/select2/data/ajax.js
+++ b/src/js/select2/data/ajax.js
@@ -82,7 +82,7 @@ define([
       }, function () {
         // Attempt to detect if a request was aborted
         // Only works if the transport exposes a status property
-        if ($request.status && $request.status === '0') {
+        if ('status' in $request && $request.status === 0) {
           return;
         }
 


PowerKiKi:
I believe this the solution to a bug that is affecting all installation of select2. As such, shouldn't it be released in a patch version (4.0.4), rather than a minor version (4.1.0) ?

It seems that including this for 4.1 is only delaying the fix without any other benefits. I don't think it's either a breaking change or a new feature.

kevin-brown:
@PowerKiKi the next planned release of Select2 is 4.1.0, we have no intentions of rolling a 4.0.4 release due to (outside) time constraints.

PowerKiKi:
Not quite what I wanted to hear, but thanks for taking the time to answer anyway :+1:

drmmr763:
Currently experiencing bug #4355. Would love to see this PR merged or the next release rolled out so I can upgrade. Probably going to apply this patch anyway for the time being. But +1 this PR!

MichaelMackus:
:+1: for a new minor version increase. This is a bugfix affecting most/all modern browsers!

PowerKiKi:
It seems that select2 lacks a proper workflow for (semi-)continuous delivery. It's a pity because bugs like these impact a lot of users, yet are very simple and safe to fix. Especially when the community do half the work to fix it.

vladkazantsev:
Any comment on when this will be released? Time frames? Experiencing the 4355 across all devices...

dist/js/select2.full.js
```diff
@@ -3444,7 +3444,7 @@ S2.define('select2/data/ajax',[
       }, function () {
         // Attempt to detect if a request was aborted
         // Only works if the transport exposes a status property
-        if ($request.status && $request.status === '0') {
+        if ('status' in $request && $request.status === 0) {
```
jpic: Is this not going to break support for all browsers that have '0' ? shouldnt this be `$request.status === 0 || $request.status === '0'`?

dist/js/select2.js
```diff
@@ -3444,7 +3444,7 @@ S2.define('select2/data/ajax',[
       }, function () {
         // Attempt to detect if a request was aborted
         // Only works if the transport exposes a status property
-        if ($request.status && $request.status === '0') {
+        if ('status' in $request && $request.status === 0) {
```
jpic: Don't change dist files, for more info check CONTRIBUTING

src/js/select2/data/ajax.js
```diff
@@ -82,7 +82,7 @@ define([
       }, function () {
         // Attempt to detect if a request was aborted
         // Only works if the transport exposes a status property
-        if ($request.status && $request.status === '0') {
+        if ('status' in $request && $request.status === 0) {
```
jpic: Same question.

jpic:
We'll need regression tests before we can merge this.

MichaelMackus:
@jpic done & submitted changes. I'm not familiar with JS testing, but will look into if I find some free time. If somebody else could take on the tests I'd be grateful.

alexweissman:
Seems like a pretty popular fix - can someone who knows their way around the testing framework please write some tests so we can get this released in the next revision?

lordplagus02:
Just reminding everyone that this exists and it's been over a year

alexweissman:
@lordplagus02 it will be released in 4.0.6.  Hold yer horses 🐴

alexweissman:

dist/js/select2.full.js
M4cker: For those who are using timeout : if ('status' in $request && $request.status === 0 && $request.statusText != "timeout")

MichielDP:
This bug is **STILL** happening in version 4.0.6-rc.1. I'm getting the error "the results could not be loaded" with even the most basic usage of select2. The request is not even reaching my backend method. When I paste the exact same url in the browser url bar, then it does reach my backend.

markoheijnen:
Any idea when this is going to be released.

pedrofurtado:
@markoheijnen In next weeks. I am reviewing some another PRs. After it, we will release a version.

smuszel:
Hey. Are you sure this solution indeed fixes the issue. At my company we are still having issues with it. Changing function at 3589 of select.js to below fixed the problem.

```
function(_, statusText) {
    if (statusText !== 'abort') {
        self.trigger('results:message', {
            message: 'errorLoading',
        });
    }
},
```

PR #4360, opened 2016-05-15
Fixes #4353. Does not jump the view to recently destroyed dropdown co…

### LINKED ISSUE #4353: Window jumps when dropdown is closed
This is reproducible on any Select2 v4 instance, [for example](https://select2.github.io/examples.html):
- Click to open the example select2 box under "Single select boxes"
- Leave the dropdown open and scroll down, so that the dropdown is out of view
- Click anywhere on the page
- You get jumped back up to the select box, and the element you clicked on doesn't fire 

(I used Chrome 50, but I think this is the case everywhere)

I've seen other tickets about clicking outside dropdown, but none that pertain to v4 and have a solution.

I think it's not obvious to the user that leaving the dropdown open will cause issues for you down the page. I can think of a couple possible solutions:
- Show a background mask on everything but the select2 box and dropdown
- Allow clicks outside the dropdown to propagate normally
- Lock scrolling while the dropdown is open (like a normal <select> element does)

prakal:
This pull request includes a
- [X] Bug fix
- [ ] New feature
- [ ] Translation

- Remove function call to focus view on destroyed dropdown upon clicking in another part of the screen.

Existing ticket #4353

### DIFF
--- a/src/js/select2/selection/base.js
+++ b/src/js/select2/selection/base.js
@@ -81,8 +81,6 @@ define([
       self.$selection.removeAttr('aria-activedescendant');
       self.$selection.removeAttr('aria-owns');
 
-      self.$selection.focus();
-
       self._detachCloseHandler(container);
     });
 


alexweissman:
I'm suspicious of any PR that deletes one line to fix a problem.  Any idea why that line was there in the first place?  We can't risk breaking other things here...

davilious:
I don't think either it is a great solution, this line is needed to focus on the `$selection` once the dropdown is closed so it can get the style to let the user know it is the `select` that just changed. 

#### Just before selection

captura de pantalla 2018-11-05 a las 17 20 16

#### Just after selection
captura de pantalla 2018-11-05 a las 17 17 02


Instead of removing that line I would set a timer to delay the focus on that element. I need to investigate it more, why this delay is needed, maybe the element is not fully present or it is hidden somehow in dom at that particular moment but it works for me and does not jump anymore.

```javascript
    container.on('close', function () {
      // When the dropdown is closed, aria-expanded="false"
      self.$selection.attr('aria-expanded', 'false');
      self.$selection.removeAttr('aria-activedescendant');
      self.$selection.removeAttr('aria-owns');

      // This needs to be delayed
      window.setTimeout(function () {
        self.$selection.focus();
      }, 0);

      self._detachCloseHandler(container);
    });
```

PR #4386, opened 2016-05-27
Add extra event results:rendered after results are rendered

ScullWM:
This pull request includes a
- [ ] Bug fix
- [X] New feature
- [ ] Translation

- Trigger new event "**results:rendered**" when results are rendered
  It could be usefull to trigger an event when results have been well formated and are displayed. Exemple: generate the item colored avatar.
  capture d ecran 2016-05-27 15 33 12


ScullWM:
I truly regret choosing this javascript plugin. 5 month after submitting this simple pull-request (with tests updated and passing) we don't have any feedback (and lot of other issues/PR don't have too).

Uffman:
Hi, 
i hade the same issue. Wanted to position tool tip based if dropdown opens up or down.
And came with my own solution changing select2 lib. version 4.03
I added an event "loaded" in function 
`EventRelay.prototype.bind`

`var relayEvents = [
                'open', 'opening',
                'close', 'closing',
                'select', 'selecting',
                'unselect', 'unselecting',
                'loaded'
 ];
`
then call trigger of the event, when data is attached to DOM. In function  
`AttachBody.prototype.bind`

in

` container.on('results:all', function () {
                          self._positionDropdown();
                          self._resizeDropdown();
                          self.trigger("loaded", {});
                      });`

i added: self.trigger("loaded", {});

and then just listen to new event

`$('select').on('select2:loaded', function (evt) {
  // Do something
});`

Probably not the best solution but did the trick for my problem

Issue #4384, opened 2016-05-27
Tabbing order after selecting single value


SgtOddball:
## Prerequisites
- [x] I have searched for similar issues in both open and closed tickets and cannot find a duplicate
- [x] The issue still exists against the latest `master` branch of Select2
- [x] This is not a usage question (Those should be directed to the [community](https://select2.github.io/community.html))
- [x] I have attempted to find the simplest possible steps to reproduce the issue
- [ ] I have included a failing test as a pull request (Optional)
## Steps to reproduce the issue
1. Add select2 inside div with associated label above 

``` HTML
    <div class="form-group">
        <label class="control-label" for="County">County</label>
        <select class="form-control" id="County">
            <optgroup label="England">
                <option value="Avon">Avon</option>
                <option value='Bedfordshire'>Bedfordshire</option>"
                <option value='Berkshire'>Berkshire</option>
                <!-- et al -->
            </optgroup>          
        </select>
    </div>
    <div class="form-group">
        <label class="control-label" for="Tabbing">Tabbing</label>
        <input id="Tabbing" name="Tabbing" type="text"/>
    </div>
```
1. Apply select2 

``` javascript
    $('select.county').select2();
```
1. Run page
## Expected behavior and actual behavior

Tab from input above works fine using keyboard, tab to the select box and select an item and now try to tab to an input below, the tabbing doesn't work and instead goes to the first tabbable element on the page. 

This also occurs if using tabbing index as well. Might be related to loss of focus issues but I've not found one that doesn't claim to have been fixed for 4.0.2 hence the issue.## Environment

Browsers
- [x] Google Chrome
- [x] Mozilla Firefox
- [x] Internet Explorer
- [x] Edge

Operating System
- [x] Windows
- [ ] Mac OS X
- [ ] Linux
- [ ] Mobile

Libraries
- jQuery version:v1.11.3
- Select2 version:v4.0.2
## Isolating the problem
- [x] This bug happens [on the examples page](https://select2.github.io/examples.html)
- [x] The bug happens consistently across all tested browsers
- [x] This bug happens when using Select2 without other pluigns
- [ ] I can reproduce this bug in [a jsbin](https://jsbin.com/)

jonahgreenthal:
Is there any known workaround for this?

If not, is there an estimate on when a fix will be out?

SgtOddball:
[...]

PR #4405, opened 2016-06-07
Tabbing order after selecting single value

hbb1453:
This pull request includes a

- [X] Bug fix
- [ ] New feature
- [ ] Translation

- Before dropdown removed from DOM focus redirected to $selection element

https://github.com/select2/select2/issues/4384

### DIFF
--- a/src/js/select2/dropdown/search.js
+++ b/src/js/select2/dropdown/search.js
@@ -60,6 +60,7 @@ define([
       self.$search.attr('tabindex', -1);
 
       self.$search.val('');
+      this.$selection.focus();
     });
 
     container.on('focus', function () {


kevin-brown:
This _should_ be happening in the [base selection](https://github.com/select2/select2/blob/566c7846fbf8c3b7674346f26d0ee872c0a77629/src/js/select2/selection/base.js#L84), which applies for both the single and multiple select. So clearly this is supposed to work, it really works sometimes, but there is something going wrong.

In any case, this solution suffers from knowing too much. The dropdown is not expected to have knowledge of the selection, so there is no reason for it to be accessing the selection's information (much less an internal element). As a result, I'm going to reject this patch.

rvion:
FWIW, the problem this patch was solving is still hitting a lot of users.

alexweissman:
#4427 appears to be the correct way to solve this problem, and has been integrated into 4.0.4.

Issue #3975, opened 2015-12-01
Doubled tags problem


nabidefacto:
Hello,

I found one bug with tags. 
If you try to type-in the same tag, that've already been added, it removes both tags. 
You can try it here: https://select2.github.io/examples.html

But, in version 3.5.3 works ok: http://select2.github.io/select2/

Thank you

mechanicious:
Strongly connected with #3951, although they don't mention them about disappearing tags.

kevin-brown:
Closing this as a duplicate of https://github.com/select2/select2/issues/3951.

PR #4416, opened 2016-06-13
Doubled tags problem

ysavary:
This pull request includes a

- [x] Bug fix: https://github.com/select2/select2/issues/3975
- [ ] New feature
- [ ] Translation

With multiple tags option enabled:
- press enter when adding again an exiting tag will do nothing
- click on an exiting tag will remove the tag

This pull-request fix the behavior when clicking on a existing tag by doing nothing like pressing enter.

### DIFF
--- a/src/js/select2/results.js
+++ b/src/js/select2/results.js
@@ -436,15 +436,7 @@ define([
       var data = $this.data('data');
 
       if ($this.attr('aria-selected') === 'true') {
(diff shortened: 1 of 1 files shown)


kevin-brown:
This pull request removes code that is in place for accessibility reasons. It also doesn't appear to fix the linked ticket at all, as one deals with the selection and tagging while the other deals with results.

PR #4433, opened 2016-06-21
Update .travis.yml

PeterDaveHello:
Also run the tests on nodejs v0.12 / v4 / v6

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,9 @@ sudo: false
 
 node_js:
   - 0.10
+  - 0.12
+  - 4
+  - 6
 
 env:
   global:


.travis.yml
```diff
@@ -4,6 +4,9 @@ sudo: false
 
 node_js:
   - 0.10
+  - 0.12
+  - 4
+  - 6
```
stonio: I would also add `   - node` to test on latest stable Node.js release.

Reference : https://docs.travis-ci.com/user/languages/javascript-with-nodejs#Available-Versions

kevin-brown:
I'm going to close this PR, not because the changes are bad, but because they don't directly affect Select2. The Node version specified in the Travis file is for the Node version used for local development, not for the version that is necessary to run Select2 without any issues.

We do have plans to update PRs so they build against a newer version of Node, but this isn't a super high priority.

PR #4441, opened 2016-06-24
docs: more appropriate terminology ("highlight" -> "focus")

uestla:
I think "focus" fits better.


kevin-brown:
Thanks for the pull request! I agree that "focus" is the more appropriate word here.

Issue #4404, opened 2016-06-07
Duplicate tags on Select with multiple


GitHubStig:
## Steps to reproduce the issue
1. go to https://select2.github.io/examples.html
2. scroll down to Tagging support example
3. type in `red`, yes with a space after.
   ![select2 double tags](https://cloud.githubusercontent.com/assets/9544391/15851281/b510f11a-2cce-11e6-95f2-5f90fa47ac4d.gif)
## Expected behavior and actual behavior

When I follow those steps, I see 2 `red` tags.

I was expecting only 1.
## Environment

Browsers
- [x] Google Chrome
- [x] Mozilla Firefox

Operating System
- [x] Mac OS X
## Isolating the problem
- [x] This bug happens [on the examples page](https://select2.github.io/examples.html)
- [x] The bug happens consistently across all tested browsers

alexweissman:
As of `4.0.6-rc.1`, I cannot seem to replicate this issue anymore.

Issue #4412, opened 2016-06-10
Select2 doesn't trim text in <option> tag


guoyunhe:
```html
<select multiple="multiple" style="width:400px">
  <option value="1">
         Dog
  </option>
  <option value="2">
         Cat
  </option>
  <option value="3">
         Monkey
  </option>
  <option value="4">
         Panda
  </option>
</select>
```

```javascript
$('select').select2();
```

When backspace delete tag, you can see here are a lot of spaces in text.

Screenshot:

![spectacle d14927](https://cloud.githubusercontent.com/assets/5836790/15976308/9b26f95a-2f5a-11e6-881a-1505876dd922.png)

Demo:

http://codepen.io/guoyunhebrave/pen/OXMBQr

kevin-brown:
Select2 pulls the `text` from the `<option>` based on the [`text` property of the `<option>`](https://developer.mozilla.org/en-US/docs/Web/API/DOMString). This is important to know, because it follows the standard HTML rules of parsing an `<option>` and relies on the browser doing it properly. Nowadays, most browsers parse it properly and a handful of them will render it properly.

Whitespace is significant at times in HTML, and I can't easily explain the whitespace rules because it's complex. But **whitespace around text within an element is significant**, and as you've seen the browser will not remove it. As a result, Select2 will also see the whitespace and doesn't strip it out (we used to, but that caused a _new class of problems_).

My recommendation is to not include the whitespace around the option text, and instead render it with the rest of the `<option>` on one line.

```html
<select multiple="multiple" style="width:400px">
  <option value="1">Dog</option>
  <option value="2">Cat</option>
  <option value="3">Monkey</option>
  <option value="4">Panda</option>
</select>
```

PR #4442, opened 2016-06-26
Correction for the bug #4404

gpguerra:
The option function was creating a new option on DOM for a entered value with spaces at the end.

This is for https://github.com/select2/select2/issues/4404.

### DIFF
--- a/src/js/select2/data/tags.js
+++ b/src/js/select2/data/tags.js
@@ -53,7 +53,8 @@ define([
           }, true)
         );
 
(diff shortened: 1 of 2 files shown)


alexweissman:
As of `4.0.6-rc.1`, I cannot seem to replicate this problem anymore.

Also, I'm not sure if we should be trimming whitespace as a general rule.  See https://github.com/select2/select2/issues/4412#issuecomment-225382014

Issue #3512, opened 2015-06-23
Hitting backspace on select element showing the previous search results in case when delay in ajax call specified


prate3k:
In case when ajax options are specified and user tries to search an element, once user gets the suggestion list if he clicks backspace continuously, until search text becomes empty string "", still it shows the suggestion list for the last character removed using backspace.

This is happening only when ajax option with some delay ( in my case 250ms) is specified. Its not cancelling the previous ajax call, when the search text becomes empty and populate the list with search results.

JS code : 

``` js
$("target_elem_id").select2({
  minimumInputLength : 1,
  ajax : {
    url : "some-url.json",
    dataType : 'json',
    delay : 250,
    data : function(params) {
      return {
          name : params.term,
          maxResults : 5
      };
    }
  }
});
```

Steps to reproduce : 
1> Type in some character say "new", so that is shows suggestion list : 

![search_with_text](https://cloud.githubusercontent.com/assets/4700266/8299542/c88394dc-1997-11e5-887a-642cfa72e3f8.PNG)

2> Remove the complete search text by hitting backspace continuously : 

![empty_search](https://cloud.githubusercontent.com/assets/4700266/8299547/ea17640c-1997-11e5-8b17-cf2fb897f08f.PNG)

The above suggestion list is same as the suggestion list shown when we type character "n" : 

![with_search_text](https://cloud.githubusercontent.com/assets/4700266/8299561/2f2f6cce-1998-11e5-9663-deedfbee4ef9.PNG)

It means the previous ajax call was in progress and in between we emptied search text, and it populated suggestion list with search results received in last ajax call.

maykSVK:
I am experiencing exactly the same problem. Watching network communication I was able to see, that after last letter beeing removed it sends two requests, but the empty char query and then query with last letter (even though input is already empty). This is causing problems with filtering and I supose it's a bug. Is there any progress with this case?

arirangz:
+1

jimbogit:
+1 - it happens on your example page too.  please fix

GTCrais:
I have the same issue.

Temporary fix - in your `processResults` method check for term's existence:

    processResults: function (data) {
        // ...

	let term = $("#your_selector").data('select2').$dropdown.find("input").val();

        if (!term) {
            return { results: [] };
        }

        // ...
							
    }

kevin-brown:
In theory, all we need to do is cancel the request when a `results:message` rolls on through (since that triggers the render).

[...]

PR #4447, opened 2016-06-29
fix for bug #3512, ajax call with delay and quick search string removing

vshulga-cogniance:
This pull request includes a
- [x] Bug fix
- [ ] New feature
- [ ] Translation

The following changes were made to fix bug #3512: 
Hitting backspace on select element showing the previous search results in case when delay in ajax call specified. Similar bug could be caught if we use ajax data adapter with minimum input length, if we type search string and remove it quickly we could see search results even if search string is less than minimum input length.

I've added checking if process results is run with the same term for which search with ajax call was called.

https://github.com/select2/select2/issues/3512

### DIFF
--- a/src/js/select2/results.js
+++ b/src/js/select2/results.js
@@ -248,6 +248,10 @@ define([
     this.$results.attr('id', id);
 
     container.on('results:all', function (params) {
+      if (params.query && self.lastParams &&
+          params.query.term !== self.lastParams.term) {
+        return;
+      }
       self.clear();
       self.append(params.data);
 
@@ -266,6 +270,7 @@ define([
     });
 
     container.on('query', function (params) {
+      self.lastParams = params;
       self.hideMessages();
       self.showLoading(params);
     });


ZuBB:
Hi there! 

select2 team could you please take a look on this?

alexweissman:
It appears that this problem comes up in a number of contexts:

- ajax delay
- ajax requests that naturally take a while to complete
- minimumInputLength
- maximumInputLength

Can someone confirm that this PR solves the problem in all of these contexts?

ZuBB:
Hi @alexweissman!

just want to let you know that our team (I work with author of MR together) decided to stop using this lib. so you need to wait for those confirmations from other people..

in any way thanks to all core contributors/maintainers for making this piece of software better 👍

alexweissman:
No prob, thanks for the update!

nils703:
Hi @alexweissman,

we didn't test all the contexts you mentioned, but the solutions works for us with a little change for the following 2 contexts:
- ajax delay
- ajax requests that naturally take a while to complete

the change we did was additionally adding the check from the event "results:all" to "results:append". because we had the same problem with paging and changing the search term to quickly.

PR #4448, opened 2016-06-29
some people may prefer use css

markpro-code:


kevin-brown:
This was done in https://github.com/select2/select2/pull/3350 about a year ago.

PR #4462, opened 2016-07-07
Better detection of aborted ajax queries

PowerKiKi:
This pull request includes a

- [x] Bug fix
- [ ] New feature
- [ ] Translation

On Chrome 50, `$request.status` is a number, not a string thus the aborted query detection failed. For the end-user this resulted in seeing the 'The results could not be loaded.' message when he typed too fast for too long, instead of the correct 'Searching…' message.

This commit check `$request.status` both as a number and as a string, keeping backward compatibility while fixing the case for Chrome.

### DIFF
--- a/src/js/select2/data/ajax.js
+++ b/src/js/select2/data/ajax.js
@@ -82,7 +82,7 @@ define([
       }, function () {
         // Attempt to detect if a request was aborted
         // Only works if the transport exposes a status property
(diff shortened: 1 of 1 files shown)


kevin-brown:
This is a duplicate of https://github.com/select2/select2/pull/4356, which is going to be pulled in for 4.1.0.

PR #4471, opened 2016-07-10
Fixes #3296 and fixes #4224 improving dropdown width calculation

### LINKED ISSUE #3296: Select2 dropdown size mismatch because of jQuery floating point issue
Hi,

This refers to version 3.5.2 of Select2 plugin. When used in a responsive design such as bootstrap, there are instance when the width of the dropdown is slightly smaller than than the parent Input/Select box. When the browser window is resized or tested with other resolutions, sometimes the dropdown is exactly the same size and parent control and sometimes it is not. Check the right edge of textbox in screenshot.
Upon further investigation, it seems like the method used to calculate the width of dropdown is the culprit:
In method positionDropdown, line 1237:
width = container.outerWidth(false)

It turns out that jQuery outerWidth method only returns integer value and ignores fractional values which is causing the above issue. Replacing the above with:

var rect = $(this.container)[0].getBoundingClientRect();
var width= rect.right - rect.left;

fixes the issue.
screenshot 2015-04-28 23 22 56

### LINKED ISSUE #4224: Dropdown not Exact width of Select Container (Simple Solution)
I noticed when using this that the dropdown width would always be about 1px off either wider or narrower than the select. Not really a big deal but when using 1px borders, it just looks "wrong".

This seems to happen when you have the select set to a percentage in a responsive layout.

The solution is easy. Since `.outerWidth` rounds the pixels, you can use a different means of getting the exact width.

On line 4249 I changed this...

``` js
  AttachBody.prototype._resizeDropdown = function () {
    var css = {
      width: this.$container.outerWidth(false) + 'px'
    };
```

To this....

``` js
  AttachBody.prototype._resizeDropdown = function () {
    var css = {
      width: this.$container.get(0).getBoundingClientRect().width + 'px'
    };
```

This is an easy fix for those interested :)

Adirael:
This pull request includes a
- [X] Bug fix
- [ ] New feature
- [ ] Translation

- Use getBoundingClientRect when available to calculate element widths

Fixes #3296 and it's duplicates (#4224)

### DIFF
--- a/src/js/select2/core.js
+++ b/src/js/select2/core.js
@@ -129,7 +129,8 @@ define([
     }
 
     if (method == 'element') {
-      var elementWidth = $element.outerWidth(false);
+      var elementWidth = $element[0].getBoundingClientRect().width || 
+                         $element.outerWidth(false);
 
       if (elementWidth <= 0) {
         return 'auto';
--- a/src/js/select2/dropdown/attachBody.js
+++ b/src/js/select2/dropdown/attachBody.js
@@ -198,8 +198,11 @@ define([
   };
 
   AttachBody.prototype._resizeDropdown = function () {
+    var containerWidth = this.$container[0].getBoundingClientRect().width ||
+                         this.$container.outerWidth(false);
+                         
     var css = {
-      width: this.$container.outerWidth(false) + 'px'
+      width: containerWidth + 'px'
     };
 
     if (this.options.get('dropdownAutoWidth')) {


Adirael:
Hey @kevin-brown, thanks for your input. 

I've never used QUnit but I guess it's a matter of giving the $parent a decimal width, setting the $select up to be 100% and check it's final width against the $dropdown's width. 

I'll give it a try and fix the pull request.

Cheers,
Angel

alexweissman:
Can someone work on the tests for this?

Also, it appears that this needs to be done in `results.js` as well - see this closed duplicate PR: https://github.com/select2/select2/pull/4738/files#diff-638a01f62c16447e7081665aac148a07R401

PR #4477, opened 2016-07-13
New feature: Added option to open the combo using the KEYS.DOWN key and focus on $…

gcabrini:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

- Added option to open combo using KEYS.DOWN
- Focus on $element after choosing an option


kevin-brown:
As noted in https://github.com/select2/select2/issues/3472, we are not planning on binding to this key until we set up support for changing the selection without opening the dropdown.

PR #4479, opened 2016-07-14
Clone the `accesskey` from the original select

dimmelsw:
This pull request includes a
- [x] New feature

The accesskey from the original select is cloned to select2 to jump to the select2 field by keyboard.


kevin-brown:
We're going to need to do some manual testing to verify that this actually works in both the single and multiple select boxes. It would also be awesome if we could get a basic test [similar to the `tabindex` ones](https://github.com/select2/select2/blob/6844be777b0fe9af1d2b9a8f6051b7eb7d532247/tests/a11y/selection-tests.js).

kevin-brown:

PR #4498, opened 2016-07-22
Stop keyup event if keydown was handled

### LINKED ISSUE #4495: Stop keyup event if keydown was handled
there's code which catches `keydown` event, converts it into `keypress` own event and stops its propagation:

```
    this.$selection.on('keydown', '.select2-search--inline', function (evt) {
      evt.stopPropagation();

      self.trigger('keypress', evt);
```

But `stopPropagation` in keydown's handler does not prevent `keyup` firing. So later there will be a `keyup` event bubbling outside Select2. 

`keyup` event is commonly used to close dialogs for example. So if I press ESC in select2 with open dropdown this dropdown will close but so does the dialog. 

What I want: if I press ESC with opened dropdown then only dropdown should close, but if I press ESC with closed dropdown then the event should bubble (e.g. dialog should close).

Currently to fix it I have to catch `keyup` on some Select2's element.  But it's tricky because it depends on the fact whether dropdown is opened or closed. But in `keyup` handler is hard to understand whether dropdown was opened or closed because it's already closed in any way:

```
            select2.$selection.on("keypress keyup keydown", function (e) {
                // if (select2.isOpen())  <- this does not work
                e.stopPropagation();
            });
```

What I suggest: beside 'keydown' add 'keyup' handler where stop propagation if `keydown` was handled:

```
    this.$selection.on('keyup', function (evt) {
        if (self._keyUpShouldPrevented) {
            evt.stopPropagation();
        }
    });
```

NOTE: there's `_keyUpPrevented` field already which would suite well but it's reset in `Search.prototype.handleSearch`.

evil-shrike:
This pull request includes a
- [x] Bug fix
- [ ] New feature
- [ ] Translation

- core.js: Select2: added 'keyup' handler to call `stopPropagation` if event was processed before in keypressed handler (uses `_keyUpPrevented` flag)
- selection.js: Search: added similar 'keyup' handler and the same logic added into existing handler 'keyup.search input.search'

Fixes #4495 Stop keyup event if keydown was handled

NOTE: 
why two handlers in Search are needed? I noticed that stopPropagation in 'keyup.search input.search' handler is not always enough. And also leaving only one for 'keyup' (a new one) is also not enough as `handleSearch` reset `keyUpPrevented` field.

### DIFF
--- a/dist/js/select2.full.js
+++ b/dist/js/select2.full.js
@@ -1968,6 +1968,12 @@ S2.define('select2/selection/search',[
           return;
         }
 
+        // if the event was processed as 'keypress' (see 'keydown' handler) prevent it from leaving 
+		if (self._keyUpPrevented) {
+		  evt.stopPropagation();
+		  return;
+		}
+
         self.handleSearch(evt);
       }
     );
@@ -5378,7 +5384,17 @@ S2.define('select2/core',[
           evt.preventDefault();
         }
       }
+
+      self._keyUpPrevented = evt.isDefaultPrevented();
     });
+
+    this.on('keypress', function (evt) {
+      if (self._keyUpPrevented) {
+        evt.stopPropagation();
+        self._keyUpPrevented = false;
+      }
+    });
+
   };
 
   Select2.prototype._syncAttributes = function () {
--- a/dist/js/select2.full.js
+++ b/dist/js/select2.full.js
@@ -1944,6 +1944,16 @@ S2.define('select2/selection/search',[
       }
     );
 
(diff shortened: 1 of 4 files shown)


evil-shrike:
Well, the solution is incomplete. It works in mulple selection mode, but not in single. 

Just open a form with auto submit (by pressing Enter) and focused select2, press Enter, dropdown opens and the form closes immediately.
Can't find a way to prevent keyup from bubbling up from single selection Select2 :( 
Any help is appreciated.

kevin-brown:

PR #4509, opened 2016-07-30
Make tag match case insensitive

yaronuliel:
This pull request includes a

New feature

- Tags match changed to be case insensitive (In order to be consistent with option matching and prevent duplication of tags)


simi:
I think it will be better to move `.idea` line to your [global gitignore](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore) since it is not project related.

yaronuliel:
@simi 
Thanks - I removed it

Issue #4516, opened 2016-08-04
Can't add option with blank value


rsmything:
i'll try to add first option with blank value with text

.coffee

```
sort_data[''] = [{ text: 'Please select', id: '' }]
$(sort_columns_select).select2('destroy').empty().select2 data: sort_data[tracker]
```

But this code inserts <option>Please select</option> instead <option value=''>Please select</option>.

I think it's wrong action, because 1st sample submit to the server string 'Please select' instead blank value.

I found value adding in the code:

```
3194:    if (data.id) {
3195      option.value = data.id;
3196    }
```

I think in this case the mistake to miss "value" parameter for the empty string in the key id.
This should be

```
if (data.id !== undefined) {
  option.value = data.id;
} 
```

rsmything:
https://github.com/select2/select2/pull/4517

PR #4517, opened 2016-08-04
Permit to add a empty string to the 'value' attribute in option tag

rsmything:
This closes https://github.com/select2/select2/issues/4516.

### DIFF
--- a/dist/js/select2.js
+++ b/dist/js/select2.js
@@ -3191,7 +3191,7 @@ S2.define('select2/data/select',[
       }
     }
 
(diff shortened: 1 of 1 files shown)


pm-Egor-Genning:
Please merge this commit https://github.com/select2/select2/commits/master?after=16f6b10628f9000918708cef9b3da615119cf6d6+69 because issue still exists

alexweissman:
It will need to be submitted as a new PR, because the original fork was deleted.

PR #4519, opened 2016-08-05
add Georgian translation

giokaxo:
This pull request includes a
- Georgian translation

### DIFF
--- a/src/js/select2/i18n/ka.js
+++ b/src/js/select2/i18n/ka.js
@@ -14,7 +14,8 @@ define(function () {
     inputTooShort: function (args) {
       var remainingChars = args.minimum - args.input.length;
 
-      var message = 'ჷთხოვთ აკრიფოთ ' + remainingChars + ' ან მეტი სიმბოლო';
+      var message = 'ჷთხოვთ აკრიფოთ ' + remainingChars +
+        ' ან მეტი სიმბოლო';
 
       return message;
     },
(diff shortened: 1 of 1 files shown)


src/js/select2/i18n/ka.js
```diff
@@ -0,0 +1,44 @@
...
+    errorLoading: function () {
+      return 'სია ცარიელია.';
+    },
+    inputTooLong: function (args) {
+      var overChars = args.input.length - args.maximum;
+
+      var message = 'გთხოვთ წაშალოთ ' + overChars + ' სიმბოლო';
+
+      if (overChars != 1) {
+        message += 's';
```
kevin-brown: I have no idea how Geogian pluralization rules work, so I just want to confirm. Do you really just throw an 's' on the end to pluralize these statements?

src/js/select2/i18n/ka.js
```diff
@@ -0,0 +1,38 @@
+define(function () {
+  // Georgian
+  return {
+    errorLoading: function () {
+      return 'სია ცარიელია.';
```
kevin-brown: I don't have anyone to reach out to about this translation, and the automated translators are giving me something along the lines of "To display".

kevin-brown:
Thanks for contributing to Select2!

Unfortunately, because a lot of these translations aren't converting back to English properly, and some of them aren't showing up as valid Georgian when looking around, I'm going to have to reject this pull request.

src/js/select2/i18n/ka.js
giokaxo: its valid translation, but because of Georgia is small country and not many people translates resources, automated translators aren't correct in most cases

PR #4550, opened 2016-08-25
Update _examples.scss

realronchi:
This pull request includes a
- [x] Bug fix
- [ ] New feature
- [ ] Translation

Add max-height for div events log.


kevin-brown:
This makes sense to me, but I want to run it by others first as there might be a better place to put this.

/cc @select2/design

alexweissman:
Docs have been moved to https://github.com/select2/docs, so this change is now obsolete.  Appreciate the help, though, and could use some help in the new docs repo!

Issue #4557, opened 2016-08-29
Fix commonjs support


pcfreak30:
You are using a half-baked/broken UMD shim that does not work in common JS. You do not set module.exports or return the factory.

kevin-brown:
Our current wrapper is based on https://github.com/umdjs/umd, an older version than what is currently there.

It appears as though there was an update which fixed this issue in https://github.com/umdjs/umd/commit/2bbe19cc557fb1b99691932911e533cdc391d1f5.

pcfreak30:
My PR uses the same I believe.

PR #4558, opened 2016-08-29
*Fix UMD commonjs support

pcfreak30:
#4557

### DIFF
--- a/src/js/wrapper.start.js
+++ b/src/js/wrapper.start.js
@@ -11,7 +11,21 @@
     define(['jquery'], factory);
   } else if (typeof exports === 'object') {
     // Node/CommonJS
-    factory(require('jquery'));
+    module.exports = function (root, $) {
+      if (!root) {
+        // CommonJS environments without a window global must pass a
+        // root. This will give an error otherwise
+        root = window;
+      }
+
+      if (!$) {
+        $ = typeof window !== 'undefined' ? // jQuery's factory checks for a global window
+            require('jquery') :
+            require('jquery')(root);
+      }
+
+      return factory($, root, root.document);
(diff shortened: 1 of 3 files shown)


src/js/wrapper.start.js
```diff
@@ -11,7 +11,21 @@
...
+        root = window;
+      }
+
+      if (!$) {
+        $ = typeof window !== 'undefined' ? // jQuery's factory checks for a global window
+            require('jquery') :
+            require('jquery')(root);
+      }
+
+      return factory($, root, root.document);
```
kevin-brown: Is there any reason why we are passing in `root` and `root.document` in here? The `factory` only appears to take in `jQuery` as the first parameter.

src/js/wrapper.start.js
```diff
@@ -11,7 +11,21 @@
...
-    factory(require('jquery'));
+    module.exports = function (root, $) {
+      if (!root) {
+        // CommonJS environments without a window global must pass a
+        // root. This will give an error otherwise
+        root = window;
+      }
+
+      if (!$) {
+        $ = typeof window !== 'undefined' ? // jQuery's factory checks for a global window
```
kevin-brown: I prefer the [UMD patterns template](https://github.com/umdjs/umd/blob/95563fd6b46f06bda0af143ff67292e7f6ede6b7/templates/jqueryPlugin.js#L10-L21) with how it breaks this ternary conditional out into a full conditional block.

```
if (jQuery === undefined) {
  if (typeof window !== 'undefined') {
    jQuery = require('jquery');
  } else {
    jQuery = require('jquery')(root);
  }
}
```

Issue #3121, opened 2015-03-09
Force showing dropdown-container always above or below


ifthenelse:
As far as I know, there's no option which cover this.
Since it can't be just achieved via css customization, the only way is to intercept the opening of the select and the reposition the dropdown-container manually, but It's definitely not the best solution.
Any chance to see it in the upcoming v4.x?

mh-nichts:
I would also be interested by this feature !

leonardomdr:
I did this in Select2 3.5.2 to force the container to start above instead of below the search box.

In select2.js you search for "positionDropdown: function() {" and change to "positionDropdown: function(forceAbove) {"

In the same function, after the var settings, paste this code:
```javascript
forceAbove = typeof forceAbove !== 'undefined' ? forceAbove : false;
if (forceAbove) {
    enoughRoomAbove = true;
    enoughRoomBelow = false;
}
```

After that, to force your select2 to show above, use this

```javascript
.on("select2-open", function() { 
    $(this).select2('positionDropdown', true);
});

**Comment by apaulo** [comment]
There is a very simpler way to implement a forceAbove:

1) In select2.js search for "positionDropdown: function()";
2) at the var settings change 
`aboveNow = $dropdown.hasClass("select2-drop-above")`
to
`aboveNow = this.opts.forceabove || $dropdown.hasClass("select2-drop-above")`

3) Now you can setup select2 with option
`forceabove = true`

**Comment by ganholete** [comment]
Same question

**Comment by tirdadc** [comment]
I was considering switching from Chosen to Select2, but I was surprised to see that this was something you couldn't control, and worse, it would be inconsistent: depending on the contents of the dropdown, it alternates between above and below.

**Comment by oslnx** [comment]
```
<a id="country" data-type="select2" data-pk="1" data-select-search="true" data-value="0" data-original-title="Todos">Todos</a>

    $('#country').editable({
        pk: 1,
        name: 'country',
        savenochange: true,
        defaultValue: '0',
        mode: 'inline',
        emptytext: 'Todos',
        source: '/?json=countries',
        showbuttons: false,
        select2: {
            width: 200,
        }
    });
    $('#country').on('hidden', function(e, reason) {
        $(this).trigger('click');
    });
    $('#country').trigger('click');

SOURCE JSON: [{"id":"0","text":"Todos"},{"id":"BR","text":"Brazil"},{"id":"MX","text":"Mexico"}]

```

[...]

PR #4560, opened 2016-08-31
fix : issues #3121 & #3354 - adding options for dropdown positionning…

DVD27:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

The following changes were made :

- Add an option for choice the dropdown position on open (above, below or both) (issues #3121)
- Add an option for choice what to do when backspace event is produce on multiple select (unselect (delete just the tag), none (do nothing), text (transform the tag to text like today)) (issues #3354)


dist/js/select2.full.js
```diff
@@ -30,13 +30,11 @@
...
 if (!S2) { S2 = {}; } else { require = S2; }
 /**
- * @license almond 0.3.1 Copyright (c) 2011-2014, The Dojo Foundation All Rights Reserved.
- * Available via the MIT or new BSD license.
- * see: http://github.com/jrburke/almond for details
+ * @license almond 0.3.2 Copyright jQuery Foundation and other contributors.
+ * Released under MIT license, http://github.com/requirejs/almond/LICENSE
  */
 //Going sloppy to avoid 'use strict' string cost, but strict practices should
 //be followed.
```
LayZeeDK: If you are removing the `sloppy` JSLint option, you should remove this comment as well.

kevin-brown: This is a part of the built JS files, and this PR upgraded Almond in the process (which is where this comment comes from).

Ivanca:
It seems you are writing `dropdownLocation` in some parts where it should be `dropdownPosition`

DVD27:
Thanks Ivanca (I don't have see it !), I have correct this issues in the new pull-request.

PR #4576, opened 2016-09-11
Slovene translation

mitja-p:
This pull request includes a slovene translation.


src/js/select2/i18n/sl.js
```diff
@@ -0,0 +1,50 @@
...
+      else if (overChars != 1) 
+        message += 'e';
+
+      return message;
+    },
+    inputTooShort: function (args) {
+      var remainingChars = args.minimum - args.input.length;
+
+      var message = 'Prosim vpišite še ' + remainingChars + ' znak';
+      if (remainingChars == 2)
```
kevin-brown: Within Select2, we enforce that braces are always used in conditionals. This largely has to do with having consistency across the code base.

mitja-p:
Ok, fixed it. Thanks for notifying me!

PR #4583, opened 2016-09-14
Docs for:  Does Select2 include extra information in these events?

guettli:
This pull request includes a
- [ ] Bug fix
- [ ] New feature
- [ ] Translation
- [x] Documentation

Added docs for "Does Select2 include extra information in these events?"

Related: http://stackoverflow.com/questions/39333354/select2-how-to-add-extra-information-for-selected-item

### DIFF
--- a/docs/_includes/options/events/jquery.html
+++ b/docs/_includes/options/events/jquery.html
@@ -43,8 +43,17 @@
   <h3>
     Does Select2 include extra information in these events?
   </h3>
+If your server part returns additional data in the json dictionary, you can access like this:
 
-  {% include options/not-written.html %}
+{% highlight js linenos %}
+$(function() {
+    $('#mywidget').on('select2:select', function(evt){
+        alert(evt.params.data.mydata);
+    });
+})
+{% endhighlight %}
+
+  
 
   <h3>
     How can I attach listeners for these events?


docs/_includes/options/events/jquery.html
```diff
@@ -43,8 +43,17 @@ <h2 id="events-public">
   <h3>
     Does Select2 include extra information in these events?
   </h3>
+If your server part returns additional data in the json dictionary, you can access like this:
 
-  {% include options/not-written.html %}
+{% highlight js linenos %}
+$(function() {
+    $('#mywidget').on('select2:select', function(evt){
+        alert(evt.params.data.mydata);
```
kevin-brown: Ideally this wouldn't necessarily be documenting the fact that you can send back additional data on the selected data object, but instead it would be documenting the fact that `evt.params.data` exists on `select2:select`. Select2 passes a bunch of data along with events through `evt.params`, and the end goal for this section was to document that.

guettli: I am happy that I found `evt.params.data`. I would like to share this information and provided this pull request. Please alter my pull request, if you think it can be improved. How do you want this change to look like?

guettli:
Hi Kevin.

This pull request is still open. If you don't want to use my proposal, then what do you suggest?

It took me some time to find this. It would be nice to fill the place holder in the docs.

alexweissman:
I have now added this to the new docs: https://select2.org/programmatic-control/events#event-data

Issue #4445, opened 2016-06-28
Add possibility to insertBefore instead of insertAfter


rippersvk:
In function `Select2.prototype._placeContainer` 
there is code to insert select2 element after actual select tag
    `$container.insertAfter(this.$element);`

I would like to have option to insert it before actual select tag.
Mainly because of error handling when I am using bootstrap.
I set it to insert error texts after inputs (without select2), which using select2 and insertAfter displays error before input...

Bmju:
I have the same problem. The problem applies to the jQuery Validation plugin as well. Once Select2 has been applied, the DOM order is: the orginal input element (hidden by Select2), then any form validation messages, then the inserted Select2 elements. So form validation messages display before the visible input element instead of after it.

Bmju:
I have now understood that the correct solution, at least for jQuery validation, is to use the errorPlacement() callback instead (http://stackoverflow.com/questions/26498899/jquery-validate-custom-error-message-location).

tanuj-github:
It is not only about errorPlacement in Query validation. I have same problem and I ended up finding no solution for such a small thing.

Can we just provide a setting in options to pass the value for insert before/after in cureent select 2 option?

alexweissman:
@MikeBeaton has deleted his fork, so if we still want this as a feature it will need to be submitted as a new pull request.  Please note that a suitable test should be added for this feature as well, in order for it to be considered for a release.

pedrofurtado:
We don't have immediate plans to provide this. We are focused now to fix UI bugs. But if you open a PR with unit tests, I will be glad to review and approve if everything is ok.

MikeBeaton:
I don't think I have permissions to re-open https://github.com/select2/select2/pull/4594 myself (?), but I'm happy to re-open it and add a test or two, if that would help.

l10nelw:
I'd like to make a case that insertBefore should be default behaviour. Currently, browser validation on a select2 field produces an error message in this awkward position:

<img width="581" alt="image" src="https://user-images.githubusercontent.com/1300917/210480816-e1c9879b-2bdf-48d6-8d06-0b597959a072.png">

If the the select2 container was inserted before the select, the validation message would appear under the select2 container as you'd expect them to be.

PR #4594, opened 2016-09-20
Add DOM placement 'insertBefore' option

Bmju:
This pull request includes a

- [x] Bug fix
- [x] New feature
- [ ] Translation

Add 'insertBefore' option to change DOM placement of Select2 generated items.

This is useful because various form validation libraries (including jqQuery Validation) put their validation messages immediately after the original form element. This means that such messages end up appearing before the visible form element instead of after it, once Select2 has been applied. (This is because the DOM order once Select2 has been applied is: the orginal input element (hidden by Select2), then any form validation messages, then the inserted Select2 elements.) Inserting the Select2 generated DOM items before the original element fixes this.

This pull request suggests implementing this as a new Select2 option called 'insertBefore', which if omitted defaults to the previous behaviour. E.g. $('#id').select2({ ... , insertBefore: true };

Addresses https://github.com/select2/select2/issues/4445

### DIFF
--- a/src/js/select2/core.js
+++ b/src/js/select2/core.js
@@ -106,8 +106,11 @@ define([
   };
 
   Select2.prototype._placeContainer = function ($container) {
-    if (this.options.get('insertBefore')) $container.insertBefore(this.$element);
-    else $container.insertAfter(this.$element);
+    if (this.options.get('insertBefore')) {
+      $container.insertBefore(this.$element);
+    } else {
+      $container.insertAfter(this.$element);
+    }
 
     var width = this._resolveWidth(this.$element, this.options.get('width'));
 
--- a/src/js/select2/core.js
+++ b/src/js/select2/core.js
@@ -106,7 +106,8 @@ define([
   };
 
   Select2.prototype._placeContainer = function ($container) {
-    $container.insertAfter(this.$element);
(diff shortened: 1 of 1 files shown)


Bmju:
I have understood that I should be using the errorPlacement() callback in jQuery Validation instead. (http://stackoverflow.com/questions/26498899/jquery-validate-custom-error-message-location)

tanuj-github:
This problem is a generic problem. Cant we provide a setting in options when initialising the  in select2 plugin to place custom select box before/after acc to the need.

I have also been stuck on this since last 3 Hrs...

MikeBeaton:
Hi - I am re-opening this as per the ongoing discussion in https://github.com/select2/select2/issues/4445 - I'm not expecting this to be approved immediately (if at all, obviously!) because it doesn't have a test.

I don't know if anyone from the project could give me a quick pointer as to where similar tests might live? I'm happy to try to add one.

MikeBeaton:
Oh, I can see that my committ for my PR is [still in the select2 repository](https://github.com/select2/select2/commit/1f81c75bc003fcd6d76b871cd20578c41cb2b06e), but I'm not quite sure how I can bring it back to life, or if it's easier to just create a new pull request from scratch?

Issue #4511, opened 2016-08-01
Dropdown does not reposition when multiple selections cause the input to increase in height


tim-lukacik:
## Prerequisites
- [x] I have searched for similar issues in both open and closed tickets and cannot find a duplicate
- [x] The issue still exists against the latest `master` branch of Select2
- [x] This is not a usage question (Those should be directed to the [community](https://select2.github.io/community.html))
- [x] I have attempted to find the simplest possible steos to reproduce the issue
- [ ] I have included a failing test as a pull request (Optional)
## Steps to reproduce the issue
1. Enable multiple selection
2. Set "closeOnSelect: false"
3. Select many items so that selections take up 2 lines or more.
## Expected behavior and actual behavior

When I follow those steps, I see...
The dropdown remains in the same position despite that the actual input has now grown in height. This blocks all selections on line 2 and beyond.

I was expecting...
The dropdown should reposition itself again so that it remains below the expanded input.
## Environment

Browsers
- [x] Google Chrome
- [ ] Mozilla Firefox (Not tested)
- [x] Internet Explorer

Operating System
- [x] Windows
- [ ] Mac OS X (Not tested)
- [ ] Linux (Not tested)
- [ ] Mobile (Not tested)

Libraries
- jQuery version: 2.1.4 (Also tested with 3.1.0)
- Select2 version: 4.0.3
## Isolating the problem
- [ ] This bug happens [on the examples page](https://select2.github.io/examples.html)
- [x] The bug happens consistently across all tested browsers
- [x] This bug happens when using Select2 without other pluigns
- [ ] I can reproduce this bug in [a jsbin](https://jsbin.com/)

devmonkey22:
I have the same issue.  Same versions of jQuery/select2.

derrick13:
I had the same issue and commented this line:

    container.on('select', function () {
      if (!container.isOpen()) {
        return;
      }

      self.setClasses();
      // self.highlightFirstItem(); // $custom (hkl) OTB74270, deactivated see https://github.com/select2/select2/issues/4584
    });

devmonkey22:
Thanks @derrick13 - this will solve a variation of this issue, but didn't solve my main problem.

[...]

PR #4601, opened 2016-09-23
Resolved issue #4511 to reposition open dropdown when a selection is made

devmonkey22:
This pull request includes a
- [X] Bug fix
- [ ] New feature
- [ ] Translation

- In AttachBody bind method, added handler to watch select and unselect events from container.  This allows the dropdown to reposition if the underlying selection object changes positions because it wraps onto another line, or if options are unselected, makes sure it snaps back up below selection object (so it doesn't float).

#4511

### DIFF
--- a/src/js/select2/dropdown/attachBody.js
+++ b/src/js/select2/dropdown/attachBody.js
@@ -32,6 +32,15 @@ define([
           self._resizeDropdown();
         });
       }
+
+	  function repositionOnSelectionChange() {
+	    if (!this.options.get('closeOnSelect')) {
+		  self._positionDropdown();
+	    }
+	  };
+	  container.on("select", repositionOnSelectionChange);
+	  container.on("unselect", repositionOnSelectionChange);
+
     });
 
     container.on('close', function () {
--- a/src/js/select2/dropdown/attachBody.js
+++ b/src/js/select2/dropdown/attachBody.js
@@ -37,9 +37,9 @@ define([
 	    if (!this.options.get('closeOnSelect')) {
 		  self._positionDropdown();
 	    }
-	  };
-	  container.on("select", repositionOnSelectionChange);
-	  container.on("unselect", repositionOnSelectionChange);
+	  }
+	  container.on('select', repositionOnSelectionChange);
+	  container.on('unselect', repositionOnSelectionChange);
 
     });
 


src/js/select2/dropdown/attachBody.js
```diff
@@ -32,6 +32,15 @@ define([
           self._resizeDropdown();
         });
       }
+
+	  function repositionOnSelectionChange() {
```
kevin-brown: The tabbing appears to be off here, as it looks like you are using a combination of spaces and tabs.

src/js/select2/dropdown/attachBody.js
```diff
@@ -32,6 +32,15 @@ define([
...
+	    if (!this.options.get('closeOnSelect')) {
```
kevin-brown: What was the reasoning behind the special case here?

kevin-brown: You may only want to reposition the dropdown if you can detect that the dropdown is still open. I think that may have been what you were going for here, but you can probably solve the issue by looking at the dropdown state instead of checking for a specific option.

src/js/select2/dropdown/attachBody.js
devmonkey22: Ah, yes - apologies.  It looks like your convention is a space indented file, so I'll modify.

src/js/select2/dropdown/attachBody.js
devmonkey22: I was originally trying to only reposition if I knew the dropdown wasn't about to be closed due to the closeOnSelect option.  I had originally incorrectly assumed the container.isOpen might have given me the wrong value depending on the order of the decorator chain, however, it does look like this check works more generically.  I'll modify and commit.

devmonkey22:
@kevin-brown Thanks for the quick review of this PR.  I looked over the positioning-tests.js - it would take me a while to digest it all and write an appropriate unit test.

It looks like I would need to try to combine the approaches in positioning-tests and possibly selectOnClose-tests.js to render the dropdown with a small width, set closeOnSelect to false, add multiple options to cause the selection element to wrap and re-check the position of the dropdown?

I know the value of making sure the tests are in place.  I don't know that I can post a test quickly.  I hate to ask, but would you be able to help write one for this change?

Thanks,
Mike

subssn21:
Can we please get this Pull Request Merged. This was causing a huge problem with our application when we upgraded from 3.x to 4.x, I had to apply this patch to a local copy of select2, but it worked beautifully and solved the problem.

Issue #4604, opened 2016-09-23
With ajax data source, SelectAdapter.prototype.option doesn't set option.value property if data.id is 0 (zero) as a number


devmonkey22:
## Prerequisites
- [X] I have searched for similar issues in both open and closed tickets and cannot find a duplicate
- [X] The issue still exists against the latest `master` branch of Select2
- [ ] This is not a usage question (Those should be directed to the [community](https://select2.github.io/community.html))
- [X] I have attempted to find the simplest possible steos to reproduce the issue
- [ ] I have included a failing test as a pull request (Optional)
## Steps to reproduce the issue
1. Setup an Ajax data source with an item with id: 0
2. Render select2 as multi-select
3. Open results dropdown
4. Try to select the item with id of 0.  Nothing happens in UI.

Using https://output.jsbin.com/folefatufa
1. Load page
2. Start typing "ang"
3. Select the first option (nothing will happen, because the processResults function forces item 0 to have id=0 for testing purposes
4. Select any other option, the selection is accepted and displayed properly.
## Expected behavior and actual behavior

When I follow those steps, I see...
- Selecting the item with id of 0 (as integer) doesn't update real selection.  UI doesn't change.
- It looks like this only affects Ajax adapter.  For arrays, the `SelectAdapter` calls the `_normalizeItem` method which converts id into a string, so the `SelectAdapter.prototype.option` method check of `if (data.id) {} still succeeds.
- For the ajax adapter, the `id` value is not normalized internally, so either the application or json data source would need to normalize it to a string to workaround this issue.

I was expecting...
Item should have gone into current selections.

Should select2 require ids to be normalized (to a string), or should the `SelectAdapter.prototype.option` method simply change its check before setting the `option.value` properly.
## Environment

Browsers
- [X] Google Chrome
- [ ] Mozilla Firefox
- [X] Internet Explorer

Operating System
- [X] Windows
- [ ] Mac OS X
- [ ] Linux
- [ ] Mobile

Libraries
- jQuery version: 3.1.0
- Select2 version: 4.0.3
## Isolating the problem
- [ ] This bug happens [on the examples page](https://select2.github.io/examples.html)
- [X] The bug happens consistently across all tested browsers
- [ ] This bug happens when using Select2 without other pluigns
- [X] I can reproduce this bug in [a jsbin](https://output.jsbin.com/folefatufa)

Will try to add more detail shortly.

PR #4605, opened 2016-09-23
Fix for #4604 to support selecting an item with id=0 values (as number)

devmonkey22:
This pull request includes a
- [X] Bug fix
- [ ] New feature
- [ ] Translation

- Changed `SelectAdapter.prototype.option` to check `if (data.id !== undefined)` instead of just `if (data.id)`

If this is related to an existing ticket, include a link to it as well. #4604

### DIFF
--- a/src/js/select2/data/select.js
+++ b/src/js/select2/data/select.js
@@ -170,7 +170,7 @@ define([
       }
     }
 
-    if (data.id != null) {
+    if (data.id !== undefined) {
       option.value = data.id;
     }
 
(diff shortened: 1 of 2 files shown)


devmonkey22:
Thanks @kevin-brown - is there a preference to move #4605 or #4517 forward, as both may be extra work for everyone?  Figured I'd ask before trying to sort through the test case code.

kevin-brown:
> is there a preference to move #4605 or #4517 forward, as both may be extra work for everyone?

They both get the job done in more-or-less the same way. As this point I'm going to accept the first one to get a test case in, or fall back to #4517 if I need to write it on my own.

PR #4606, opened 2016-09-24
answer question about using html in selection template

SpyMaster356:
This pull request includes a
- [x] Documentation change

- answer question about using html in selection template

### DIFF
--- a/docs/_includes/options/selections/templating.html
+++ b/docs/_includes/options/selections/templating.html
@@ -31,7 +31,8 @@ $('select').select2({
   
 {% highlight js linenos %}
 function template(data, container) {
-  return $('<span>' + data.text + '</span>');
+  return $('<span></span>')
+    .text(data.text);
 }
 
 $('select').select2({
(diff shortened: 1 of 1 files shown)


docs/_includes/options/selections/templating.html
```diff
@@ -27,11 +27,21 @@ <h2 id="templateSelection">
     I am using HTML in my selection template but it isn't displaying it
   </h3>
 
-  {% include options/not-written.html %}
+  If you want to use html in your selection template, you will need to return a jQuery object.
+  
+{% highlight js linenos %}
+function template(data, container) {
+  return $('<span>' + data.text + '</span>');
```
kevin-brown: Ideally, to avoid XSS issues, this should be something like

```js
var $template = $('<span></span>');
$template.text(data.text);
return $template;
```

That way the `data.text` is sanitized before you display it, so you don't need to worry about HTML being injected where you weren't intending it.

PR #4611, opened 2016-09-29
Throw proper errors everywhere

nateous:

- Sometimes `throw new Error(...` was used, other places `throw "...` was used.
- All `throw` calls now use `new Error`


kevin-brown:

PR #4622, opened 2016-10-07
Fix for tags not being enterable via Enter keypress when they match an existing search result

### LINKED ISSUE #4093: I can not create a tag if the text is part of the selected tag
Example - http://codepen.io/alexpts/pen/QyOyWm (3) , select form.

ghost:
This pull request includes a
- [ x ] Bug fix

- Check if an item exists in the query results array that exactly matches the search parameter. If it does and the results:select trigger is fired, it will select that item instead of the highlighted one (which may be a partial match for an existing item, when you really wanted to select a newly-entered tag).

This fixes #4093


subssn21:
Tried applying this patch locally and after applying it, could no longer add any tags under any circumstances. (Completely new name, substring, or exact same as an existing one). I am guessing that some other work that has been done has caused this patch stop working. I recommend it either get closed or the fix updated.

Here is the options I passed to select2 to test with:

```
    $('#program_name').select2({
        placeholder: "Select Program or type new Program Name",
	    //selectOnBlur: true,
        tags: true,
        tokenSeparators: [],
        //selectOnClose: true,
	    createTag: function(params){
	        return {id:params.term, text:params.term};
        },
	    data: ${json(programs)}
    });

```

On Master:HEAD pressing enter on a tag with these options creates a new tag. Typing the name of an already existing tag selects that tag. Typing a substring of the currently selected tag does not create the new tag (The bug this Pull Request is supposed to fix)

kevin-brown:

PR #4624, opened 2016-10-11
Make original select invisible

Steveb-p:
This pull request includes a
- [x] Bug fix
- [ ] New feature
- [ ] Translation

- Visibility CSS attribute of select element is changed to `hidden`

Changes visibility of original select element, so it becomes ignored by Chrome's find in page visibility.  
This targets a specific use-case, when selects with `multiple` attributes are in the page.

Before, when browsing page in Chrome and using built-in find (CTRL+F) values from select were found. You can observe this behaviour in [Select2 examples page](https://select2.github.io/examples.html) - try looking for `California`. With this change, searching for mentioned string results in only two places being selected, which are on visible elements.

### DIFF
--- a/src/scss/core.scss
+++ b/src/scss/core.scss
@@ -42,6 +42,7 @@
   padding: 0 !important;
   position: absolute !important;
   width: 1px !important;
+  visibility: hidden !important;
 }
 
 @import "theme/default/layout";


kevin-brown:
I'm not comfortable accepting this pull request, as while the original issue is valid, setting `display: none` also kills almost all client-side validation as the browser starts to consider it as a hidden, non-user-facing element.

I was able to reproduce the issue you mentioned that triggered this pull request, and I think there would still be value in keeping it around as a separate ticket (linked to this pull request). Once we figure out how we are going to solve the validation and usability issues caused by both solutions, then I'd be comfortable accepting a pull request.

Steveb-p:
@kevin-brown I was also considering effects on assistive technologies, screen readers and so on. However, I'm not too familiar with them - yet - and can't say what effects it would have.

Proposed change is for `visibility`, not `display` (since select2 is already changing it's position to absolute, and reducing it to 1px/1px practically invisible element). Setting it to `display:none` in my case caused issues with select2 itself.

On a side note, this ctrl+f behaviour of Chrome is considered a bug - I've read it somewhere and will look for it again when I get home - but it hasn't been fixed for a couple of years now and is probably here to stay.

I will open an issue for it and link this PR.

Issue #4584, opened 2016-09-14
Higlighted Item Problem [MultiSelect with closeOnSelect=false] 


BonDoQ:
## Steps to reproduce the issue
1. Select an element for the top of the list.
2. Scroll down and select the last Element
## Expected behaviour

dropdown still at the same scroll position
## Actual behaviour

dropdown scroll up to highlight the first "selected" Item
## Environment

Browsers
- [x] Google Chrome
- [x] Mozilla Firefox
- [x] Internet Explorer

Operating System
- [x] Windows
- [x] Mac OS X
- [x] Linux
- [x] Mobile

Libraries
- jQuery version: 3.1.0
- Select2 version: 4.0.3
## Isolating the problem
- [x ] I can reproduce this bug in a ([JSFiddle](http://jsfiddle.net/BonDoQ/rzvemnqy/))
## Could be a Solution

check this condition before trigger that function would prevent the behaviour for that case

[result.js:279](https://github.com/select2/select2/blob/master/src/js/select2/results.js#L279)

``` javascript
if(!(self.options.get('multiple') && !self.options.get('closeOnSelect'))) {
  self.highlightFirstItem();
}
```

also for the unselect

kevin-brown:
Duplicate of https://github.com/select2/select2/issues/4417.

alexweissman:
Please see https://github.com/select2/select2/issues/4417#issuecomment-244741804 to understand where this bug came from.

hemant-rao:
this code working fine for my side, I hope this will help you.

$('#selectmultiple').select2({
    closeOnSelect: false
})
 .on('select2:selecting', e => $(e.currentTarget).data('scrolltop', $('.select2-results__options').scrollTop()))
 .on('select2:select', e => $('.select2-results__options').scrollTop($(e.currentTarget).data('scrolltop')));

WestCraneLi:
my code
```javascript
$('#select2Demo')
  .empty()
  .select2({ closeOnSelect: false, data: data, placeholder: 'select' })
  .on('select2:open', function (e) {
    const scrolltop = $(this).data('scrolltop') || 0
    $('.select2-results__options').scrollTop(scrolltop)
  })
  .on('select2:selecting', function (e) {
    $(this).data('scrolltop', $('.select2-results__options').scrollTop())
  })
  .on('select2:select', function (e) {
    const scrolltop = $(this).data('scrolltop') || 0
    $('.select2-results__options').scrollTop(scrolltop)
[...]

PR #4625, opened 2016-10-11
Higlighted Item Problem [MultiSelect with closeOnSelect=false]

RomanMarkelov:
This pull request includes a
- [x] Bug fix
- [ ] New feature
- [ ] Translation

check this condition before trigger that function would prevent the behaviour for that case

resolve https://github.com/select2/select2/issues/4584


b-franco:
I think the unselect needs the same conditions.

pedrofurtado:
Resolved in https://github.com/select2/select2/pull/5150

PR #4627, opened 2016-10-12
load-grunt-tasks added in the project.

douglasdeodato:
This pull request includes a
- [ x] New feature

- make Gruntfile.js  a bit smaller, we saved a bit of lines.
- Very Easy now if someone install a dependency in the project from package.json, does not need to load this  in the Gruntfile.js

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -324,20 +324,10 @@ module.exports = function (grunt) {
     }
   });
 
-  grunt.loadNpmTasks('grunt-contrib-clean');
-  grunt.loadNpmTasks('grunt-contrib-concat');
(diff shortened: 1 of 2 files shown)


douglasdeodato:
when usually  happen the merges? thanks

kevin-brown:
I'm honestly not sure if the extra dependency is worth it. I actually don't mind the fact that we have to be explicit when loading in Grunt tasks.

> when usually happen the merges? thanks

It usually takes more than a day, we're a small team with other full-time jobs.

douglasdeodato:
@kevin-brown  np, I just want to contribute with the project, even more with scss part, but I understand your approach. 

it is a scss documentation for this project? thank you very much

PR #4639, opened 2016-10-17
Remove unanswered tag from bootstrap modal question

zack6849:
This pull request includes a
- [x] Bug fix
- [ ] New feature
- [ ] Translation

- Removed the unanswered question banner for an answered question

This question is answered, but not marked as such in the docs


kevin-brown:
This question has a short answer, but it doesn't cover all of the common cases and it doesn't actually cover why the `dropdownParent` option works for most people. The tag was left in for anyone who feels like expanding the documentation within that section.

PR #4640, opened 2016-10-17
Fix titles not being cleared (#3895)

### LINKED ISSUE #3895: Tooltip on the select box still appearing after clear.
I have a select box and i have used select2.js to create a select control out of it. My initialization code is :  $("#mycontrol).select2({  });

and i clear it like this $("#mycontrol").select2("val","");

Its getting cleared fine but the tooltip on the select2 dropdown control still displays the older value.
The same thing happens if i use the croos icon to clear the selected value. The tooltip still has the value same as last selected item.

timothykang:
This pull request includes a
- [x] Bug fix

- Remove title attribute on selection clear
- Added test assertions

Fixes #3895.

### DIFF
--- a/src/js/select2/selection/multiple.js
+++ b/src/js/select2/selection/multiple.js
@@ -55,7 +55,9 @@ define([
   };
 
   MultipleSelection.prototype.clear = function () {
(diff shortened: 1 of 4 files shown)


alexweissman:

Issue #4276, opened 2016-03-31
Dropdown closes automatically on small window size


vostreltsov:
Hi there. I've run into a weird issue. On small screens with certain scroll position, the dropdown menu closes automatically.

In my case, this is reproducible when opening a page on an android phone.

I use 3.5.4, but this bug is also reproducible in 4.x (see the screencast attached).
[bug.mp4.zip](https://github.com/select2/select2/files/197204/bug.mp4.zip)

Hope this can be fixed in both 4.x and 3.x

kevin-brown:
Can you expand upon the issue you are facing by adding a few more pieces of information to your bug report?

- What steps can be used to reproduce the issue?
- What is the bug and what is the expected outcome?
- What browser(s) and Operating System have you tested with?
- Does the bug happen consistently across all tested browsers?
- Can you reproduce this issue [on the examples page][examples]?
- What version of jQuery are you using? And what version of Select2?
- Are you using Select2 with other plugins?

The [contributing guide][cg] contains a useful section with a few questions which can help us track down the bug and speed up the process of finding a fix for it.  The creation of [an isolated test case][test-case] would also be useful.

[cg]: https://github.com/select2/select2/blob/master/CONTRIBUTING.md#reporting-bugs-with-select2
[examples]: https://select2.github.io/examples.html
[test-case]: http://css-tricks.com/6263-reduced-test-cases/

vostreltsov:
Steps to reproduce:
1. open the [example page](https://select2.github.io/examples.html#single)
2. resize browser window to a small size and scroll so that select2 is located in the bottom part and drop menu appears above it (see the video in the original post)
3. click on the select2

The bug is that the drop menu is automatically closed, and I expect it not to be closed.

Browsers/OS: chrome, firefox on windows, linux and android.

Bug happens consistently when the window size is small.

Yep, the video is recorded using the examples page.

Jquery version is 1.11.3.

Select2 version is 3.5.4.

michaeldale:
Just adding a +1 to this issue, we are seeing it in multiple browsers/devices.

MaxxSoftware:
+1
We have the same issue.

NightJar:
This is due to the events used to (accurately) emulate a 'native' `<select>` element.

The `mousedown` event on the chosen (or placeholder) element renders the dropdown section. In the test case on the examples page this then focusses the search input element, which centres it in the viewport (for whatever reason it seems to always centre, not just be 'not partially occluded').

[...]

PR #4644, opened 2016-10-20
Fix remove click when multiselect box is higher than browser windows (#4642)

### LINKED ISSUE #4642: Multipleselect: deletion of one element is not possible when select box is higher than browser window
## Prerequisites
1. generate a long multiselect with all selected items
2. make window height greater than select box
3. try to delete any selected item

Example see here: https://jsfiddle.net/vwsg4tn5/

I was expecting that clicking on any items `x` would delete that item
## Environment

Browsers
- [x] Google Chrome
- [ ] Mozilla Firefox
- [ ] Internet Explorer

Operating System
- [ ] Windows
- [x] Mac OS X
- [ ] Linux
- [ ] Mobile

Libraries
- jQuery version: 3.1.1
- Select2 version: 4.0.3
## Isolating the problem
- [ ] The bug happens consistently across all tested browsers
- [x] I can reproduce this bug in [a jsfiddle](https://jsfiddle.net/vwsg4tn5/)

mobilutz:
This pull request includes a
- [X] Bug fix
- [ ] New feature
- [ ] Translation

- I changed the event for the remove callback
- now removing an item is possible when the multiselect box is taller than the browser window (and the box is not yet active)

This fixes #4642 

_NOTE_:
This changes the following behaviour:
When you click on remove, the select dropdown will not open automatically. Which I personally think is better anyways.

### DIFF
--- a/src/js/select2/selection/multiple.js
+++ b/src/js/select2/selection/multiple.js
@@ -26,8 +26,14 @@ define([
 
     MultipleSelection.__super__.bind.apply(this, arguments);
 
+    this.$selection.on('click', function (evt) {
+      self.trigger('toggle', {
+        originalEvent: evt
+      });
+    });
+
     this.$selection.on(
-      'click',
+      'mousedown',
       '.select2-selection__choice__remove',
       function (evt) {
         // Ignore the event if it is disabled
@@ -46,12 +52,6 @@ define([
         });
       }
     );
-
-    this.$selection.on('click', function (evt) {
-      self.trigger('toggle', {
-        originalEvent: evt
-      });
-    });
   };
 
   MultipleSelection.prototype.clear = function () {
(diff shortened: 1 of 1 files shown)


mobilutz:
Could you provide any information when this can become part of select? Thanks in advance.

mobilutz:
@alexweissman How can I help moving this forward?

alexweissman:
@mobilutz there is a major dilemma about using `mousedown` vs `click` right now, in a number of issues and proposed fixes.  Have you reviewed the following issues and PRs?

- https://github.com/select2/select2/issues/4276
- https://github.com/select2/select2/pull/5069

It seems like a lot of people are thinking that an easy fix would be to swap `mousedown` for `click` in various places (or vice-versa), but in reality this could introduce a whole new set of bugs.  Have you already tested this in mobile views?

mobilutz:
@alexweissman Thanks for the response.

I looked into both issues and PRs and saw the discussion there. It does look like there needs to be a general solution for the fix.

As we do not have a mobile view in our application, this is not so important for us. We went the way of forking the repo and implementing the fix there.

Issue #4217, opened 2016-02-26
Documentation for Bootstrap modals


chipco:
## Steps to reproduce the issue
1. Open the modal in this fiddle http://jsfiddle.net/81sr8v4h/2/
2. Click the select input to open the options
3. Try to input something in the search
## Environment

Browsers
- [x] Google Chrome
- [x] Mozilla Firefox

Operating System
- [x] Windows

chipco:
`$.fn.modal.Constructor.prototype.enforceFocus = $.noop;` wasn't a solution for me since it messes tabbing between inputs.

Attaching select2 dropdown to the modal frame seems to work fine for me: http://jsfiddle.net/81sr8v4h/4/

mukesh011:
This issue is also reproducible with any kind of modal popup. I am using knockout modal popup.

boiawang:
@chipco  thanks.

kevin-brown:
I've updated the locked message at https://github.com/select2/select2/issues/600#issuecomment-102857595 to mention `dropdownParent`.

I'm leaving this ticket open because this is something which needs to be documented, as people commonly use Bootstrap with Select2.

daza110:
As per my comment on another thread (https://github.com/select2/select2/issues/4218#issuecomment-213393499) this JS fiddle does not work when scrolling down the modal.
The "$.fn.modal.Constructor.prototype.enforceFocus = $.noop;" fix doent work at all for me

vnt:
I have the same problem!!

wadleo:
I just solved this now, removed the tabindex='-1' on the modal (<div id="myModal" class="modal fade" tabindex='-1' role="dialog" aria-labelledby='myModalLabel' aria-hidden='true'>)

FramnkRulez:
Attaching the dropdown to the modal frame is the best way to solve this problem.  Overriding bootstrap's enforceFocus() function or removing tabindex='-1' will break the keyboard tabbing on bootstrap modal forms.

alexweissman:
This might be a good candidate for a "troubleshooting" or "common problems" section.

alexweissman:
This is now documented at https://select2.org/troubleshooting/common-problems.

PR #4646, opened 2016-10-21
Problem using Bootstrap modals with Select2

wadleo:
This pull request includes a
- [x] Bug fix
- [ ] New feature
- [ ] Translation

- Docs for problems with Bootstrap modals.
- Fix for #4217

Added a solution for problems with using select2 with Bootstrap modals

### DIFF
--- a/docs/_includes/options/dropdown/placement.html
+++ b/docs/_includes/options/dropdown/placement.html
@@ -29,7 +29,11 @@ $('select').select2({
     Use the <code>dropdownParent</code> option, setting it to the modal.
   </p>
 
-  {% include options/not-written.html %}
+  <p>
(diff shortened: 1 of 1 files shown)


alexweissman:
Although this breaks the accessibility to some extent, it would seem that this is a [very popular fix](https://stackoverflow.com/questions/18487056/select2-doesnt-work-when-embedded-in-a-bootstrap-modal/18487440#18487440) so I propose that we merge this in for the time being.

alexweissman:
Actually, I am closing this because there are two much better alternatives:

- https://stackoverflow.com/a/19574076/2970321
- https://stackoverflow.com/a/33884094/2970321

PR #4649, opened 2016-10-24
Trigger 'input' event before 'change' events

marcandre:
This pull request includes a
- [x] Bug fix
- [ ] New feature
- [ ] Translation

- Triggers an `input` event before triggering a `change` event
- Tests looking for `change` event being triggered now look for `input` then `change`.

While not widely known, the `input` event is triggered before all `change` events for inputs and selects (see [the spec here](https://www.w3.org/TR/html5/forms.html#send-select-update-notifications))

FWIW, `Parsley` relies on the `input` event, so the current code is [causing bad interactions](https://github.com/guillaumepotier/Parsley.js/issues/1122)

### DIFF
--- a/tests/data/select-tests.js
+++ b/tests/data/select-tests.js
@@ -167,12 +167,14 @@ test('duplicates - single - same id on select triggers change',
   var data = new SelectData($select, data);
   var second = $('#qunit-fixture .duplicates option')[2];
 
-  var changeTriggered = false;
+  var changeTriggered = false, inputTriggered = false;
 
   assert.equal($select.val(), 'one');
 
   $select.on('change', function () {
-    changeTriggered = true;
+    changeTriggered = inputTriggered;
+  }).on('input', function() {
+    inputTriggered = true;
   });
 
   data.select({
@@ -188,8 +190,13 @@ test('duplicates - single - same id on select triggers change',
   );
 
   assert.ok(
+    inputTriggered,
+    'The input event should be triggered'
+  );
+
+  assert.ok(
     changeTriggered,
-    'The change event should be triggered'
+    'The change event should be triggered after the input event'
   );
 
   assert.ok(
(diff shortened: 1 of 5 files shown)


tests/data/select-tests.js
```diff
@@ -185,8 +187,13 @@ test('duplicates - single - same id on select triggers change',
...
 
   assert.ok(
+    inputTriggered,
+    'The input event should be triggered'
+  );
+
+  assert.ok(
     changeTriggered,
-    'The change event should be triggered'
+    'The change event should be triggered after the input event'
```
kevin-brown: I'm not sure how I feel about enforcing the order of the events within past tests. I can understand adding tests to check that the `input` event is triggered, and adding tests to ensure that one event is caught before the other, but I don't think that requires you to modify old `change`-only tests to cover both those cases.

marcandre:
Thanks for taking a look at this PR :smile:

1) Every `change` event should be preceded by at least one `input` event. Most inputs have a single `input` event triggered just before a `change` event, but some like `type="text"` usually have many `input` events (e.g. after each key / paste / ...) before a single `change` event. Another input like that is the `type="color"`

2) For tests, we could have one test to insure that `change` is triggered, another to check that `input` is triggered, and a third one to insure that they are triggered in the right order, but the first two are made redundant by the third one so that why I felt this was the best solution. Note that the order of the event is mandated by the spec.

Let me know how I can be of further assistance

marcandre:
@kevin-brown Is there something else that is blocking this PR from being merged?

marcandre:
I've rebased my branch.

marcandre:
This PR is really simple, and is still valid.

marcandre:
This PR is really simple, and is still valid.

MarcoSulla:
Please merge this pull request, is simple and [adhere to the standards](https://html.spec.whatwg.org/multipage/indices.html#event-input)

marcandre:
Stale bot: 0
Me: 2

marcandre:
Rebased.

edwinvdpol:
Ow come on, merge already!! 👍

kevin-brown:
Rebased this pull request against the latest dev so the new checks can run. The goal of this is to get it merged in tonight for the next patch release.

marcandre:
Cool, thanks!

Issue #3275, opened 2015-04-21
Select2 loses focus when {closeOnSelect: false}


ArxSargos:
I tried to get through all issues with focus, but didn't found similar or duplicate issue. Sorry if this is solved or already tracked as issue...

When {closeOnSelect: false} is set, after selection made by keaboard select2 loses it's focus. http://jsfiddle.net/522bsupt/  (comparison of standard and closeOnSelect:false option)

piotr-cz:
I must confirm the issue

gryn:
I'd like to confirm that the behavior with closeOnSelect=false doesn't work at all really. I'm working on having select2 act more like a menu with checkboxes, so I am using SingleSelection as my working base for a custom Adapter on a <select multiple>. 

There are at least two issues:

1) Changing selection via keyboard or mouse steals the highlight/focus to be the first selected item. (this is due to the code in results.setClasses -- I have made a monkey patch to work around this issue).

2) Using the keyboard's ENTER will only select a new item (and then lose focus to the first selected item), using ENTER on an already select item *closes* the dropdown (which directly violates the closeOnSelect=false option). Using space+control is technically the correct key press, but with closeOnSelect=false I feel ENTER and space+control should work the same. However, space+control works but still fails to keep focus in the correct place.

In summary, the focus is completely messed up and ENTER should behave differently if closeOnSelect=false.

gryn:
Here is my focus fix hack:
				select.on('select2:selecting', recordFocus);
				select.on('select2:unselecting', recordFocus);
				select.on('select2:select', setFocus);
				select.on('select2:unselect', setFocus);

				var curOption;

				function recordFocus(e) {
					var data = e.params.args.data;

					// select2 doesn't create jquery friendly id's (they may have funny chars)
					// so use getElementById
					curOption = data && data._resultId ?
						$(document.getElementById(data._resultId)) :
						null;
				}
				function setFocus(e) {
					curOption && curOption.trigger('mouseenter').trigger('focus');
				}

gryn:
And I won't post the hack I have done for the keys, since there is no clean way to change the keyboard behavior (it's added globally at initialization and can't be customized or disabled). But effectively I was able to make everything "work" by changing the behavior for KEYS.ENTER to be results:toggle, and also I removed the control key requirement for the space bar (since that's how checkboxes would work and that's how my UI is presented in my case). Being able to customize this behavior would be nice instead of having to hack it. (Just allowing an override function is better than nothing).

[...]

PR #4669, opened 2016-11-03
Fix focus loss with closeOnSelect=false

dbatishchev:
This pull request includes a

- [x] Bug fix
- [ ] New feature
- [ ] Translation

- I changed ensureHighlightVisible function
- Now select with closeOnSelect=false option not loses focus after selection

Fix for [3275](https://github.com/select2/select2/issues/3275)

### DIFF
--- a/src/js/select2/results.js
+++ b/src/js/select2/results.js
@@ -493,9 +493,14 @@ define([
     var currentOffset = this.$results.offset().top;
     var nextTop = $highlighted.offset().top;
     var nextOffset = this.$results.scrollTop() + (nextTop - currentOffset);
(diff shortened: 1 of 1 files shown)


trostli:
I've noticed this bug too, it was not an issue in 4.0.2. I think it was introduced by this change in 4.0.3: https://github.com/select2/select2/commit/481c43883e23874e9c35879d173eb8cc5b994b12

Essentially i think if closeOnSelect is false the selected option should still be in view and highlighted

alexweissman:
Is this problem still occurring as of `4.0.6-rc.0`?

PR #4680, opened 2016-11-10
Fix keyboard not closing when closing dropdown on iOS 10

willmacdonald:
This pull request includes a

- [x] Bug fix
- [ ] New feature
- [ ] Translation

- Add  `self.$search.blur();` 

We wanted to have a select2 dropdown with search field that had auto focus, so you could type directly.

The problem we experienced was that the keyboard would not hide after choosing from the filtered list.

We only had this problem on iOS 10 devices. This fixes the problem.

Credit: @EvertLagerberg


alexweissman:
PRs with changes in `dist/` cannot be merged.  Please remove them to have this PR considered.

alexweissman:

PR #4681, opened 2016-11-10
Add native event dispatching

alissonbovenzo:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

- Add native dispatcher event support for change events
- Add support to angular 2 event binding syntax

``` html 
<select 
    #select 
    class="select2"
    (change)="onSelectChange($event)">. . . </select> 
```
Event changed for listen nativately with the DOM

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -14,7 +14,8 @@
     "QUnit": false,
     "require": false,
     "test": false,
-    "window": false
+    "window": false,
+    "CustomEvent": false
   },
   "indent": 2,
   "maxlen": 80,
--- a/src/js/select2/results.js
+++ b/src/js/select2/results.js
@@ -448,10 +448,14 @@ define([
         return;
       }
 
-      self.trigger('select', {
+      var payload = {
         originalEvent: evt,
         data: data
-      });
+      };
+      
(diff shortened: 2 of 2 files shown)


TCMiranda:
:+1:

leonanluppi:
👍

kevin-brown:
It's not clear to me why using `CustomEvent` works when the standard for jQuery plugins is to trigger events with `.trigger('change')`, [which we do in a few places](https://github.com/select2/select2/blob/563198a4a2e0eb9f7adbc4775d360db1037a13db/src/js/select2/data/select.js#L66).

TCMiranda:
@kevin-brown so you think that we should use the CustomEvent api to power to whole component?
Sorry if I misunderstood.

Shouldn't it be better to dispatch a singular event only to allow Angular2 bindings to it?
That's good that internally it uses `.trigger` since its a jQuery plugin. 

The reason for the CustomEvent is that Angular2 can only bind events natively with the `(event)=""` syntax when real DOM Events are dispatched by the component which jQuery does not dispatch.

tedliang:
IE don't support 'new CustomEvent(...)'.
http://caniuse.com/#feat=customevent

Issue #195, opened 2012-07-06
select all / select none header


ethanj:
it would be great to have this type of functionality added...

see
http://www.erichynds.com/examples/jquery-ui-multiselect-widget/demos/#headers

kevin-brown:
This can be done by overriding the `dropdownAdapter` and decorating the `render` method l[ike we do for the search](https://github.com/select2/select2/blob/473267404aeee8c4380ac2ba4558016872236398/src/js/select2/dropdown/search.js).

We've decided to not include the proposed `formatHeader` and `formatFooter` methods because of this.

ChetnaN:
Solution given by MortadaAK and sieppl are working fine in Firefox, but not working on Chrome.
Changed "Keypress" to "Keydown" for Chrome, now its calling the event, but still the functionality not working.
Please help.

kevin-brown:
I finally had some time to get around to creating a demo with the decorator that adds a "Select all" button to the dropdown.

http://jsbin.com/seqonozasu/1/edit?html,js,output

shorif2000:
this should be added to the docs, very useful feature and i am sure a lot of people need this. also some easy to find demo's, I have been searching a long time before i came across this thread

shorif2000:
How can i use this without using requires amd loader?

bkdotcom:
@shorif2000 

the amd loader is included with select2

what's not obvious is how to make the adapter reusable

	/*
          Define the adapter so that it's reusable
	*/    	   
	$.fn.select2.amd.define('select2/selectAllAdapter', [
		'select2/utils',
		'select2/dropdown',
		'select2/dropdown/attachBody'
	], function (Utils, Dropdown, AttachBody) {

		function SelectAll() { }
		SelectAll.prototype.render = function (decorated) {
			var self = this,
				$rendered = decorated.call(this),
				$selectAll = $(
					'<button class="btn btn-xs btn-default" type="button" style="margin-left:6px;"><i class="fa fa-check-square-o"></i> Select All</button>'
				),
				$unselectAll = $(
					'<button class="btn btn-xs btn-default" type="button" style="margin-left:6px;"><i class="fa fa-square-o"></i> Unselect All</button>'
				),
				$btnContainer = $('<div style="margin-top:3px;">').append($selectAll).append($unselectAll);
			$rendered.find('.select2-dropdown').prepend($btnContainer);
			$selectAll.on('click', function (e) {
				var $results = $rendered.find('.select2-results__option[aria-selected=false]');
[...]

PR #4701, opened 2016-11-27
Simplify plugin integration

tedliang:
This pull request includes a

- [x] New feature

- Support decorator options: dataDecorator, resultsDecorator, dropdownDecorator and selectionDecorator

It will simplify third-party plugin integration. User can specify decorator which will decorate the default adapter. Don't need to use $.fn.select2.amd.require and construct adapter manually using Utils.Decorate.

### DIFF
--- a/src/js/select2/defaults.js
+++ b/src/js/select2/defaults.js
@@ -67,6 +67,11 @@ define([
         AdapterClass = chains.shift();
       }
 
+      var decorator = options[type + 'Decorator'];
+      if (decorator != null) {
+        chains.push(decorator);
+      }
+
       $.each(chains, function (i, DecoratorClass) {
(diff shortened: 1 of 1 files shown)


tedliang:
For example, the ['Select all' button plugin](http://jsbin.com/seqonozasu/1/edit?html,js,output)  for #195 can be simplified like this:
```
function SelectAll() {}

SelectAll.prototype.render = function (decorated) {
  var $rendered = decorated.call(this);
  ...
}

$("select").select2({
  placeholder: "Select Option(s)...",
  dropdownDecorator: SelectAll
});
```
User don't need to know how to construct the adapter which can be quite complicated.

tedliang:
@jpic Could you please review and merge this?

PR #4706, opened 2016-12-02
Updates option docs with little hints

flobee:
This pull request includes a

- [x] Docs improvement

- adds some notes where to find data structure when using ajax

Kind regards
Florian


alexweissman:
Thanks! I've added something similar to the docs.

Issue #2957, opened 2015-01-22
Problem with declaring callbacks in the data-* attributes


akozhemiakin:
What I like about select2 is the idea to specify all configuration parameters in data-\* attributes.

Suppose I have some sort of templating function stored in my namespace:

``` javascript
window.myNamespace = {
    templateResult: function(item) {
        if (item.loading) return item.text;
        return '<div>' + item.text + ' (' + item.price + ')' + '</div>';
    }
}
```

Now I want to specify this function as a callback in my select tag in data-template-result attribute like that:

``` html
<select class="select2" data-template-result="myNamespace.templateResult"></select>
```

Clean and cool but does not work. The reason is, as we can read from the Jquery docs: "Every attempt is made to convert the string to a JavaScript value (this includes booleans, numbers, objects, arrays, and null)". No function in this list. So, now there is no legal way to declare callbacks using data-\* attributes.

For now I'm using the following workaround: 
I'm prefixing all my callback names with '_fn::' like this: "_fn::myNamespace.templateResult". Then, before activating any select tag with select2(), I'm using the following function to let jQuery know about callbacks declared in data-\* attributes:

``` javascript
function(obj) {
    var getFunctionByName = function (functionName, context) {
        context = typeof context === 'undefined' ? window : context;
        var namespaces = functionName.split(".");
        var func = namespaces.pop();
        for(var i = 0; i < namespaces.length; i++) {
            context = context[namespaces[i]];
        }
        return context[func];
    }
    var $obj = $(obj);
    $.each($obj.data(), function(key, el) {
        if (typeof el === 'string' && el.substr(0, 5) == '_fn::') {
            $obj.data(key, getFunctionByName(el.substr(5)));
        }
    });
}
```

Works just fine, but looks like a hack.

Maybe we can do something with select2 behavior itself to let users legally specify callbacks in data attributes without using such external hacks?

donaldinou:
agreed :+1:

meridius:
I'd also like to see this implemented. Until then here are my two cents to this since I had a hard time to copy-paste use the solution @numesmat posted, so I tweaked it a bit.

```html
<select data-template-selection="_fn::userManagementComponent.formatUserName">
...
</select>
```

```javascript
// somewhere close to the <select> in question since the 
// templating function is supposed to be coined for specific use just with that one element
window.movieDb.userManagementComponent = {};
window.movieDb.userManagementComponent.formatUserName = function(item) {
	return (item.text === "") 
[...]

PR #4715, opened 2016-12-11
Adds an option to pass a global function to templateResult

vladkucherov:
Hello there,

Small improvement I made so we'll be able to pass global functions into templateResult option.
That way I will be able to pass a string to the templateResult, and if a global function exists in that name, it would be fired instead of returning an error.

Now we can just add a data attribute to a certain element like this:
`data-template-result="foo"`
and it will try to find a function with that name:
```
function foo(data) {
    // Do your stuff
    return data.text;
}
```

But of course things may be even more convenient for large projects. Like this for example:
```
var Main = function () {
    window.foo = bar

    var bar = function (data) {
        // Do more stuff here
        return data.text;
    }
}

$(function () {
    Main();
});
```

Let me know what you think.

### DIFF
--- a/src/js/select2/results.js
+++ b/src/js/select2/results.js
@@ -505,10 +505,15 @@ define([
   };
 
   Results.prototype.template = function (result, container) {
+    var content;
     var template = this.options.get('templateResult');
     var escapeMarkup = this.options.get('escapeMarkup');
 
-    var content = template(result, container);
+    if (typeof window[template] == 'function') {
+        content = window[template](result, container);
+    } else {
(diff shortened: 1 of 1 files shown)


Karmalakas:
Would be nice to have the same for `templateSelection`

knightq:
Seems like a duplication of issue #2957 fixed by @th3byrdm4n with 5142d31, isn't it?

Any chance to get this solved on 4.0.6?

dwasyl:
Any chance of reviving this?

Issue #4505, opened 2016-07-27
Filtering for optgroups should show their contents


ypnos-web:
## Prerequisites
- [x] I have searched for similar issues in both open and closed tickets and cannot find a duplicate
- [x] The issue still exists against the latest `master` branch of Select2
- [x] This is not a usage question (Those should be directed to the [community](https://select2.github.io/community.html))
- [x] I have attempted to find the simplest possible steos to reproduce the issue
- [x] I have included a failing test as a pull request (Optional)
## Steps to reproduce the issue
1. Go to https://select2.github.io/examples.html#select-boxes-with-labels
2. Enter "pac"
## Expected behavior and actual behavior

Currently, it will find the resp. optgroup and only show that. This is almost useless, as an optgroup cannot be selected anyways.

Instead, the optgroup with all its associated options should be shown, so that the user can filter for the group and then select the desired option.
## Environment

Browsers
- [x] Google Chrome

Operating System
- [x] Linux

Libraries
- jQuery version: 3.1.0
- Select2 version: 4.0.3
## Isolating the problem
- [x] This bug happens [on the examples page](https://select2.github.io/examples.html)

SgtOddball:
Thats not a bad idea, but how will it cope with having `pac` in a different group? 

I would propose that it's be better instead to completely ignore the optgroups instead.

ypnos-web:
What I am suggesting is that whenever an optgroup matches, all its corresponding options are treated as being matched (and therefore shown under the optgroup). So there is no conflict with multiple matches in optgroups or options (which now and then shows the option under its by-itself nonmatching optgroup).

However I agree that for some scenarios people might want to disable matching of optgroups all together.

SgtOddball:
But what if what you want to match on is a partial for an optgroup as well as match to an option in another optgroup?

It would be confusing for the users since they've got no way to select that they want to match the optgroup values or not.

kevin-brown:
This can be handled through a custom `matcher`. It's very close to the [original matcher](https://github.com/select2/select2/blob/3f8492b62f72b34f6a25ed5ae8ffc9a6826cb69d/src/js/select2/defaults.js#L309-L352), just you wouldn't need to [strip out non-matching child objects](https://github.com/select2/select2/blob/3f8492b62f72b34f6a25ed5ae8ffc9a6826cb69d/src/js/select2/defaults.js#L327-L330).

chrispcb:
Like jquery chosen plugin no ?
[...]

PR #4722, opened 2016-12-19
#4505 Filtered optgroups show their children

yappkahowe:
This pull request includes a

- [ ] Bug fix
- [ ] New feature
- [ ] Translation
- [x] Others

- Filtering for optgroups should show their contents as mentioned in issue #4505

### DIFF
--- a/src/js/select2/defaults.js
+++ b/src/js/select2/defaults.js
@@ -312,6 +312,14 @@ define([
         return data;
       }
 
+      var original = stripDiacritics(data.text).toUpperCase();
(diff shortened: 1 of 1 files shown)


markusweb:
please reopen this issue. 
could we make it optional whether to match all entries in optgroup or not. 
i have tested the pull request https://github.com/select2/select2/pull/4722 with my version 4.0.0 and it seems to work. as expected. when i search for the text in optgroup i get all the entries in this optgroup, even other select options outside the optgroup if text matches

markus

PR #4729, opened 2016-12-22
Added an option that disables scrolling to the parent item when a child is highlighted

tzellman:
Fix is in 1d45ed0b38e195ce4b4ba873fa96edea70584b32.

I was experiencing an odd issue occurring when child options were highlighted, especially in cases where there were many children. The highlight code was constantly causing a scroll to occur, disabling from actually reaching elements toward the bottom of the list.

This highlights the bug (using select2/stable/3.5): https://jsfiddle.net/shfs7s8e/3/
This highlights the fix (using tzellman/stable/3.5) https://jsfiddle.net/22rzn48g/3/

Apologies for not making the pull request sooner - slipped off my radar. Thanks!

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,6 @@
 {
     "name": "select2",
-    "version": "3.5.3",
+    "version": "3.5.4",
     "main": ["select2.js", "select2.css", "select2.png", "select2x2.png", "select2-spinner.gif"],
     "dependencies": {
         "jquery": ">= 1.7.1"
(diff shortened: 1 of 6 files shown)


tzellman:
Out of curiosity, is anyone maintaining the 3.5 branch for further bug fixes?

alexweissman:
@tzellman not really, although if enough bug fixes for 3.5 pile up, I might consider doing a patch release!

PR #4738, opened 2017-01-05
Dropdown list width shouldn’t round

masrodjie:
When main dropdown box width is not round,  the dropdown width list is round since using jQuery outerWidth function always return round value, and the result of dropdown box and dropdown list is not exact same width.

### DIFF
--- a/src/js/select2/core.js
+++ b/src/js/select2/core.js
@@ -129,7 +129,7 @@ define([
     }
 
     if (method == 'element') {
-      var elementWidth = $element.outerWidth(false);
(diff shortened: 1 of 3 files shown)


jpic:
Check out the build output on travis:

```
    src/js/select2/results.js
    401 |      params.element.css('width', this.$container[0].getBoundingClientRect().width + 'px');
                                                                                                   ^ Line is too long.
```

alexweissman:
This is actually a duplicate of #4471.

PR #4745, opened 2017-01-11
Show same options in example as used in demo

FROGGS:
This pull request includes a

- [X] Bug fix
- [ ] New feature
- [ ] Translation

- The templateSelection option was used in the demo, but was not shown in the example code.
This commit fixes it, so that copy&pasting works as expected.


jpic:
Correct, the example looks different when using the example
2017-01-30-221235_757x89_scrot

Thanks !

PR #4758, opened 2017-01-19
Fix ajax.abort() detection to handle transport.status code as integer

desclapez:
This pull request includes a

- [x] Bug fix
- [ ] New feature
- [ ] Translation

- Actually the transport.status property is handled as string but it fails when this property comes as as integer. Tested on OSX - Chrome 55.0.2883.95

Original commit cfb66f5


desclapez:
Any update on this?

alexweissman:
See #4356.

Issue #4555, opened 2016-08-29
Tags Don't Wrap Correctly with AllowClear


vol7ron:
http://jsfiddle.net/vol7ron/y9g9swsr/
## Prerequisites
- [x] I have searched for similar issues in both open and closed tickets and cannot find a duplicate
- [x] The issue still exists against the latest `master` branch of Select2
- [x] This is not a usage question (Those should be directed to the [community](https://select2.github.io/community.html))
- [x] I have attempted to find the simplest possible steps to reproduce the issue
- [ ] I have included a failing test as a pull request (Optional)
## Steps to reproduce the issue
1. 535px container with 1.3rem font-size
2. Select options that are just wide enough to cause a problem, but not wide enough to force a newline
## Expected behavior and actual behavior

When I follow those steps, I see...
that the tags don't wrap nicely

I was expecting...
tags to wrap nicely
## Environment

Browsers
- [ ] Google Chrome
- [x] Mozilla Firefox
- [x] Internet Explorer

Operating System
- [x] Windows
- [ ] Mac OS X
- [ ] Linux
- [ ] Mobile

Libraries
- jQuery version: 1.12.x
- Select2 version: 4.0.3
## Isolating the problem
- [ ] This bug happens [on the examples page](https://select2.github.io/examples.html)
- [x] The bug happens consistently across all tested browsers
- [ ] This bug happens when using Select2 without other pluigns
- [x] I can reproduce this bug in [a jsbin](https://jsbin.com/)

RobertDiebels:
There is a workaround for this mentioned in this issue: https://github.com/select2/select2/issues/4470
So yours is a duplicate of that one.

alexweissman:
Closing as duplicate of #4470.

Issue #4470, opened 2016-07-09
allowClear will cause selections shifted to the next line in multiple select


JLHwung:
## Prerequisites
- [x] I have searched for similar issues in both open and closed tickets and cannot find a duplicate
- [x] The issue still exists against the latest `master` branch of Select2
- [x] This is not a usage question (Those should be directed to the [community](https://select2.github.io/community.html))
- [x] I have attempted to find the simplest possible steos to reproduce the issue
- [ ] I have included a failing test as a pull request (Optional)
## Steps to reproduce the issue
1. Add an array of seven 3-digit choices, use `allowClear` option and `multiple` option, set the parent width to 600px;
2. Select all the choices
## Snippet (Talk is cheap)

http://codepen.io/JLHwung/pen/YWAGjG
## Related issues
#3906
## Expected behavior and actual behavior

When I follow those steps, the 6th choice looks really bad.

I was expecting all the choices look good.
## Environment

Browsers
- [x] Google Chrome
- [x] Mozilla Firefox
- [ ] Internet Explorer

Operating System
- [ ] Windows
- [x] Mac OS X
- [ ] Linux
- [ ] Mobile

Not tested on other operating system but I believe it is a general CSS problem.

Libraries
- jQuery version: 2.1.4
- Select2 version: 4.0.3
## Isolating the problem
- [ ] This bug happens [on the examples page](https://select2.github.io/examples.html)
- [x] The bug happens consistently across all tested browsers
- [x] This bug happens when using Select2 without other pluigns
- [x] I can reproduce this bug in [a codepen](http://codepen.io/JLHwung/pen/YWAGjG)

myDisconnect:
Having the same issue. Have anyone managed to find a workaround/fix?

JLHwung:
I do find a CSS workaround (written in SASS for brevity) if you add `data-allow-clear` attribute to the `select` DOMNode.

```
<select data-allow-clear="true"></select>
<script>
$('select').select2()
</script>
```

This issue got no attention so I am not encouraged to give a PR.

```
select[data-allow-clear="true"] ~ .select2-container .select2-selection--multiple {
  .select2-selection__rendered {
    padding-right: 24px;
    .select2-selection__clear {
      position: absolute;
      right: 0;
    }
  }
}
&[dir="rtl"] {
  select[data-allow-clear="true"] ~ .select2-container .select2-selection--multiple {
    .select2-selection__rendered {
      padding-left: 24px;
      .select2-selection__clear {
[...]

Issue #3906, opened 2015-11-05
Bad inline layout at certain width with select multiple


alxndrsn:
# Config

Version: 4.0.0
# Demo

<img width="475" alt="screen shot 2015-11-05 at 17 15 40" src="https://cloud.githubusercontent.com/assets/191496/10974138/e7e2338e-83e0-11e5-9a29-dfe39264ac21.png">
# Steps to recreate
1. open select2 _Examples_ at https://select2.github.io/examples.html
2. scroll down to _Multiple select boxes_ example
3. execute `$(".js-example-basic-multiple").select2({ allowClear: true });` in js console
4. select _Alaska_ and then _Hawaii_
5. right-click on _Hawaii_, and choose _Inspect Element_
6. select the `<li>` element for _Hawaii_ in _Inspector_/_Elements_ debug tab
7. for the selected element, set `width: 249px`
8. observe weird layout
# Notes

This example may seem contrived, but will happen IRL if content lengths are correct, e.g.:

<img width="693" alt="screen shot 2015-11-05 at 14 06 43" src="https://cloud.githubusercontent.com/assets/191496/10974183/1658b30a-83e1-11e5-9be8-56f33652af1e.png">

kevin-brown:
This has to do with how we place the "clear" button (as the first element), and how we never really designed it to be used with a multiple select (but it "works").

RobertDiebels:
https://github.com/select2/select2/issues/4470 has a fix for this.

PR #4767, opened 2017-01-27
Fix_allowClear_bad_allign_bug

chrisNom:
Fixes 3 issues #4470, #4555 and #3906

This pull request includes a

- [x] Bug fix
- [ ] New feature
- [ ] Translation

- modified the css class for and added bottom-margin 5px .select2-container--default .select2-selection--multiple .select2-selection__clear

### DIFF
--- a/src/scss/theme/default/_multiple.scss
+++ b/src/scss/theme/default/_multiple.scss
@@ -30,6 +30,7 @@
     font-weight: bold;
     margin-top: 5px;
     margin-right: 10px;
+    margin-bottom: 5px;
   }
 
   .select2-selection__choice {


jpic:
Works for me, why did you choose 5px for the bottom margin ? Is it just to reflect the top margin or is there any other reason ? Another user used 10px in their fork https://github.com/pculture/unisubs/commit/d2474675c94b473cb2c5257e0e1fe543446f46ed I'm wondering why but can't find a reason.

Also, can you get rid of the extra commit please ?

bendk:
I picked 10px mostly just because that's what I saw.  It also seemed to line up with the line-height, which seemed good to me.  I'm sure it's not the only, or the best, way to do it though.

chrisNom:
I picked 5px to be symmetrical to the top 5px margin. 10px will work but would broaden the margin between lines of multiselect, and I did not want to have it.

jpic:
@bendk does that make sense for you ? will it be ok for you if we change it
to 5px in the core ?

bendk:
Sorry for not replying to this, but 5px sounds fine by me.

Issue #3917, opened 2015-11-10
Similar Error #3016 - 'allowClear' with 'data-placeholder'


maikealame:
Hey,

I was reading about error #3016 and realized the same error occurs when there is a 'data-placeholder' attribute on the element without the parameter in the call of select2, as shown below:

 # No errors

``` js
$("#select").select2({
    ajax: { ... },
    allowClear: true,
    placeholder: " something..."
});
```

``` html
<select id="select"></select>
```

 # Will generate an error

> Uncaught TypeError: Cannot read property 'id' of undefined 
> AllowClear._handleClear  @   select2.full.js:1754
> (anonymous function) @   select2.full.js:571
> (anonymous function) @   select2.full.js:1714
> m.event.dispatch @   jquery-1.11.3.min.js:4
> r.handle @   jquery-1.11.3.min.js:4

``` js
$("#select").select2({
    ajax: { ... },
    allowClear: true
});
```

``` html
<select id="select" data-placeholder="something..."></select>
```

grandehombre:
I've encountered the same problem and have used the following mod to fix it, until the next release comes out

```
  AllowClear.prototype._handleClear = function (_, evt) {
    // Ignore the event if it is disabled
    if (this.options.get('disabled')) {
      return;
    }

    var $clear = this.$selection.find('.select2-selection__clear');

    // Ignore the event if nothing has been selected
    if ($clear.length === 0) {
      return;
    }

    evt.stopPropagation();

    var data = $clear.data('data');

    for (var d = 0; d < data.length; d++) {
      var unselectData = {
        data: data[d]
      };

      // Trigger the `unselect` event, so people can prevent it from being
      // cleared.
      this.trigger('unselect', unselectData);

      // If the event was prevented, don't clear it out.
      if (unselectData.prevented) {
        return;
      }
    }


// ----------------------------------- temporary fix
	if (typeof this.placeholder !== 'undefined') {
		this.$element.val(this.placeholder.id).trigger('change');
	} else {
		this.$element.val("").trigger('change');
	}
// ----------------------------------- end of temporary fix

    this.trigger('toggle', {});
  };

```

joshi1983:
I created a pull request #4776 that uses @grandehombre fix.  It uses the single quotes required by the project's coding style and leaves out the temporary fix comments.

santaclaus21:
[...]

Issue #3817, opened 2015-10-07
Placeholder cutting text off when initialized inside hidden element


brunoDegaspare:
I can see that this is an old issue that still has not been fixed.

I am using the latest version and still have this issue. I added the following css however it breaks select2 on IE.

```
.select2-container .select2-search--inline {
    float: none !important;
}
```

Any solution for this case?
![screen shot 2015-10-07 at 15 12 28](https://cloud.githubusercontent.com/assets/7714748/10339968/db2283a2-6d05-11e5-9bc6-64f375de4c34.png)

kevin-brown:
This is _usually_ an issue with Select2 not properly setting the width somewhere.

I also thought this had a duplicate that was already open, but I wasn't able to find it after a quick search.

vakopian:
Here is a jsfiddle to reproduce this (including a workaround): https://jsfiddle.net/agq08j5z/1/
Notice that the control is hidden initially (`display:none` on the parent div). So it seems that if the control is hidden the internal input's width is not set correctly by select2. If you remove `display:none` everything works.

catmeme:
I'm experiencing this bug when initializing select2 in a hidden tab. I'm not sure if this issue is related.

tanloverluke:
@vakopian thank you VERY MUCH the workings of your workaround, though I had to take the width of an input field above the select2. strange but it works anyway.

alexweissman:
There are now two competing proposed solutions for this bug:

- #4776
- #4898

These need someone to review them.

vluzrmos:
My workaround was trigger the event `change` when the `select` becomes visible:

```javascript
//please use your own selector/context....
$('select').trigger('change');
```

callmejed:
This still seems to be an issue, at least for me. I did notice that the .select2-search__field **parent** has a float: left, that when disabled fixes the width issue. Not sure if that contributes. I'm not really sure what it's accomplishing; I didn't dive into that part too deeply.

sp193:
This issue seems to be changing slightly over the years. Or we are observing the effects of multiple problems.

For me, I have issues with select2 assuming the wrong width if belongs to a hidden parent: https://jsfiddle.net/NozomiTypeR/6pLzy1ak/44/
In this fiddle, there are 2 similar select2 controls. The working one is always visible, while the hidden one can be revealed with the button.
[...]

PR #4776, opened 2017-02-03
fixed bug with undefined this.placeholder in allowClear.js and update…

joshi1983:
…d corresponding output files

This bug and fix was mentioned in #3817

This pull request includes a

- [x] Bug fix
- [ ] New feature
- [ ] Translation

Like the commit says, I applied a fix mentioned in the comments of that issue.  It fixed problems for me and I wanted to share the fix with any other select2 user.

-[https://github.com/select2/select2/issues/3917](https://github.com/select2/select2/issues/3917)

### DIFF
--- a/src/js/select2/selection/allowClear.js
+++ b/src/js/select2/selection/allowClear.js
@@ -60,7 +60,11 @@ define([
       }
     }
 
-    this.$element.val(this.placeholder.id).trigger('change');
+    if (typeof this.placeholder !== 'undefined') {
+        this.$element.val(this.placeholder.id).trigger('change');
+    } else {
+        this.$element.val('').trigger('change');
+    }
 
     this.trigger('toggle', {});
   };


jpic:
Nice patch ! Can you add a test case ? :guitar:

dist/js/select2.js
```diff
@@ -1802,7 +1802,11 @@ S2.define('select2/selection/allowClear',[
       }
     }
 
-    this.$element.val(this.placeholder.id).trigger('change');
+    if (typeof this.placeholder !== 'undefined') {
+        this.$element.val(this.placeholder.id).trigger('change');
+    } else {
+        this.$element.val('').trigger('change');
+    }
```
jpic: Please remove changes in dist: https://github.com/select2/select2/blob/master/CONTRIBUTING.md#submitting-a-pull-request

dist/js/select2.full.js
```diff
@@ -1802,7 +1802,11 @@ S2.define('select2/selection/allowClear',[
       }
     }
 
-    this.$element.val(this.placeholder.id).trigger('change');
+    if (typeof this.placeholder !== 'undefined') {
+        this.$element.val(this.placeholder.id).trigger('change');
+    } else {
+        this.$element.val('').trigger('change');
+    }
```
jpic: Change in dist/ better left off: https://github.com/select2/select2/blob/master/CONTRIBUTING.md#submitting-a-pull-request

dist/js/select2.js
joshi1983: Why is the distribution tracked when it is generated off other sources?

jpic: We don't want it to create conflicts with other PRs, it'll be updated later. Looking forward to review the test case :guitar:

dist/js/select2.full.js
joshi1983: dist directory changes have been removed.

joshi1983:
I removed the changes to the dist directory but creating a test for the bug in #3917 might take a while for me because I'd have to dig deeper into this bug and how the tests work while being pretty busy.  I tried to trace the code and tests/selection/allowClear-tests.js seems like a good place for one but it could be a couple weeks.  If that's too long, can someone else build take my commit and finish off the test case?

jpic:
No problem, I'd rather wait two months now, than see a regression in two months and have you redo your patch ^^

hxnan:
The same error I met. 
I can not find this fix in the master now(2017-11-25 18:13).
Is this fix has't merge to the master or another fix ?

joshi1983:
@hxnan it isn't in master yet.

You could make the requested test and submit a new pull request that is more merge-able to help this fix get into master.

joshi1983:
Hopefully someone can finish my fix for this so the issue doesn't get closed by the github bot.  It still fixes the bug even though there's no automated test on it.

PR #4662, opened 2016-10-30
Docs for: What should the results returned to Select2 look like?

sedrubal:
This pull request includes a
- [ ] Bug fix
- [ ] New feature
- [ ] Translation
- [x] Documentation

The following changes were made
- Added docs for chapter `What should the results returned to Select2 look like?`

If this is related to an existing ticket, include a link to it as well.

PR #4814, opened 2017-03-01
Add docs on results from remote data source

SebCorbin:
This pull request includes

- [ ] Bug fix
- [ ] New feature
- [ ] Translation
- [x] A much needed documentation addition

See also #4706

### DIFF
--- a/docs/_includes/options/data/ajax.html
+++ b/docs/_includes/options/data/ajax.html
@@ -19,7 +19,33 @@
     What should the results returned to Select2 look like?
   </h3>
 
-  {% include options/not-written.html %}
+  <p>
+    In short a JSON object with this structure should be returned:
+  <p>
+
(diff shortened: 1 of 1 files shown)


alexweissman:
There appear to be multiple PRs addressing this section:

#5002 
#4863 
#4861 
#4814
#4662 

We should pick one of these, or integrate several to provide the most comprehensive answer.

alexweissman:
What does `error: null` do in the context of Select2?  Also, someone suggested that `id` needs to be an integer and start at 1.  Is this actually true?

alexweissman:
At any rate, I have merged some of the other PRs for this section, and it is now reflected on the live docs.  See http://select2.github.io/options.html#what-should-the-results-returned-to-select2-look-like

PR #4817, opened 2017-03-02
Simplified Norwegian(nb) translation to a more correct text

bovan:
This pull request includes a

- [ ] Bug fix
- [ ] New feature
- [x] Translation

- Simplified norwegian text to a more correct one

### DIFF
--- a/src/js/select2/i18n/nb.js
+++ b/src/js/select2/i18n/nb.js
@@ -12,13 +12,7 @@ define(function () {
     inputTooShort: function (args) {
       var remainingChars = args.minimum - args.input.length;
 
-      var message = 'Vennligst skriv inn ';
-
-      if (remainingChars > 1) {
-        message += ' flere tegn';
-      } else {
-        message += ' tegn til';
-      }
+      var message = 'Vennligst skriv inn ' + remainingChars + ' tegn til';
 
       return message;
     },


ToreOlavKristiansen:
Hi.

I have also modified this file. The word "til" implies "more" so it might be better to go for this:

`inputTooShort: function (args) {
      var remainingChars = args.minimum - args.input.length;

      var message = 'Vennligst skriv inn minst ';

      if (remainingChars > 1) {
        message += t + ' tegn';
      } else {
        message += ' et tegn til';
      }

      return message;
    }`

What do you think?

ToreOlavKristiansen:
Please disregard my comment. I reverted my changes and used same as you. Yours is closer to the current English version.

alexweissman:
This appears to conflict with #4896.  Could @bovan and @freost please discuss this to determine which is more correct?

bovan:
@alexweissman The rule in Norwegian is that small numbers are written as a word, for that reason I like the idea to use "ett" instead of "1".

However, I do have a feeling that if you also should write "two" and "three" ("to", "tre") if you are using words, which is the reasoning behind my commit (also it's easier to read)

I'm not sure which is more correct of the two solutions, but both are better than the original

alexweissman:
Ok, so from what I understand, it looks like #4896 is a good alternative to your commit, but if you wanted to add some logic for "to" and "tre", that would be even better?

bovan:
Looked it up to be sure..  

> Typically, letters in small numbers (0-12) are used in running text:
>   Nils is five years old. He has three football balls.
> But sometimes you want to emphasize the numerical value or be more precise , and then it's natural to use numbers

I'm fairly sure the "emphasize the numerical value" could be used as an argument to not use "ett", but instead keep 1

freost:
I don't really care myself as long as the message is understandable and correct so #4817 gets my vote.

ToreOlavKristiansen:
#4817 gets my vote too.

alexweissman:
Alright, we'll get this merged in with the next release!

Issue #3347, opened 2015-05-12
Enabled/disabled state not respected on options


virgiliu:
http://jsfiddle.net/upgrmrL3/1/

When clicking the disable button all options get disabled. When clicking "enable" the options in the select2 dropdown still appear as disabled.

![screen shot 2015-05-12 at 13 53 06](https://cloud.githubusercontent.com/assets/426097/7585719/43e5a544-f8ae-11e4-8f10-5012acf9e9b2.png)

Sometimes the disable attribute is not respected either: by clicking disable the options in the dropdown won't get disabled.

![screen-shot-2015-05-12-at-13 54 13](https://cloud.githubusercontent.com/assets/426097/7585797/b22f185a-f8ae-11e4-913a-aa5158340451.png)

The screenshots were taken on the same fiddle, with some cache clearing and refreshes between them.
Testing was done on Firefox 37.0.2 on Mac OS

kevin-brown:
There's a duplicate of this somewhere, and it is a bug caused by us caching the `disabled` state of results.

vyrotek:
This is affecting us as well. Our work around was to destroy it and recreate it but it's not very elegant.

kevin-brown:
I couldn't find the duplicate (guess it never existed), but the issue can be fixed similar to the `selected` property.

https://github.com/select2/select2/commit/54154c3df6b7e2f59dc76b663f230f9b773397f2

Also, this is not limited to multiple selects.

advokata:
Is there any workaround for this issue ?

virgiliu:
Destroying the select2 element and instantiating it again should fix it.

chrisspen:
The workaround is hide/disable the option(s) and then call `.select2()` on the select again. It's annoying but it works.

SaharZehavi:
so is there a solution for this problem? 
for some unknown reason i can't re-call select2 on the element.

michmzr:
I created a hack  for hiding disabled optgroups:
```javascript
   $.initialize(".select2-results__option[role='group']", function(index, optionGroup)
    {
        let optGroupName = $(this).attr("aria-label");

        let select2ResultsId =$(this).closest("ul.select2-results__options").attr("id");

        if(select2ResultsId)
        {
            let ptrn = /^select2-([\w]+)-results/i;
            let res =ptrn.exec(select2ResultsId);

            if(res.length > 0)
            {
                let relSelect2Id = res[1];

[...]

PR #4818, opened 2017-03-03
Fix disabled option not updating when property changes

Jabulba:
This fixes the problem where disabled options couldn't be re-enabled.

This pull request includes a

- [X] Bug fix
- [ ] New feature
- [ ] Translation

- Modified condition to consider the element's current disabled state instead of the cached one for disabled elements. If no element exists the cached value will be used.

Fix for #3347

### DIFF
--- a/src/js/select2/results.js
+++ b/src/js/select2/results.js
@@ -175,8 +175,8 @@ define([
       'aria-selected': 'false'
     };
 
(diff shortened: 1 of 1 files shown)


Jabulba:
I never used this kind of workflow before, my bad on the line break. I tried creating a test for this specific issue but I still know very little of QUnit unfortunately.

PR #4830, opened 2017-03-13
Add multiple selection specification to docs

marcaparent:
This pull request includes a

- [ ] Bug fix
- [ ] New feature
- [ ] Translation

- Adding the `multiple` specification to the docs 

### DIFF
--- a/docs/_includes/examples/tags.html
+++ b/docs/_includes/examples/tags.html
@@ -3,13 +3,14 @@
 
   <p>
     Select2 can be used to quickly set up fields used for tagging. 
-    <code>multiple</code> have to be enabled on the field for tags support.
   </p>
 
   <p>
     Note that when tagging is enabled the user can select from pre-existing
     options or create a new tag by picking the first choice, which is what
     the user has typed into the search box so far.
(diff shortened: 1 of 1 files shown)


subssn21:
Multiple does not have to be added to a select to enable tags support. Single select tags (which allows you to select an already entered option or create your own are allowed)

Recommend closing the pull request as the documentation changes would be inaccurate

marcaparent:
You are indeed right, thanks for this clarification. 

I still think some improvement is possible in the documentation, as it would make sense to explain both possibilities. I've been faced more than once with trying to setup a multiple tags select input and always forget to include the multiple attribute. I'll edit my pull request to clarify, up to you to merge or not.

subssn21:
I would clarify that the multiple attribute is for the HTML, perhaps an example too just so people aren't confused that you are not saying that it is a config option on select2.

Once that's done I would recommend that this gets merged.

marcaparent:
I added an example for a single select element, to keep the default multiple example first.

alexweissman:
I think this can be integrated fairly easily. I'll rewrite it a bit, to clarify exactly what "tagging" is.

PR #4844, opened 2017-03-21
Correct the inputTooShort message

Jir4:
This pull request includes a

- [ ] Bug fix
- [ ] New feature
- [x] Translation

- The french translation for the `inputTooShort` message give the false idea that this limit is strict. Add the 'or more' equivalent that was missing.

### DIFF
--- a/src/js/select2/i18n/fr.js
+++ b/src/js/select2/i18n/fr.js
@@ -24,6 +24,8 @@ define(function () {
         message += 's';
       }
 
+      message += ' ou plus';
+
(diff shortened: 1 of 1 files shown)


elchristo:
@Jir4 Don't you think it would be better french to say "Saisissez au moins n caractère(s)" ?

alexweissman:
Looks like this has been dealt with in recent changes to `fr.js`.  Please feel free to discuss if you believe that additional changes to the French translation are needed.

PR #4850, opened 2017-03-29
Create af.js

ettiennelouw:
Added afrikaans language file

This pull request includes a

- [X] Translation


alexweissman:
This appears to have some syntax errors which is causing the build to fail.  Furthermore, there is another PR for Afrikaans at #4998.  Could @ettiennelouw and @pjjroux please resolve this together?

Issue #1908, opened 2013-11-15
Standard (native) JS events are no longer emitted in v4


t-nelis:
We'd like to see Select2 emit _standard_ events targeting the original `<select>` element.

Specifically, we needed `focus` and `keydown` events for use in our implementation of a field navigation component that helps us define application-specific navigation paths between fields declaratively _without using tab indices_.

While we realize that Select2 emits custom events, including `select2-focus`, we don't wish to special-case some code in our field navigation component just for Select2 (we use a few other types of controls). Instead, it'd be great if Select2 would emit the standard events _from the original `<select>` element_ alongside its custom events, whenever a one-to-one mapping exists and when Select2 doesn't have any counterpart event type.

[List of DOM3 Event Types](http://www.w3.org/TR/DOM-Level-3-Events/#event-types-list) (W3C DOM Level 3)
[Standard Events](https://developer.mozilla.org/en-US/docs/Web/Reference/Events#Standard_events) (MDN Web Technology Reference)
[Select2 Events](http://ivaynberg.github.io/select2/#events) (Select2 Documentation)

Note that Select2 doesn't emit any keyboard events targeting the original `<select>` element, so this makes our use-case impossible to implement in any kind of _supported_ way.

For those who may have the same issue, we've successfully worked around it; although reading this code is not a healthy thing to do. I'll post it below in the hope that:
1. It helps in understanding the issue and finding an appropriate solution.
2. Someone can think of a more straightforward workaround in the meantime (or point us toward something that would remove the need for any workaround).
3. It helps someone else who may have difficulties working around the issue.

There are a few things specific to our codebase (jQuery, AMD) but it should be fairly understandable. We're trying to make a habit of documenting hideous hacks properly, so I include the lengthy comment intended for internal documentation, which should be useful.

Thank you.

```
define(function (require) {
    var $ = require('jquery');
    var key = require('keymap');
    require('select2');

    /*
     * Here be dragons.
     *
     * As of November 2013, select2 doesn't seem to play nice with the
     * other kids.  More specifically, it doesn't make use of standard
     * events on the select element.
     *
     * Instead, it provides control-specific events prefixed with
     * `select2-' [1].  Since we don't want to special-case the
     * navigation setup code just for select2, we won't assign handlers
     * for these events.  Instead, we'll add handlers for standard
     * events on the select element, and trigger the corresponding
     * actions on the control.
     *
     * This works fine for the focus event.  Since all select2 controls
[...]

PR #4855, opened 2017-04-04
Adds a `select2:blur` event back to select2

Spotswood:
Following from the discussion in https://github.com/select2/select2/issues/1908 this PR adds the `blur` event at the `EventRelay` list.


digitalstaub:
why this is not merged?

Issue #4570, opened 2016-09-08
Lost focus after selecting a value with v4.0.3


CrazyTuna:
Using the v4.0.3 every time I select a new value, the component lose the focus.

I've create two JSFiddle:
- [select2 v4.0.0](https://jsfiddle.net/823azsnd/4/)
- [select2 v4.0.3](https://jsfiddle.net/L1sahbx9/2/)

Using the v4.0.3, the focus is lost after selecting a value using mouse click and keyboard.

Maybe related to this issue:
- [Autofocus submit button after closing not working](https://github.com/select2/select2/issues/4388).

Thanks.

SgtOddball:
It's related to this issue #4384 and there's a couple of workarounds listed in there too.

kikonen:
Encountering this issue also myself; have to try if those workarounds work

almirb:
Hello,

It will be solved when merging this PR:
https://github.com/select2/select2/pull/4427/commits/294e7e664d95eafb7f41264c2b87412ad9aa48fc

I tested It and works!

alexweissman:
@almirb it would appear that [#4958](https://github.com/select2/select2/pull/4958/files) also claims to fix this problem.  Could you please pull and test it when you get a chance?  It would appear that it fixes a few other problems with IE tagging as well.

alexweissman:
This has been fixed as of 4.0.4.  See #4427.

Issue #4398, opened 2016-06-02
Focus is lost while entering tags in Internet Explorer (IE11)


luetge:
## Prerequisites
- [x] I have searched for similar issues in both open and closed tickets and cannot find a duplicate
- [x] The issue still exists against the latest `master` branch of Select2
- [x] This is not a usage question (Those should be directed to the [community](https://select2.github.io/community.html))
- [x] I have attempted to find the simplest possible steos to reproduce the issue
- [ ] I have included a failing test as a pull request (Optional)
## Steps to reproduce the issue
1. Go to https://select2.github.io/examples.html with IE11
2. Open the autocomplete in the section "Tagging support"
3. Hit one letter, eg "b", wait a few seconds, then continue typing
## Expected behavior and actual behavior

When I follow those steps, the search field loses focus during waiting, typing has no effect, when clicking on the text to restore focus the text disappears. This completely prevents tagging functionality in this browser. FF&Chrome&Safari work.

I was expecting to be able to use tagging.
## Environment

Browsers
- [ ] Google Chrome
- [ ] Mozilla Firefox
- [x] Internet Explorer

Operating System
- [x] Windows
- [ ] Mac OS X
- [ ] Linux
- [ ] Mobile

Libraries
- jQuery version: 1.11.2
- Select2 version: 4.0.3
## Isolating the problem
- [x] This bug happens [on the examples page](https://select2.github.io/examples.html)
- [x] The bug happens consistently across all tested browsers
- [x] This bug happens when using Select2 without other pluigns
- [ ] I can reproduce this bug in [a jsbin](https://jsbin.com/)

kevin-brown:
This sounds like either it might be a regression or nobody has noticed it in the last year. Would it be possible for you to test if this issue exists in 4.0.0 - 4.0.2 as well?

If it's a regression, it should be considerably easier to fix than if this issue has always existed.

luetge:
Good idea, I just checked and it worked all the way up to 4.0.2-rc.1, 4.0.3 is the first tag where the problem occurs. Actually I just bisected the issue and it turns out the following commit introduced the problem: ea79a197e0ffe55aa600eed6d18cbd1c804c3176. The commit is not very long, hopefully you can quickly see what goes wrong. Thanks for your help!

fundrepertory:
fixed after re-focus input

paulshoughton:
We have seen this problem in IE 11 also.  A fix for this would be hugely appreciated.

[...]

PR #3394, opened 2015-05-21
Prevent IE from navigating back on backspace using .focus()

nfriend:
This fixes #3168.

The `evt.preventDefault()` statement isn't preventing the browser's default back behavior when backspace is pressed because focus is lost to the `<body>` element (in IE).  This change re-focuses the current `<input>` element on a backspace keypress.

PR #4860, opened 2017-04-07
Fixes IE11 issue with select losing focus having selected an item.

ToreOlavKristiansen:
This pull request includes a

- [ x] Bug fix
- [ ] New feature
- [ ] Translation

- Changed from calling focus on the search input element to the select element.

If this is related to an existing ticket, include a link to it as  #4570 #3332.


ToreOlavKristiansen:
I closed the pull request because one of the automated tests failed.

ToreOlavKristiansen:
I did more testing and I don't see any issue even if the test fails. The input field might not have focus, but the select has had its focus call and thus everything works.

src/js/select2/selection/search.js
```diff
@@ -175,7 +175,7 @@ define([
 
     this.resizeSearch();
     if (searchHadFocus) {
-      this.$search.focus();
+      this.$element.focus();
```
jeznag: Will this have side effects on other types of inputs? I did another PR which only modifies the behaviour for tag input as a precaution. https://github.com/select2/select2/pull/4958/files

ToreOlavKristiansen: $element.focus() invokes the focus function of the select2 plugin, not the "tag". This function should handle all cases so I don't think it has any side effects. We are using select elements in our solutions and the bug needs fixing for that too.

Did I understand you correctly?

alexweissman: It seems that there are several competing solutions to this problem:

https://github.com/select2/select2/issues/4398#issuecomment-281721604
https://github.com/select2/select2/pull/4958/files
https://github.com/select2/select2/commit/294e7e664d95eafb7f41264c2b87412ad9aa48fc

alexweissman:
@ToreOlavKristiansen @jeznag can either of you confirm that this is still an issue in Select2 v4.0.5?  We fixed a more glaring issue with `focus` in 4.0.4., so I'm not sure if this is just another manifestation of that problem, or a separate issue specific to IE.  Thanks!

ToreOlavKristiansen:
@alexweissman it is fixed in v4.0.5

ToreOlavKristiansen:
@alexweissman it is improved in v4.0.5, but it is still a problem for multi-select: focus is lost when selecting an item in IE11. It works fine in Chrome.

alexweissman:
Alright, I'll merge this in!

alexweissman:
Actually, we need to fix the broken build.  If you believe that the failing tests do not reflect accurately on the expected behavior, then we need to change either your code or the tests 😛 

I notice that #4958 is handling the focus correctly, so maybe for your solution to work, we just need to add some logic for `searchHadFocus`?  We can still leave out their check for `isTagInput` if you believe that is not actually necessary.

ToreOlavKristiansen:
I will test some more tomorrow. In my test earlier today I did not alter base.js and it looked like Chrome was ok without it. If that is the case, I will add another pull request.

ToreOlavKristiansen:
I tested. We need both changes. I modified to please the unit tests. I also fixed an obvious duplicate text issue in the Norwegian translation.

alexweissman:
Awesome, thank you!  Not sure what happened with the Norwegian translation, I must have performed a bad merge conflict resolution 😕

ToreOlavKristiansen:
We use the select2 control a lot and really appreciate all your hard work on it. I am glad to contribute. Thanks :)

alexweissman:
Merged into `develop` for the 4.0.6-rc.1 pre-release.

alexweissman:
Actually, one more question - did you try just checking `searchHadFocus` instead of `isTagInput`?

ToreOlavKristiansen:
I did. Then everything appears to be working fine, but the unit tests fail on a check for "search" having focus.

alexweissman:
Ah ok, I understand.  Everything is wrapped in `if (searchHadFocus)` anyway 👍

alexweissman:
@ToreOlavKristiansen I'm still seeing a possibly related problem in #3168 - can you take a look?

ToreOlavKristiansen:
@alexweissman I do get this issue in IE11. I wouldn't say it is related to the lost focus issue though.

I see that the selected items are cleared first and then IE gets the backspace key event and goes back. Should we check for a backspace key event when clearing the items and prevent it from bubbling up?

alexweissman:
We could try that, but I'd like to know why this only seems to affect IE 😕

ToreOlavKristiansen:
I see that having used backspace, it looks like the control has focus, but is not possible to start typing in Chrome. So the search textbox does not have focus. IE then probably figures since no input field has focus and goes back when it gets the backspace. The issue could be fixed by calling this.$element.focus(); having cleared the items. Then it is perhaps possible to start typing too.

alexweissman:
It would seem that there are other issues related to IE losing focus too easily (#2503, for example).  Maybe it has ADHD 🙂 

At any rate, #3394 claims to fix this.  Can you check it out?

ToreOlavKristiansen:
No, that does not fix it.

I was able to fix it by adding this code as shown in the image:
        ```
evt.preventDefault();
var self = this;
window.setTimeout(function () {
    self.$element.focus();
}, 0);

```
image

ToreOlavKristiansen:
The prevent default on the event fixes the IE browser back issue. The focus fixes not being able to start typing for both IE and Chrome.

alexweissman:
Interesting...I wonder what #3394 is trying to achieve, then.  Seems like it is doing something similar, but in `search.js` instead of `allowClear.js`.

ToreOlavKristiansen:
Let me know if you want a pull request for this.

alexweissman:
Yes, that would be great!

ToreOlavKristiansen:
Here it is: #5161

I made the mistake of including the dist files. Will you be able to sort that out?
I don't use git often. I see that all my changes are included in the pull request, not just the allowClear.js modifications. Please let me know if I am doing something wrong here.

alexweissman:
Thanks!  Yes, it would be more helpful if the PR didn't include changes to `dist`.  However, if they are in a separate commit, I can just `cherry-pick` the src commits.

nehachaudhary:
@ToreOlavKristiansen The fix you gave in 5161 doesn't work either.

PR #4861, opened 2017-04-07
Adding sample JSON object to Ajax documentation

adw99:
This pull request includes a

- [ ] Bug fix
- [ ] New feature
- [ ] Translation

Hi, I'm just trying to add a sample JSON object to the Ajax documentation - it would have saved me a little time when I was looking for it. The change is in docs\_includes\options\data\ajax.html. Please take a look and let me know what you think.

### DIFF
--- a/docs/_includes/options/data/ajax.html
+++ b/docs/_includes/options/data/ajax.html
@@ -19,7 +19,28 @@
     What should the results returned to Select2 look like?
   </h3>
 
-  {% include options/not-written.html %}
+  <p>
+    A result object with two select options and more available via pagination would look like this:
+  </p>
+
+{% highlight json linenos %}
+{
+  "results": [
+    {
+      "id": "1",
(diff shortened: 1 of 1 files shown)


TheRealGitCub:
It would be also worth adding that the `id` attribute is required, and must start at 1.

alexweissman:
There appear to be multiple PRs addressing this section:

#5002 
#4863 
#4861 
#4814
#4662 

We should pick one of these, or integrate several to provide the most comprehensive answer.

alexweissman:
Not sure what `"status" : 200` does in the context of Select2, but the pagination info is useful.

alexweissman:
Thanks, this has been merged with some of the other PRs for this section, and is now reflected on the live docs.  See http://select2.github.io/options.html#what-should-the-results-returned-to-select2-look-like

Issue #1513, opened 2013-07-09
MultiSelect, closeOnSelect:false - Jumps to top of list each selection


zambodi:
When using a select with mode of multiple and closeOnSelect as false, the drop down jumps to the top of the list with each selection.  This requires the user to scroll back down to where they were in the list to find the next selection.  Search works well; however, you need to know what you're searching for.  Sometimes, as a user, you don't know what choices your list contains. 

Here is a jsFiddle:
http://jsfiddle.net/WP97R/1/

JahBrewski:
Hmm. I am still getting this issue in 3.5.1 -- It is very frustrating for my client to have to scroll back down to where they were after each selection.

kevin-brown:
Reopening this pending further debugging.

kevin-brown:
Confirmed as fixed in 3.5.2 again.

BonDoQ:
This Bug is happening again with version 4.0.3

[http://jsfiddle.net/BonDoQ/rzvemnqy/](http://jsfiddle.net/BonDoQ/rzvemnqy/)

Problem: 
[https://github.com/select2/select2/blob/master/src/js/select2/results.js#L279](results.js:279)
It always highlight the  firstSelectedItem / firstItem

Solution: 
check this condition before trigger that function would prevent the behaviour for that case

```javascript
if(!(self.options.get('multiple') && !self.options.get('closeOnSelect'))) {
  self.highlightFirstItem();
}
```

NewLIST:
I solved this with the following:

```javascript
var scrollTop;
$('#mySelect2').on("select2:selecting", function( event ){
    var $pr = $( '#'+event.params.args.data._resultId ).parent();
    scrollTop = $pr.prop('scrollTop');
});
$('#mySelect2').on("select2:select", function( event ){
    var $pr = $( '#'+event.params.data._resultId ).parent();
    $pr.prop('scrollTop', scrollTop );
});
```
Perhaps not the most elegant solution, but essentially we listen for the selecting event and then grab the current scrollTop of the selectable options panel. On the actual selection, we reset the panel's scrollTop back to its original. This happens so fast that you should see no jump in the control window. Only test with Select2 v. 4.x.

The nice thing about this solution is that you don't have to hack the component or include anything in your config functions.

jackson-sandland:
[...]

Issue #4417, opened 2016-06-13
when use multi-select with [closeOnSelect: false] each time you select option the menu scroll to the first option. v4.0.3


AWasseem:
when use multi-select with [closeOnSelect: false] each time you select option the menu scroll to the first option .
then  scroll again down to select another option its do not keep the last position i stood on v4.0.3
example:
1- select (a)

![image](https://cloud.githubusercontent.com/assets/18052887/16017614/fb29c024-31a0-11e6-9e23-f84e55f79079.png)

2- after select (i) the menu scroll to (a) again

![image](https://cloud.githubusercontent.com/assets/18052887/16017606/f621d63e-31a0-11e6-8220-6ebfcf816917.png)
## Environment

Libraries
- Select2 version:4.0.3

jgbishop:
This bug is a byproduct of the commit to fix #4238. I'm not sure what the best course of action on this is, but I'd argue that the auto-highlight feature should be adjusted to work either only in the infinite scroll case (while the popup is open and the user is scrolling), or via a user-set option.

EstherAF:
Totally agree with @jgbishop here. I don't know enough about this project to try to fix this by myself and make a pull request.
If @AlexanderLeonov is not answering, @kevin-brown may have an opinion about how to tackle this issue...
By now, I'm downgrading to 4.0.2, where this issue is not happening.

AlexanderLeonov:
@jgbishop, you are not entirely correct. I've walked through the commits and discovered that this change was actually introduced by the commit [9f58128](https://github.com/select2/select2/commit/9f581285d88128b29a01fc1e5fd2d445d610b553), particularly by adding `this.ensureHighlightVisible();` call in the `highlightFirstItem()`. What is more interesting is that the behavior had not actually changed a lot - the first item had been highlighting even before my commit, it just hadn't been forced to scroll into view until @kevin-brown added above call.
So, I guess the solution would be to remove `this.ensureHighlightVisible();` call, but I'm not sure why it was done in the first place.
@kevin-brown, can you please comment this moment?

iangus:
The issue stems from the call to self.highlightFirstItem() on line 279 of results.js. This was introduced in commit [e897d00](https://github.com/select2/select2/commit/e897d008a672da262ba84cee2a144578696ada29) as self.focusFirstItem(). @kevin-brown @AlexanderLeonov, why is this call made on select? Is there any reason for highlighting the first selected element on selection? Removing this call will solve the issue.

[...]

PR #4869, opened 2017-04-11
Prevent Scroll to First Selection on Select/Unselect

MicahBrown:
Resolves issue for multiselects using `closeOnSelect: false` that caused the list of results to scroll to the first selection after each select/unselect

This pull request includes a

- [x] Bug fix
- [ ] New feature
- [ ] Translation

- Removed behavior for scrolling to first selection after each selection/unselection
-
-

https://github.com/select2/select2/issues/1513

### DIFF
--- a/src/js/select2/results.js
+++ b/src/js/select2/results.js
@@ -112,6 +112,8 @@ define([
       // in the dropdown
       $options.first().trigger('mouseenter');
     }
+
+    this.ensureHighlightVisible();
   };
 
   Results.prototype.setClasses = function () {
--- a/src/js/select2/results.js
+++ b/src/js/select2/results.js
@@ -112,8 +112,6 @@ define([
       // in the dropdown
       $options.first().trigger('mouseenter');
     }
-
-    this.ensureHighlightVisible();
   };
 
   Results.prototype.setClasses = function () {
--- a/src/js/select2/results.js
+++ b/src/js/select2/results.js
@@ -276,7 +276,6 @@ define([
       }
 
       self.setClasses();
-      self.highlightFirstItem();
     });
 
     container.on('unselect', function () {
@@ -285,7 +284,6 @@ define([
       }
 
       self.setClasses();
-      self.highlightFirstItem();
     });
 
     container.on('open', function () {


os1r1s110:
Just a suggestion, instead of commenting out the feature for the highlightFirstItem prototype, maybe you could simply comment the line where it highlights the first item on select (line 277 in your version of the code now)?
This way when closing/reopening the selectbox, you get the first item in sight, which IMO should be the case. What is your opinion about that?

MicahBrown:
@os1r1s110 I revised it.

alexweissman:
@MicahBrown is there any indication in the commit history as to why `self.highlightFirstItem();` was added in the first place?  I want to be sure that removing these lines doesn't cause a regression in other features.  A test or two to assert the behavior that this patch fixes, would also be appreciated.

MicahBrown:
@alexweissman Unfortunately, I don't know if I'll have time to write any tests since I'm very busy now. I did do a quick look-through on the commit history and found that the `self.highlightFirstItem()` was first introduced in this commit: https://github.com/select2/select2/commit/e897d008a672da262ba84cee2a144578696ada29 (it's been renamed from `self.focusFirstItem()`). When it was first introduced, it didn't have the `this.ensureHighlightVisible()` call, which is what's causing the issue now and bringing the first item into view. I don't see why we would want `self.highlightFirstItem()` called on the `select` or `unselect` triggers, so that's why I took it out of those triggers and left the `results:all` one. Hopefully that helps!

Also, we've been using my change in production for a few months now and haven't noticed any regressions. I know that doesn't cover all cases, but thought I'd mention it.

alexweissman:
No problem, and thanks for getting back to me so quickly.

It would appear that `highlightFirstItem` was added for the infinite scroll (pagination) feature, as per @jgbishop's comment:

> I'd argue that the "highlight first item" feature should be adjusted to work either only in the infinite scroll case (while the popup is open and the user is scrolling), or via a user-set option.

alexweissman:
Please see https://github.com/select2/select2/issues/4417#issuecomment-244741804 to understand why we cannot simply remove the `self.highlightFirstItem()` calls entirely.  We need to add some logic so that `self.highlightFirstItem()` is only called when infinite scroll (pagination) is used.

alexweissman:
I notice that #4625 also adds some conditional logic for scrolling to the first item, to handle some other use cases.

It almost seems like maybe "scroll-after-select" should be a configurable option?

alexweissman:
Please continue the discussion for this feature in #5150.

Issue #4535, opened 2016-08-16
Elements with ARIA roles must ensure required owned elements are present (1)


jpggvilaca:
Accessibility Developer tools throws a severe error on this one.

"Elements with ARIA roles must ensure required owned elements are present (1)"

It's the AX_ARIA_08 error, explained here:
https://github.com/GoogleChrome/accessibility-developer-tools/wiki/Audit-Rules

Can someone fix this? thanks.

apokryfos:
Just to add to this: according to https://www.w3.org/TR/wai-aria/roles#combobox (which I think is the specification) when using `role='combobox'` elements with roles of listbox and textbox must be owned. I checked the HTML generated and could only find the textbox role. What I would expect to be the listbox is actually a tree.

apokryfos:
If you do intend to keep working on this library then I don't think you should let this issue die off as stale since it's an accessibility issue and it's not good to overlook those.

kevin-brown:
Kicking this again since it's stale. This is still an issue, but it might be wrapped up under https://github.com/select2/select2/issues/3744.

kevin-brown:
I'm going to make my life easier and wrap this up under https://github.com/select2/select2/issues/3744, where it is one of the many things that we need to do. If you're interested in helping to make Select2 more accessible, we accept pull requests.

PR #4881, opened 2017-04-24
Add role and aria-readonly attributes to single selection dropdown value

michalkleiner:
This pull request includes a

- [x] Bug fix
- [ ] New feature
- [ ] Translation

- added `role="textbox"`
- added `aria-readonly="true"`

Fix for #4535.


alexweissman:

Issue #4870, opened 2017-04-13
'The results could not be loaded.' message during typing when request was cancelled by select2


eternalfame:
During the typing in select2 searchbox with 'ajax' option on in Chrome browser, select2 sends requests to server, and some of them it cancels. When it happens, the message 'The results could not be loaded.' appears in dropdown, then gets replaced with results. Seems a bit confusing.

Haven't tested in other browsers.

This bug happens also on the examples page (https://select2.github.io/examples.html)

Found such code in select2.full.js:
```
        // Attempt to detect if a request was aborted
        // Only works if the transport exposes a status property
        if ($request.status && $request.status === '0') {
          return;
        }
```
The condition here could be replaced with `($request.status && $request.status === '0' || $request.status === 0)` condiiton, but I can't be sure than it doesn't break something else

eternalfame:
bug appears also in Firefox

BigChief45:
@eternalfame I started experiencing this all of a sudden. Seemed to work fine before.\

Did you find any solution?

nelson6e65:
I'm having this issue too. Specially noticeable in slow connections.

kikonen:
Seeing this also, and given workaround fixes it. It seems that in JQuery status is 0, not "0", which triggers issue.

lordplagus02:
Is there a PR for this yet?

alexweissman:
@lordplagus02 there is a fix in #4356, but it needs some test cases before it can be merged in.

lordplagus02:
is it possible that there's also an issue with the `delay` option? It seems to actually ignore the delay for me, displaying the aforementioned "error" message before I'm done typing.

lordplagus02:
My bad, just used the code from the PR, **definitely fixes the issue**.

alexweissman:
#4356 has been merged into the 4.0.6-rc.0 pre-release.

kikonen:
I just tried out "4.0.6-rc.0" and I'm still seeing these errors

raptor235:
Hey guys, 

I'm still getting this issue... just downloaded latest version of ACF I'm seeing the following 

http://www.flowpress.com/screenshots/screenshot_2018-02-14_at_6.50.21AM.png

ajax call returns the data

http://www.flowpress.com/screenshots/screenshot_2018-02-14_at_6.50.56AM.png

Any thoughts?

canilao:
[...]

PR #4882, opened 2017-04-24
Fix: #4870 — fixed response status check

eternalfame:
refer to https://github.com/select2/select2/issues/4870


denodster:
your fix doesn't consider if status is undefined, I think it needs to be:
```
if ($request.status && ($request.status === 0 || $request.status === '0')) {
```

alexweissman:
This appears to be a duplicate of https://github.com/select2/select2/pull/4356.  Please contribute to that PR, if possible.  It really just needs someone to write a few tests and then it can get merged in!

PR #4364, opened 2016-05-18
perform deep merge for Defaults.set()

nobodyman:
This pull request includes a
- [x] Bug fix
- [ ] New feature
- [ ] Translation

The following changes were made
- Changed Defaults.set() behavior to perform a "deep" merge,  so that it is possible, for example, to change multiple ajax defaults (e.g. "ajax.dataType"  and "ajax.delay")  without one call overwriting the other.
- Added test to assert new behavior

PR #4887, opened 2017-04-27
Set defaults recursively

semenyukdmitry:
Without the patch, when you set defaults like this:

```
$.fn.select2.defaults.set("ajax--delay", 250)
$.fn.select2.defaults.set("ajax--minimumResultsForSearch", 2)
```

you get defaults == `ajax: { minimumResultsForSearch: 2 }` (`delay` is dismissed).

This pull request includes a

- [x] Bug fix
- [ ] New feature
- [ ] Translation

- Set default options recursively (so that you can set `ajax` defaults)

### DIFF
--- a/src/js/select2/defaults.js
+++ b/src/js/select2/defaults.js
@@ -387,7 +387,7 @@ define([
 
     var convertedData = Utils._convertData(data);
 
-    $.extend(this.defaults, convertedData);
+    $.extend(true, this.defaults, convertedData);
   };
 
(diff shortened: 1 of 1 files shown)


alexweissman:
I'm not quite convinced that this is a bug.  Isn't `set` _supposed_ to override, rather than extend?

alexweissman:
Actually, this is a duplicate of #4364, except that #4364 also includes some additional tests.

Issue #3669, opened 2015-08-11
Truncated placeholder for hidden multiple select


SkiF91:
So, here is example: https://jsfiddle.net/spcnek5u/

After firing 'change' event everything ok (of cource if select visible).

kevin-brown:
This has to do with Select2 not being able to get the width when it is hidden because the browser does not actually report the width of hidden elements (they have no width).

This appears to be a duplicate of https://github.com/select2/select2/issues/3292, and it's not something we can (cleanly) fix.

gwilliamtam:
I fix it modifying the width after document ready:

```
$(document).ready( function() {
    // Correct bug to show placeholder
    $('.select2-search__field').css({"width":"200px"});
});
```

Issue #3292, opened 2015-04-27
Placeholder width is 0px when select2 is initialized


brandonhall:
We have a Backbone app which inits `select2` before it is visible on the page. As a result, the `select2-search__field` width is always `0px` because that's the `innerWidth` when it's initialized. This caused placeholders to be hidden.

Here's the relevant block of code in `select2`...

```
 if (this.$search.attr('placeholder') !== '') {
  width = this.$selection.find('.select2-selection__rendered').innerWidth();
} else {
  var minimumWidth = this.$search.val().length + 1;
  width = (minimumWidth * 0.75) + 'em';
}
```

As a result, the solution was to use `setTimeout` which initialized select2 after it was visible on the page
but that causes all sorts of bouncing around on the page as the elements are initialized.

I added an OR condition to use 100% if innerWidth is 0 to prevent this. Like so: `width = this.$selection.find('.select2-selection__rendered').innerWidth() || '100%'`

I'm not sure it's the preferred solution but it does the job and fixes the issue.

kevin-brown:
While there is a lot of good information here, would it be possible for you to create a jsbin/jsfiddle where we can reproduce this?

brandonhall:
I don't know if it can be replicated in jsbin/jsfiddle since it requires a template language, dynamic DOM insertion. Without the element in the DOM, it would be tricky to reproduce in one of those tools.

donald-slagle:
I'm running into the same issue. I'm using Aureila and created a custom control that uses select2. The custom control initializes select2 before it is rendered and the placeholder width is set to 0px. If I inspect it in chrome and remove the 0px width it displays correctly.

wbercx:
Same issue here, also in a Backbone app where Select2 gets instantiated on render, rather than when actually shown in the DOM, so innerWidth() fails.

What I ended up doing was `width = this.$selection.find('.select2-selection__rendered').innerWidth() || '100%'`, as suggested by @brandonhall, but that still resulted in the input field not being wide enough to display the entire placeholder in some cases. This seems to be the case because the inline search is floated to the left.

An extra line of CSS to stop `float: left` from being applied when there are no selected items did the trick there: 
    `.select2-container .select2-search--inline:only-child { float: none; }`

Joyrex:
I'd like to add that this fix works exactly like @wbercx mentions - the additional CSS keeps longer than usual placeholders from being cut off when using the JS fix proposed by @brandonhall.

Is there any reason this still isn't merged into the current release?

PR #4898, opened 2017-05-08
Fix: placeholder width cutting off text #291

### LINKED ISSUE #291: Placeholder width cutting off text when Select2 is initialized inside a hidden element
We're having to patch select2 like so to get our placeholder text to appear properly.
- ln 345: width: auto !important;

This happens when the new width option is either not provided or is set to 'element'. The placeholder gets an inline css width set, which mysteriously goes away and never comes back when the input is interacted with.

bews:
This pull request includes a

- [x] Bug fix
- [ ] New feature
- [ ] Translation

Fixes #291, #3669 and other duplicates.

### DIFF
--- a/src/js/select2/selection/search.js
+++ b/src/js/select2/selection/search.js
@@ -206,16 +206,19 @@ define([
     this.$search.css('width', '25px');
 
     var width = '';
+    var widthParent = '';
 
     if (this.$search.attr('placeholder') !== '') {
-      width = this.$selection.find('.select2-selection__rendered').innerWidth();
+      width = '100%';
+      widthParent = '100%';
     } else {
       var minimumWidth = this.$search.val().length + 1;
-
       width = (minimumWidth * 0.75) + 'em';
+      widthParent = 'auto';
     }
 
     this.$search.css('width', width);
+    this.$search.closest('.select2-search--inline').css('width', widthParent);
   };
 
   return Search;


jpic:
The width change makes sense but is there any risk that the closest call fails to pick the right element ?

If both this.$search and its .select2-search--inline are children of the widget's parent element then it would be safe.

I wonder what @kevin-brown meant in the comment that there is no clean fix here  https://github.com/select2/select2/issues/3669#issuecomment-130096418

bews:
I didn't look deep into the code, but I believe `.select2-search--inline` is always there. And it's the direct parent of `this.$search`.

src/js/select2/selection/search.js
```diff
@@ -206,16 +206,19 @@ define([
...
+      widthParent = '100%';
     } else {
       var minimumWidth = this.$search.val().length + 1;
-
       width = (minimumWidth * 0.75) + 'em';
+      widthParent = 'auto';
     }
 
     this.$search.css('width', width);
+    this.$search.closest('.select2-search--inline').css('width', widthParent);
```
bews: This can be replaced by `.parent()` to be sure.

alexweissman: Do you still want to make that change?

cssagogo:
When is this fix going out?

src/js/select2/selection/search.js
bews: It works fine the way it is, but I can change it if you think `.parent()` is a better approach. I'm ok with either.

alexweissman:
It just seems suspicious to me that replacing a calculated value like `width = this.$selection.find('.select2-selection__rendered').innerWidth();` with a hardcoded value like `100%` would be a fix.

Surely there must be a reason that it wasn't set to 100% in the first place?

By the way, a more conservative patch is suggested in #3292:

`width = this.$selection.find('.select2-selection__rendered').innerWidth() || '100%';`

bews:
> `width = this.$selection.find('.select2-selection__rendered').innerWidth() || '100%';`

This doesn't work, since innerWidth returns incorrect value, not 0. It may be bigger than it should, or a fixed 100px.

I dunno why it returns that and don't have time to figure out.

ala-musleh:
this css works perfect for me (multiple selection, select starts hidden):

.select2-container .select2-selection__rendered > *:first-child.select2-search--inline {
width: 100% !important;
}
.select2-container .select2-selection__rendered > *:first-child.select2-search--inline .select2-search__field {
width: 100% !important;
}

vpn:
Thanks for that, @ala-musleh.

spo0okie:
this problem stil exists on visible elements (not only for hidden ones from closed issues) for multiple select, and that fix works perfect. why it sitll not merged?

seaexplorer2024:
just commenting I still get the same issue on V4.0.13 for multi selects.

Using @xander-mu's styling has fixed it for me

DonSYS91:
Issue still exists as of now as well on latest version.
Thank you so much @xander-mu  for the workaround

PR #4908, opened 2017-05-18
Improving .select2-hidden-accessible

ffoodd:
To match last standards, see [CSS hide and seek](http://hugogiraudel.com/2016/10/13/css-hide-and-seek/) for more explanations :)

Also I removed the negative margin because of unwanted effects in some case, as seen on Bootstrap.


alexweissman:

Issue #4725, opened 2016-12-20
Events are not detached properly, causing memory leaks


muthuraja:
Hi,
Events not properly detached. Its eating lot of memory. I need the solution.   
![select2](https://cloud.githubusercontent.com/assets/26075/21353500/4db53982-c6ec-11e6-964e-966d650dd09b.png)

Muthuraja

kevin-brown:
What event? And when is it supposed to be removed?

derekm:
User also reported their error here: https://github.com/iStefo/ember-select-2/issues/143

Dshoe:
We can reproduce listener leaks in an Angular 1.x app. What is proving difficult is root causing the issue. Any suggestions?

pedrofurtado:
We don't have immediate plans to provide this. We are focused to fix some major UI bugs (that are majority of issues and PR's). But if you open a PR with unit tests, I will be glad to review and approve if everything is ok :+1:

Issue #4236, opened 2016-03-07
Bug Scroll search ajax in bootstrap modal


mahadoang:
## Steps to reproduce the issue
1. add select2 (search ajax) in bootstrap modal
2. if modal height over window browser  have scroll y
3. click select2 in modal select2 show box for search 
4. then select2 closed the box for search,modal cannot use scroll
## Environment

Browsers
- [ ] Google Chrome
- [ ] Mozilla Firefox
- [ ] Internet Explorer

Operating System
- [ ] Mac OS X

Libraries
- jQuery version: 2.1.4
- Select2 version: 4.0.1

mahadoang:
I Fixed By change Line Number : 4142 

$(this).scrollTop(position.y);
TO
$(self).scrollTop(position.y);

kalilz4485:
We had the issue too and used @mahadoang 's fix.

It would be nice if this was merged.

kevin-brown:
This was a duplicate of https://github.com/select2/select2/issues/3125, it should have been fixed in 4.0.2.

kalilz4485:
I still have the issue after updating to 4.0.2.

Can't scroll after changing option in a select (note that in multi selects I can re-scroll if I delete an option, but adding one re-block the scrolling)

mahadoang:
i fixed issue 

Before  fixed

    $watchers.on(scrollEvent, function (ev) {
        var position = $(this).data('select2-scroll-position');
        $(this).scrollTop(position.y);
    });


After fixed
    $watchers.on(scrollEvent, function (ev) {
        var position = $(this).data('select2-scroll-position');
        if($(this).find('.select2-container--open').size()>0)
        {
            $(this).scrollTop(position.y);
        }
    });

working is ok.

kalilz4485:
@mahadoang Have you tried to upgrade to 4.0.2 ? If so did it fix the bug without having you to patch it yourself ?

devshangari:
I tried the same with the upgrade to 4.0.2 and was still having the same issue, the fix by @mahadoang works!

luisDanielRoviraContreras:
buenas amigos el aporte de @mahadoang  esta exelente pero para que funcione en el ultimo jquery en ves de :
`$watchers.on(scrollEvent, function (ev) {
    var position = $(this).data('select2-scroll-position');
    if($(this).find('.select2-container--open').size()>0)
    {
        $(this).scrollTop(position.y);
    }
});`
es 
[...]

PR #4942, opened 2017-06-07
Fix cleanup of scroll positioning handler

UCIS:
If the Select2 DOM elements are removed from the DOM or moved to another parent element while the select2 dropdown is open, the scroll event would not get detached from the correct elements. This breaks scrolling and causes a memory leak. This patch fixes the problem by storing the elements to which event handlers were bound.

This pull request includes a

- [x] Bug fix

### DIFF
--- a/src/js/select2/dropdown/attachBody.js
+++ b/src/js/select2/dropdown/attachBody.js
@@ -87,9 +87,12 @@ define([
     var scrollEvent = 'scroll.select2.' + container.id;
     var resizeEvent = 'resize.select2.' + container.id;
     var orientationEvent = 'orientationchange.select2.' + container.id;
+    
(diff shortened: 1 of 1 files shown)


alexweissman:
@UCIS is #4236 a manifestation of this bug?  Do you have another issue that references this fix?

UCIS:
@alexweissman it doesn't look directly related. In our case we were removing the DOM element containing the select2 in response to a change event fired by select2, which happened before select2 cleaned up its dropdown and scroll handler. I don't think the select2 container is removed from the DOM in #4236.

alexweissman:
Ok, is there a failing test you can implement, which this patch fixes?

alexweissman:
Also, is #4725 a manifestation of this?

PR #4958, opened 2017-06-16
fix($tag-input): Fix tag input on IE11

jeznag:
#4398 #4570

This pull request includes a

- [x] Bug fix
- [ ] New feature
- [ ] Translation

- Fix bug with tag input in IE11 and Chrome


alexweissman:
@jeznag can you please comment on #4860 so we can implement this fix?  Thanks!

alexweissman:
Actually, this is equivalent to #4860, but #4860 is slightly cleaner.  So, I'm gonna close this one.  But, thanks for contributing!

PR #4966, opened 2017-06-21
add enableMousewheel property

silver0r:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

- enableMousewheel means to use $.fn.mousewheel.
- If you don't use $.fn.mousewheel, you can not use mouse scroll in select2 body.

### DIFF
--- a/dist/js/select2.full.js
+++ b/dist/js/select2.full.js
@@ -1177,7 +1177,7 @@ S2.define('select2/results',[
       self.displayMessage(params);
     });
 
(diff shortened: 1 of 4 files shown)


kevin-brown:
Just following up with this pull request, was there a case where you had jQuery Mousewheel installed but did not want Select2 to prevent the page from scrolling when people over-scrolled on the page?

kevin-brown:

PR #5019, opened 2017-08-25
Add style property to package.json

soundasleep:
By adding a reference to the main dist CSS, `gulp` tools such as [`sass-module-importer`](https://www.npmjs.com/package/sass-module-importer) can automatically import the stylesheet.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,11 +30,11 @@
   ],
   "license": "MIT",
   "main": "dist/js/select2.js",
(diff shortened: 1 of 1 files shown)


alexweissman:
This seems like a pretty safe change to merge in.  `style` is not an [official property of `package.json`](https://docs.npmjs.com/files/package.json), but it seems like it is on its way to becoming so.

alexweissman:

PR #5033, opened 2017-09-07
Replace docs with a redirect

claudiulodro:
I've deleted all of the files in the docs folder except README.md and index.html

README.md has updated info about this.
index.html just redirects to https://select2.org.

The main repo readme will need to be updated, but that can be a separate PR when everything is ready.

cc @alexweissman


alexweissman:
Ok.  Trying to get the new site's forever home set up first.

alexweissman:
We should actually set up the other pages to redirect as well:

- https://select2.github.io/examples.html
- https://select2.github.io/options.html
- https://select2.github.io/announcements-4.0.html
- https://select2.github.io/community.html
- https://select2.github.io/options-old.html

PR #5041, opened 2017-09-16
fix issue with select2 input focus on tag deletion

mayuri-aquis:
This pull request includes a

- Bug fix

The focus from select2 input was lost to body tag on deleting an option from multiple select list due to which the browser page was navigated to previous page on further backspace key hits. 

The changes were made to keep the focus on select list on deletion of options from multiple select list.


alexweissman:
Nice!  A couple things:

- Could you implement the fix directly in the source file, rather than the compiled `dist/` files?
- Could you add a test that fails without the fix, and passes with the fix?

kevin-brown:

Issue #5045, opened 2017-09-19
AllowClear event


killbox:
Good day!
How can I tell the allowClear event?
Thanks!

alexweissman:
It is not clear what you are asking.  Documentation for the `allowClear` event is available here: https://select2.org/selections#clearable-selections, and for events is available here: https://select2.org/programmatic-control/events

alexweissman:
The `clear` event is now available as of v4.0.6-rc.0.

Issue #3773, opened 2015-09-16
allowClear with multiple="multiple" triggers a 'change' on every item cleared.


vans163:
If you clear a multiple select with say 5 items. change event will trigger 6 times.  1 for each item cleared and once for finalization. This is on firefox.

The problem is line 1714: this.trigger('unselect', unselectData);   apparently unselect triggers the change event.

Possible fixes are comment out this line as it looks useless.  
Another fix is to trigger a custom event called unselect-select2 for example to not trigger the change event.
Another fix is to rewrite change event to be called change-select2 and bind that, but this needs to be propogated throughtout the code.

apbeilchick:
Same thing is happening in IE11.  Also, after invoking the "clear all", the dropdown list appears without adding the select-container--focus class

gracedigital:
Using select2 4.0.3 from CDN.  I've created a multi-select via HTML and initialized it via javascript where multiple options are selected. However if I clear one of the selected options they all clear.  Is there a fix or work around for this?

alexweissman:
This issue is solved as of 4.0.6-rc.0 (see https://github.com/select2/select2/pull/5058#issuecomment-334550960).

al:
Apologies for commenting on such an old issue but I'm seeing the above behaviour on 4.0.13 (select2.full.js), i.e. multiple change events being fired in response to clearing a multiple select.

As this has been marked as resolved I just wanted to clarify whether or not it is considered a bug?

In my scenario I'd just like to submit a form on change and would like to keep that logic generic rather than have to worry about whether the event is coming from Select2 or otherwise have to intercept and handle Select2 specific events.

```js
$form.on('change', 'select, input', (event) => {...})
```

Does it deserve a new issue or have I missed something maybe?

Narghor:
Hello

I have the same problems with the latest release.
Change event is firing for each item cleared ... 

How can we prevent this ?

Mckill3r:
I have the same issue with 4.1.0-rc.0

aanish27:
"select2": "4.1.0-rc.0",
"select2-bootstrap-5-theme": "1.3.0"

I had the same issue and found a workaround by calling the built-in event select:clear which is triggered after clearing along with change ,

https://select2.org/programmatic-control/events
clear event
$('.select2-select').on('select2:clear', function () {
   // data retrieval
})

change event - add this as the first line of code because it still triggers the 'change' on every item cleared 
if($(this).val().length == 0) return;

Issue #3772, opened 2015-09-16
When I clear one selection on a multi-select with allowClear set to true, all selections are cleared


Intrepidd:
Here's an example fiddle :

http://jsfiddle.net/Intrepidd/ytL9s97k/

Try to add several results to the multiple select then remove only one, it will remove everything, and also the clear button appears.

Here's a gif demonstrating the issue : 

![clear_multiple](https://cloud.githubusercontent.com/assets/803765/9909877/98da21ce-5c9a-11e5-87fe-a66137ddc2fc.gif)

I understand you should not use `allowClear` with a multiple select but in some cases you just want to apply select2 to all selects.

A quick fix would be to call select2 two times with the following selectors :

```
$('select[multiple]').select2({width: '100%'});
$('select:not([multiple])').select2({allowClear: true, width: '100%'});
```

but this is messy.

IMHO `allowClear` should be ignored if the select is multiple

thoughts ?

kevin-brown:
This is definitely a duplicate, but part of the issue is that you are missing `placeholder`, which is required for `allowClear`. I think we trigger a warning about that when you have `debug: true` set.

Until I hunt down the ticket, this one will stay open.

SlyDave:
Whilst it may be the current requirement, I don't get why "placeholder" is a requirement for this behaviour, they do not seem related at all.

kevin-brown:
It's a requirement for single selects, and `allowClear` was designed with a single select in mind. For a single select, we need to reset it to a specific value instead of just clearing all selections.

In theory, [this line](https://github.com/select2/select2/blob/8ad8f200ba8c9429d636453b8ee3bcf593e8c87a/src/js/select2/selection/allowClear.js#L63) could be conditional (default to `null` if `this.placeholder` isn't there) which would remove the requirement for multiple selects. But then there would be a configuration difference between single and multiple selects, which is something we are trying to avoid.

alexweissman:
This is a duplicate of #3335.  Since @kevin-brown offers some useful insight, I will repost his comments in #3335.

PR #5058, opened 2017-09-26
Event improvements

### LINKED ISSUE #4929: Event on clear
When setting multiple to true `'multiple' => true` on a select2 widget allowClear is also set to true `'allowClear' => true`, there are two ways to deselect options: 

1. `.select2-selection__clear` which clears all options 
2. `.select2-selection__choice__remove` deselects a single choice

How can I tell the difference between the two? Currently they both trigger the `unselecting` and `unselect` events.

### LINKED ISSUE #5049: select2:unselect fires before changing value in single-select dropdowns
1. Create a single-select dropdown with `allowClear`
2. Subscribe to `select2:unselect` event
3. Select some option
4. Click the clear button (_X_)
5. `$(dropdown).val()` still returns the value in event handler
6. Now do the same with multi-select. When `select2:unselect` fires, the value is already removed.

This should be consistent for single-selects and multi-selects. Currently for single-selects `select2:unselect` event is fired before removing the value, but for multi-select value is already removed when the event fires.

## Environment

Browsers

- [x] Google Chrome
- [x] Mozilla Firefox
- [?] Internet Explorer

Operating System

- [x] Windows
- [?] Mac OS X
- [?] Linux
- [?] Mobile

Libraries

- jQuery version: 3.2.1
- Select2 version: 4.0.3, master (#c454407)

## Isolating the problem

- [ ] This bug happens [on the examples page](https://select2.github.io/examples.html)
- [x] This bug happens when using Select2 without other plugins
- [x] I can reproduce this bug in [a gistrun](https://gist.run/?id=033b9625607481e21d39e535c6487dab)

gronostajo:
This pull request includes a

- [x] Bug fix
- [x] New feature
- [ ] Translation

- When user clears the select, value is now cleared before `unselect`  event is emitted (fixes #5049). If the event is cancelled, the original value will be restored.
- A `clear` event was added which is emitted only when clearing and unlike `unselect`, it's always emitted only once for multiple select (instead of once for each cleared item). This event is also cancellable, and if so, no `unselect` events are emitted. (closes #4929 and probably #5045)

### DIFF
--- a/src/js/select2/core.js
+++ b/src/js/select2/core.js
@@ -417,7 +417,8 @@ define([
       'open': 'opening',
       'close': 'closing',
       'select': 'selecting',
-      'unselect': 'unselecting'
+      'unselect': 'unselecting',
+      'clear': 'clearing'
     };
 
     if (args === undefined) {
--- a/src/js/select2/selection/allowClear.js
+++ b/src/js/select2/selection/allowClear.js
@@ -48,8 +48,17 @@ define([
     var previousVal = this.$element.val();
     this.$element.val(this.placeholder.id);
 
+    var unselectData = {
+      data: data
+    };
+    this.trigger('clear', unselectData);
+    if (unselectData.prevented) {
+      this.$element.val(previousVal);
+      return;
+    }
+
     for (var d = 0; d < data.length; d++) {
-      var unselectData = {
+      unselectData = {
         data: data[d]
       };
 
--- a/src/js/select2/selection/allowClear.js
+++ b/src/js/select2/selection/allowClear.js
@@ -45,6 +45,9 @@ define([
 
     var data = $clear.data('data');
 
+    var previousVal = this.$element.val();
+    this.$element.val(this.placeholder.id);
+
     for (var d = 0; d < data.length; d++) {
       var unselectData = {
         data: data[d]
@@ -56,11 +59,12 @@ define([
 
       // If the event was prevented, don't clear it out.
       if (unselectData.prevented) {
+        this.$element.val(previousVal);
         return;
       }
     }
 
-    this.$element.val(this.placeholder.id).trigger('change');
+    this.$element.trigger('change');
 
     this.trigger('toggle', {});
   };
--- a/src/js/select2/selection/eventRelay.js
+++ b/src/js/select2/selection/eventRelay.js
@@ -9,10 +9,13 @@ define([
       'open', 'opening',
       'close', 'closing',
       'select', 'selecting',
-      'unselect', 'unselecting'
+      'unselect', 'unselecting',
+      'clear', 'clearing'
     ];
 
-    var preventableEvents = ['opening', 'closing', 'selecting', 'unselecting'];
+    var preventableEvents = [
+      'opening', 'closing', 'selecting', 'unselecting', 'clearing'
+    ];
 
     decorated.call(this, container, $container);
 
--- a/tests/selection/allowClear-tests.js
+++ b/tests/selection/allowClear-tests.js
@@ -190,6 +190,94 @@ test('preventing the unselect event cancels the clearing', function (assert) {
   );
 });
 
+test('clicking clear will trigger the clear event', function (assert) {
+  assert.expect(5);
+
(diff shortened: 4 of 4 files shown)


alexweissman:
This looks good, and I appreciate that you added some tests as well!

Could you add a few comments to explain what your changes in `allowClear.js` are doing?

gronostajo:
Sure. If you need some comments in code, I can add them as well.

### The Problem

Let's say you have a single select with `allowClear: true`. Clicking the clear button emitted `select2:unselect` first, then cleared the value, so checking for selected value in event handler returned something. This was especially problematic when building a generic single-or-multi-select component, because removing single items from multiple select uses the same `select2:unselect` event, but it worked the opposite way: item was removed first, then the event fired. This basically meant that you couldn't rely on select's value in `unselect` handlers.

### The Fix, part 1

I've described that issue in #5049. This PR fixes it. The value before clearing is stored in a temporary variable and actual value is cleared. Then `select2:unselecting` and `select2:unselect` is emitted for each removed item (this part is left unchanged). Hovewer, if any of `select2:unselecting` events is cancelled, the value is restored from temporary variable and no more `unselect` events are emitted. If none of `select2:unselecting` events is cancelled, a `change` event is sent to select to update select2's DOM (this is where value was updated previously).

### The Fix, part 2

In addition, I've added another event, `select2:clear`, and accompanying `select2:clearing` (related issue #4929). These differ from `select2:unselect`:

- `clear` events are emitted only when clearing the select, not when removing single items from multiple select
- `clear` events are always sent only once, no matter how many items are cleared
- `clear` event happens before `unselect`s
- if `select2:clearing` is prevented, `select2:clear` and `unselect` event's won't be emitted at all

alexweissman:
Ok, I think I'm starting to understand.  Basically, this makes the behavior of "clear selection in single-select" and "remove selection in multi-select" consistent.  In that with your patch, in both cases the actual value of the control (as accessed via `.val`) is unset _before_ the `clear` and `unselect` events are fired, right?

The data passed to these events though, is the same, right?  It's the entire [data properties](https://select2.org/programmatic-control/events#event-data) object of the selection that was cleared/unselected?

alexweissman:
Also, what happens with `allowClear` in the multiselect case?

gronostajo:
Exactly, `unselect` is now consistent in both cases and `clear` works the same way. `unselect` data properties haven't changed and `clear` has an array of all cleared items. (`unselect` events contain single items)

To be honest, I haven't tested `allowClear` in multiselects, but I suppose it also presented the inconsistent behavior. There's just one piece of code that handles `allowClear` in both cases and it's actually written with multiselects in mind (single select is a special case of multiselect with single value). So there's no reason to suppose it works any different from single selects.

alexweissman:
Cool!  I only ask because I know that there are some issues about `allowClear` and multi-select.  See https://github.com/select2/select2/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aopen%20allowclear%20multi, in particular #3773 (which I think your `clear` event solves) and #3772.

Can you test this?  This might be a good opportunity to deal with all of these `allowClear` issues in one blow 💪

hoanghuan0701:
For the issue with [#3772](https://github.com/select2/select2/issues/3772). I've apply the new code changes of @gronostajo, but it does not work as expected. When you press backspace in multi selection, all selected items will be removed.

alexweissman:
@hoanghuan0701 thanks!  But from what I can tell of #3772, this was a problem even without @gronostajo's patch?  Can you confirm whether this patch is actually introducing a new bug?

gronostajo:
@alexweissman It seems that #3773 is resolved.

#3772 is present in [my gist](https://gist.run/?id=033b9625607481e21d39e535c6487dab) which demonstrates the problem I attempted to fix. Just add `allowClear: true` in the second select and it will start happening. And my patches indeed don't fix it as @hoanghuan0701 noted. Unfortunately I'm too busy to fix it anytime soon.

alexweissman:
That's fine, I just want to make sure that we're not introducing any _new_ bugs.

alexweissman:

Issue #4763, opened 2017-01-25
mousedown/mouseup causes problems in making selections on mobile devices


mkmayer:
## Prerequisites

- [x] I have searched for similar issues in both open and closed tickets and cannot find a duplicate
- [x] The issue still exists against the latest `master` branch of Select2
- [x] This is not a usage question (Those should be directed to the [community](https://select2.github.io/community.html))
- [x] I have attempted to find the simplest possible steps to reproduce the issue
- [ ] I have included a failing test as a pull request (Optional)

## Steps to reproduce the issue

1. Stylize "select2-container--open" to stretch to whole page (mobile phone screen)
2. Click on select2 dropdownlist (mousedown). Results are displayed
3. Release mouse button (mouseup). Value is selected

## Expected behavior and actual behavior

When I follow those steps, the value is selected on one mouse click. It's because mousedown on dropdownlist shows results. It show results over whole page (which is wanted on mobile phones). On releasing mouse, it selects the value. But on mobile phone, are these events fired at once. So user can't choose from results.
If I don't use plugin closeOnSelect, results are shown, but the bug is still there (value is selected on touch)

I was expecting, that opening of results is bounded to mouse click, not to mouse down.
And selecting of results should be bounded to mouse click too.

This prevent using of select2 on mobile phones (after stylizing results to cover the select2 element).
Please could you **change the events from mouse down and mouse up to mouse click**?
Then it is usable on mobile phones too. It's ease to stylize results to whole page, but select2 is not usable then.

## Environment

Browsers

- [x] Google Chrome
- [x] Mozilla Firefox
- [x] Internet Explorer

Operating System

- [ ] Windows
- [ ] Mac OS X
- [ ] Linux
- [x] Mobile

Libraries

- jQuery version: 1.11.3
- Select2 version: 4.0.3

## Isolating the problem

- [x] This bug happens [on the examples page](https://select2.github.io/examples.html)
- [x] The bug happens consistently across all tested browsers
- [x] This bug happens when using Select2 without other pluigns
- [x] I can reproduce this bug in [a jsbin](http://jsbin.com/vacecikane/edit?html,css,js,output)

kevinob11:
I love the idea of filling the screen on mobile, but I'm experiencing the same issue you are. I click to open the select and it immediately clicks the response located below the touch point.

sleepydada:
@kevinob11 same here

alexweissman:
It appears that PR #4536 addresses this problem.

[...]

PR #4536, opened 2016-08-17
Close the dropdown when user tap the background on iOS safari browser

letrunghieu:
This pull request includes a
- [x] Bug fix
- [ ] New feature
- [ ] Translation

The following changes were made
- Listen to the `touchstart` event from outside to close the dropdown
- Stop propagating the `touchstart` event from the dropdown container

Related issues: #4087, #3577

Stackoverflow question: http://stackoverflow.com/q/31396258/1399505

> What I'm seeing on a test site is that on all browsers except iOS, when you click outside of the dropdown area, the dropdown closes (which is the desired behavior). On iOS, the only way to get it to close is to click inside the input.

With this PR, the dropdown is close when users click on its outside area.

Issue #4482, opened 2016-07-14
Doesn't work in iOS UIWebView


adam-lynch:
## Prerequisites
- [x] I have searched for similar issues in both open and closed tickets and cannot find a duplicate
- [x] The issue still exists against the latest `master` branch of Select2
- [x] This is not a usage question (Those should be directed to the [community](https://select2.github.io/community.html))
- [x] I have attempted to find the simplest possible steos to reproduce the issue
- [ ] I have included a failing test as a pull request (Optional)
## Steps to reproduce the issue
1. Use basic example in an iOS WebView
2. Tap on it
3. Nothing happens
## Expected behavior and actual behavior

Expected it to open but it doesn't.
## Environment

Browsers
- [ ] Google Chrome
- [ ] Mozilla Firefox
- [ ] Internet Explorer
- [x] UIWebView (Safari kinda)

Operating System
- [ ] Windows
- [ ] Mac OS X
- [ ] Linux
- [ ] Mobile
- [x] iOS

Libraries
- jQuery version: 2.1.4
- Select2 version: 4.0.3
## Isolating the problem
- [x] This bug happens [on the examples page](https://select2.github.io/examples.html)
- [ ] The bug happens consistently across all tested browsers
- [x] This bug happens when using Select2 without other pluigns
- [x] I can reproduce this bug in [a jsbin](https://jsbin.com/)

---

Works fine everywhere (including iOS Safari) but the `mousedown` event doesn't seem to get fired in the WebView at all. My workaround is to duck punch `$.fn.on` and `$.fn.off` mapping `mousedown` to `click`;

``` coffeescript
        if appIsCurrrentlyRunningInOurMobileApp
                convertMousedownEvent = (originalMethod) ->
                    return (event, callback) ->
                        if typeof event is 'string' and event.indexOf('mousedown') is 0 and
                        (event.indexOf('select2') isnt -1 or this.get(0)?.className?.indexOf('select2') isnt -1)
                            args = Array.prototype.slice.call arguments
                            args[0] = event.replace 'mousedown', 'click'
                            args[1] = (e) ->
                                e.which = 1 # pretend it's a left-click because they check in some places
                                return callback.call this, e
                        else
                            args = arguments

                        return originalMethod.apply this, args

                $.fn.on = convertMousedownEvent $.fn.on
                $.fn.off = convertMousedownEvent $.fn.off
```

[...]

PR #5069, opened 2017-10-05
Fix https://github.com/select2/select2/issues/4276

UksusoFF:
This pull request includes a

- [x] Bug fix
- [ ] New feature
- [ ] Translation

- Change mousedown to click event. Fix for https://github.com/select2/select2/issues/4276

### DIFF
--- a/dist/js/select2.full.js
+++ b/dist/js/select2.full.js
@@ -1521,7 +1521,7 @@ S2.define('select2/selection/single',[
     this.$selection.find('.select2-selection__rendered').attr('id', id);
     this.$selection.attr('aria-labelledby', id);
 
-    this.$selection.on('click', function (evt) {
+    this.$selection.on('click.select2Single', function (evt) {
       // Only respond to left clicks
       if (evt.which !== 1) {
         return;
--- a/dist/js/select2.full.js
+++ b/dist/js/select2.full.js
@@ -1612,7 +1612,7 @@ S2.define('select2/selection/multiple',[
 
     MultipleSelection.__super__.bind.apply(this, arguments);
 
-    this.$selection.on('click', function (evt) {
+    this.$selection.on('click.select2Toggle', function (evt) {
       self.trigger('toggle', {
         originalEvent: evt
       });
--- a/dist/js/select2.full.js
+++ b/dist/js/select2.full.js
@@ -1619,7 +1619,7 @@ S2.define('select2/selection/multiple',[
     });
 
     this.$selection.on(
-      'click',
+      'click.select2MultipleRemove',
       '.select2-selection__choice__remove',
       function (evt) {
         // Ignore the event if it is disabled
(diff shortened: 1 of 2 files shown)


alexweissman:
There are some other issues related to `mousedown` vs `click` events:

- #4763 (addressed in #4536)
- #4642 (addressed in #4644)
- https://github.com/select2/select2/issues/4482#issuecomment-264916443
- https://github.com/select2/select2/issues/4276#issuecomment-212663921

Actually #4276 makes a good point:

> Simply changing the opening event from `mousedown` to `click` fixes it, at the cost of losing functionality parity with a base `<select>` element.
A small (humanly imperceptible) delay between an opening `mousedown` and a selecting `mouseup` may be another way around it (ie, within the range of a normal 'click'), but again is not all that ideal.
Perhaps just an option on which event to listen for...?

Could you review these please?  It sounds like `mousedown` was originally used instead of `click` because of the behavior of mobile browsers, but this behavior has since been removed from most mobile browsers.  

I just want to make sure we don't introduce any new regressions.  It sounds like maybe a configuration setting for this might still be worthwhile, even if we change the default to be `click`?

UksusoFF:
@alexweissman 
Checked this fix for a long time in Chrome/Safari on Mac OS X. 
Checked in Android Firefox/Chrome, iOS Safari/Chrome, Windows Chrome/Firefox/IE just now.

Looks like all of this issues have same trouble because dropdown overlap selection container:
https://github.com/select2/select2/issues/4763
https://github.com/select2/select2/issues/4642
https://github.com/select2/select2/issues/4276

I can't check this issue:
https://github.com/select2/select2/issues/4482
But if `mousedown` really not fired in iOS WebView (found many questions about it on StackOverflow [e.g.](https://stackoverflow.com/questions/26885670/mouseup-and-mousedown-not-working-on-ios)) PR must be help.

I don't think that need to configure that because we can bind `$('#mySelect2').select2('open');` on any actions.

alexweissman:
Ok so as far as #4644 goes, do you think we should leave it as `click` in the multi-select case?

I'm actually not quite sure how this PR addresses #4200 - that issue seems to be related to `mouseup`, not `mousedown`.  Could you explain further?

UksusoFF:
> I'm actually not quite sure how this PR addresses #4200 - that issue seems to be related to mouseup, not mousedown. Could you explain further?
My bad. I mentioned wrong issue.

> Ok so as far as #4644 goes, do you think we should leave it as click in the multi-select case?
What must occur when we click in Alaska in provided example? https://jsfiddle.net/vwsg4tn5/ 
We can delete item by cross. And we can show dropdown by clicking on Alaska. I don't think this behavior is wrong.

alexweissman:
So, #4642 and #4295 describe some potential issues with using `click` for multi-select.

alexweissman:
@UksusoFF it seems like this is not a good fix in the long term.  See my comment at https://github.com/select2/select2/issues/4276#issuecomment-334624014.

UksusoFF:
Okay let's continue in issue.

PR #5079, opened 2017-10-10
Add jsDelivr hits badge

LukasDrgon:
Hey, I noticed that select2 gets 36M hits per month on jsDelivr, so I thought you might like this badge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
 Select2
 =======
 [![Build Status][travis-ci-image]][travis-ci-status]
+[![jsDelivr Hits][jsdelivr-badge]][jsdelivr-hits]
 
 Select2 is a jQuery-based replacement for select boxes. It supports searching,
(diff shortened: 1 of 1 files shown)


alexweissman:
Thanks!  However I'm not sure how useful it is to display CDN hits in the repo, as this tells us more about the sites using Select2 rather than the project itself.  There are so many other CDNs that also host Select2, so this would also be a distorted picture either way.

Not sure if you work for jsDelivr, but it would be more useful if you could approximate the number of _sites_ using the CDN, perhaps by checking the Referer header or something.

MartinKolarik:
@alexweissman 

>it would be more useful if you could approximate the number of sites using the CDN, perhaps by checking the Referer header or something

That's an interesting idea. Considerably harder to build, but I added it to our list of stats we might add later.

Issue #4914, opened 2017-05-22
How to add Delay to keydown event ?


ghost:
Select2 is slow when typing keywords if the select has a fair number of items.
Maybe everytime keydown event runs searching.
I want to insert like ```setTimeout```

alexweissman:
For large numbers of items, you may want to consider using a remote data source with [server-side pagination](https://select2.org/data-sources/ajax#pagination).

KasemOmary:
True that pagination is a good idea - but the number of people using a rock or a potato as a computer is ... fairly decent.  Callbacks to the server will not do the UX for said users much help ;  they will still experience input/update delays on their end.  Working on the client end is, perhaps, a better solution for said user.  In my case, this is a couple dozen entries.  Not exactly a book.

alexweissman:
I've seen some other issues about [performance in v4](https://github.com/select2/select2/labels/performance).  At the time I just assumed that these were limited to cases where the developer made a bad decision to render thousands of entries at once.  But now that you say:

> In my case, this is a couple dozen entries causing UX sluggishness.

I am more concerned about the issue.  It might be worth digging into the underlying causes of the performance problems in Select2 v4, rather than using a keydown delay as a workaround.

> I am forking the project to create a version which supports timeouts. Consider the suggestion for inclusion within the primary project.

I believe the term for this is "pull request" 😉

KasemOmary:
"Technically" I did do a fork, as I need this working now.  I'll get a pull request for the primary release in as soon as I'm done with testing my modification out.  So far, it's looking nice.

KasemOmary:
https://github.com/select2/select2/pull/5082

KasemOmary:
I have a working addition to address this made - however I'm having some difficulty grabbing options. 

![image](https://user-images.githubusercontent.com/3031521/31676540-f7186b4e-b335-11e7-80c8-a8c973dba922.png)

    this.$search.on('keyup input', function (evt, options) {
        var timeout = options;
        timeout = 1000; // <= Need this. 
        if (timeout > 0) {
            if (this.SearchTimeout != null) {
                clearTimeout(this.SearchTimeout);
            }

            var param = { self: self, evt: evt };
            var s2Timeout = function () {
                this.self.handleSearch(this.evt);
            };
            this.SearchTimeout = setTimeout(s2Timeout.bind(param), timeout);
        } else {
            self.handleSearch(evt);
        }
    });

[...]

PR #5082, opened 2017-10-17
Add timeout support and "searchTimeout" option.

KasemOmary:
This pull request includes a

- [ ] Bug fix
- [X] New feature
- [ ] Translation

- Added the keyword "searchTimeout"
- Added functionality to allow for a msec timeout to be set before <select> options are searched.

https://github.com/select2/select2/issues/4914

### DIFF
--- a/src/js/select2/dropdown/search.js
+++ b/src/js/select2/dropdown/search.js
@@ -42,8 +42,21 @@ define([
       $(this).off('keyup');
     });
 
-    this.$search.on('keyup input', function (evt) {
-      self.handleSearch(evt);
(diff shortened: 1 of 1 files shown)


KasemOmary:
* Fixing the issues and will setup a new pull request.

alexweissman:
It would seem that jshint doesn't like `setTimeout` and `clearTimeout`: https://travis-ci.org/select2/select2/builds/289120165#L680-L684

KasemOmary:
Locally I have the issue fixed.   Keeping this request open for reference while I work on a better solution.

PR #5089, opened 2017-10-24
Update select2.js

kutec:
Adding ending statement at the start of the code as absence of that causes console error while concatenating files.

This pull request includes a

- Bug fix

- Added semi column at the starting of the code to avoid concatenation issue with Grunt

If this is related to an existing ticket, include a link to it as well. - NO

### DIFF
--- a/dist/js/select2.js
+++ b/dist/js/select2.js
@@ -5,7 +5,7 @@
  * Released under the MIT license
  * https://github.com/select2/select2/blob/master/LICENSE.md
  */
-(function (factory) {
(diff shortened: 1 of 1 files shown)


alexweissman:
Just FYI, this change needs to be made in `src/`, not in `dist/`.  Please review https://github.com/select2/select2/blob/master/.github/CONTRIBUTING.md#contributing-to-select2 to see how Select2 is built - thanks!

alexweissman:

kutec:
Thanks :)

Issue #3523, opened 2015-06-25
Up arrow generates JS error


Karmalakas:
https://jsfiddle.net/urkaj76k/

Pressing Up arrow key, generates _Uncaught TypeError: Cannot read property 'top' of undefined_

kevin-brown:
I was able to reproduce this, here's the traceback

```js
Uncaught TypeError: Cannot read property 'top' of undefined
(anonymous function) @ select2.full.js:1076
S2.define.Observable.invoke @ select2.full.js:614
S2.define.Observable.trigger @ select2.full.js:604
S2.define.Select2.trigger @ select2.full.js:5167
(anonymous function) @ select2.full.js:5100
S2.define.Observable.invoke @ select2.full.js:614
S2.define.Observable.trigger @ select2.full.js:604
S2.define.Select2.trigger @ select2.full.js:5167
(anonymous function) @ select2.full.js:5018
S2.define.Observable.invoke @ select2.full.js:614
S2.define.Observable.trigger @ select2.full.js:608
(anonymous function) @ select2.full.js:1820
jQuery.event.dispatch @ jquery-git.js:4979
jQuery.event.add.elemData.handle @ jquery-git.js:4659
```

The issue appears to be that hitting UP will trigger the highlighting routines to shift the highlight, but that can't be done because there are no options in the dropdown. This should only happen if there is nothing in the results box, which currently can only happen if there is an AJAX error. Otherwise it will display the normal "no results found" message.

I'm tempted to call this a duplicate of https://github.com/select2/select2/issues/3501.

Karmalakas:
Actually I noticed this when I was trying to make options appear for multiple select, but was passing Ajax delay and dataType parameters. So no Ajax call was made, but also no options displayed :) That link to JsFiddle is what I mean

raapperez:
This error still occurs in version 4.0.5.

alexweissman:
Seems like it only happens in controls populated via AJAX, when no results are returned by a request (even a successful request).

alexweissman:
Fixed in #5127.

PR #5127, opened 2017-11-14
Fix up arrow error when there is no options

raapperez:
This pull request includes a

- [x] Bug fix
- [ ] New feature
- [ ] Translation

- Select2 throws an error when user presses up arrow key and there is no options. I saw that in this case currentIndex in results.js is -1. So I just changed the function exit condition.
- This error is in select2 since 2015 and is causing more than 1k errors per week in our logs. If this is not the correct fix. Please fix it.

https://github.com/select2/select2/issues/3523

### DIFF
--- a/src/js/select2/results.js
+++ b/src/js/select2/results.js
@@ -340,7 +340,8 @@ define([
       var currentIndex = $options.index($highlighted);
 
       // If we are already at te top, don't move further
-      if (currentIndex === 0) {
+      // If no options, currentIndex will be -1
+      if (currentIndex <= 0) {
         return;
       }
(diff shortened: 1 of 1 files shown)


alexweissman:
> This error is in select2 since 2015 and is causing more than 1k errors per week in our logs. If this is not the correct fix. Please fix it.

There is no central full-time organization responsible for developing Select2.  Everyone contributes their free time when they can.  Thank you for finding and fixing this bug, but please do not dictate demands.

alexweissman:
Will be released in `4.0.6-rc.1`.

PR #5140, opened 2017-11-21
Fixes #5137 - Switch to select2 inside label tag loses focus

### LINKED ISSUE #5137: select2 Inside of label Not Functioning Correctly
A select2 does not work correctly when inside of a label element. In this situation the select2 does not correctly handle text entry.

For example, using this select inside of the a label:
```
<label>
  Select2 Inside Label
  <select>
    <option>a</option>
    <option>b</option>
  </select>
</label>
```

Setting this up as a select2 will create the select2, but when you select the select2 and try to type, the text is not entered into the textbox. I also saw very odd behavior with this situation in my application where typing seemed to open and search inside the select element the select2 replaced.

Here's a working demonstration of this problem:
https://jsfiddle.net/Gavin_Paolucci_Kleinow/pmvbw2z1/2/

nino-s:
This pull request includes a

- [x] Bug fix
- [ ] New feature
- [ ] Translation

- Focus container only if container does not have focus yet

[#5137](https://github.com/select2/select2/issues/5137)

### DIFF
--- a/src/js/select2/dropdown/search.js
+++ b/src/js/select2/dropdown/search.js
@@ -63,7 +63,7 @@ define([
     });
 
     container.on('focus', function () {
-      if (!container.isOpen()) {
+      if (!container.hasFocus()) {
         self.$search.focus();
       }
     });


alexweissman:
Can you add a test to demonstrate that this solves the problem in #5137?

nino-s:
I tried to do a test case for it, but i cannot simulate it. The bug in #5137 only occurs, if you open the select container with real mouse click. I tried to simulate a mousedown event, but the bug does not show up.

alexweissman:
Ah ok, no problem.  This is why we have pre-releases!

Stadly:
Any news on this? Would be great if this PR could be merged :)

hbrecht:
Yes, we also need this - we manually inserted the change into our code...

os1r1s110:
Quick question, wouldn't that also need to be applied to https://github.com/select2/select2/blob/develop/src/js/select2/selection/single.js#L61 ? I think it's a good idea to only apply focus if it doesn't already have it (but I didn't test anything and don't know the side effects of that change, I'm really just curious :) )

hbrecht:
No - at least in our case, the focus then stays on the underlying <select>

os1r1s110:
Don't know if there will be any patch release anytime soon, I guess I will also copy it locally and apply the fix there :) Or keep using 4.0.3 which is the last version before the bug was introduced

PR #5160, opened 2017-12-09
Fix #3320 - Prevent select2:open when clearing selection

### LINKED ISSUE #3320: Prevent select2:open when clearing selection

taai:
This pull request includes a

- [x ] Bug fix

- Prevent select2:open when clearing selection.
- Include originalEvent in ```clear``` event, the same as it has been included in ```unselect``` event in ```multiple.js``` file

Link to the issue ticket: https://github.com/select2/select2/issues/3320

### DIFF
--- a/src/js/select2/selection/allowClear.js
+++ b/src/js/select2/selection/allowClear.js
@@ -30,6 +30,8 @@ define([
   };
 
   AllowClear.prototype._handleClear = function (_, evt) {
+    evt.stopPropagation();
+
     // Ignore the event if it is disabled
     if (this.options.get('disabled')) {
       return;
@@ -42,14 +44,13 @@ define([
       return;
     }
 
-    evt.stopPropagation();
-
     var data = Utils.GetData($clear[0], 'data');
 
     var previousVal = this.$element.val();
     this.$element.val(this.placeholder.id);
 
     var unselectData = {
+      originalEvent: evt,
       data: data
     };
     this.trigger('clear', unselectData);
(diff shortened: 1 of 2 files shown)


alexweissman:
Awesome, thanks.  So, now we need to decide if this is a "bug fix" (and release with 4.0.6), or if this is going to be a significant enough change that we need to put it off to 4.1.

taai:
@alexweissman This issue has been existing at least since May 6, 2015 (the creation date of that issue ticket). It has annoyed me probably even longer until now I decided to dig into code and resolve it. I think it is a bug and I suggest to publish the fix as soon as possible. And it doesn't change API, there are no new option params, so nothing will be broken if this would be published as a bugfix...

alexweissman:
Seems fair.

asdsteven:
With the change in `multiple.js`, the selections of multiple select cannot be unselected at all.  To resolve the issue, the line `evn.stopPropagation()` should be moved a bit lower, to after `self.trigger('unselect', ...)`.  Tested with version 4.0.3.

gorj-tessella:
This pull request was never merged and as of 4.0.6rc1 the weird/inconsistent open-dropdown-on-unselect behavior still happens. Was it lost in the shuffle?

taai:
Still not merged in v4.0.7. I wrote a test case: https://jsbin.com/sacunojefe/edit?js,output

Issue #3168, opened 2015-03-17
Deleting selections from multi-select using backspace causes browser back in IE11


RochDs:
Deleting the tag using backspace causes browser back. This occurs in IE 11 but not in Chrome 41.0.

Jsfiddle: http://jsfiddle.net/6XvqX/822/

kevin-brown:
I was able to reproduce the issue locally in IE, will do more testing later.

RochDs:
Thank you Kevin!

ithasyou:
Any news regarding this issue? It's really annoying

ithasyou:
Still having the problem in IE9 and rc2

ithasyou:
Solved modifying the following code:
this.$selection.on('keydown', '.select2-search--inline', function (evt) {
	evt.stopPropagation();
	self.trigger('keypress', evt);
	self._keyUpPrevented = evt.isDefaultPrevented();
	var key = evt.which;
	if (key === KEYS.BACKSPACE && self.$search.val() === '') {
		var $previousChoice = self.$searchContainer.prev('.select2-selection__choice');
		if ($previousChoice.length > 0) {
			var item = $previousChoice.data('data');
			self.searchRemoveChoice(item);
			self.$search.val("").blur().focus();
			evt.preventDefault();
		}
	}
});

kevin-brown:
This is fixed on master but is not yet in a release.

RochDs:
Hi Kevin, when will this fix be available in the release?

kevin-brown:
This _should have_ been included in the 4.0.0 release that went out about two weeks ago. Can you confirm if you are still having issues with it?

RochDs:
Hi Kevin, I have not yet taken the latest release. I will get it, and revert incase of issues.

nfriend:
I am experiencing this issue in IE 11 when using `select2.js` or `select2.min.js`, but *not* when using `select2.full.js`.

As a side note, the fix in `select2.full.js` disables the backspace key completely after the deletion of a single tag, which is certainly better than navigating backwards in history, but still not as satisfactory as the expected behavior of continued tag deletion.

kevin-brown:
> I am experiencing this issue in IE 11 when using select2.js or select2.min.js, but not when using select2.full.js.

Do you want to create a pair of jsfiddles (one with the full, one with the regular build) and a step by step set of instructions for how you are testing?

I'm not sure what would be included in the full builds that would be affecting this.

[...]

PR #5161, opened 2017-12-09
Fixes IE browser back issue when clearing selections using backspace key

ToreOlavKristiansen:
This pull request includes a

- [X] Bug fix
- [ ] New feature
- [ ] Translation

I changed allowClear.js to:
- Fix IE browser back issue when clearing selections using backspace key.
- Fix issue with not being able to start typing for both IE and Chrome after clearing using backspace or delete key.

#3168


alexweissman:
@cordmata @dmarginian can you please take a look at this PR, and see if it solves your other issues with IE?

dist/js/select2.full.js
```diff
@@ -2019,7 +2027,13 @@ S2.define('select2/selection/search',[
 
     this.resizeSearch();
     if (searchHadFocus) {
-      this.$search.focus();
+      var isTagInput = this.$element.find('[data-select2-tag]').length;
+      if (isTagInput) {
```
TRoorda: To make this work in IE11 with multiple tag support, I had to combine this code with the code presented in a comment on https://github.com/select2/select2/issues/4398:
```var isTagInput = this.$element.find('[data-select2-tag]').length;
if (isTagInput) {
    // fix IE11 bug where tag input lost focus
    window.setTimeout(function() {
        self.$element.focus();
    }, 0);
} else {
    window.setTimeout(function() {
        self.$search.focus();
    }, 0);
}

caohanh1502:
I have a problem here. When I apply this fix, the browser would focus on the select2 again if I click on an input while the dropdown selection is still opened. It does not happen in v4.0.3.
I also tested on the newest dev version 4.0.6-rc.1 with Chrome 68 and IE 11, and the same things happen.

Codepen: [https://codepen.io/caohanh1502/pen/MBRQWR](https://codepen.io/caohanh1502/pen/MBRQWR)
Debug view for IE 11: [https://s.codepen.io/caohanh1502/debug/MBRQWR/vPrKKzzxBPKr](https://s.codepen.io/caohanh1502/debug/MBRQWR/vPrKKzzxBPKr)

Click on the select2 to open the dropdown selection, then click on the input of datetimepicker. Result:
- Chrome: The datetimepicker is shown but it hides immediately. The focus is set to the select2, and I cannot type anything into the input.
- IE 11: The datetimepicker does not shown (I did not see the div appear in the IE's DOM explorer). The focus is set to the select2, but I can type into the input.
- (Same with the normal input).

pedrofurtado:
https://github.com/select2/select2/pull/3394

dist/js/select2.full.js
andreikondratev: This one worked for me

kevin-brown:

Issue #5022, opened 2017-08-29
When I click on select2 my page scrolls to top


skazhikadyadya:
Hello. When I click on select2 dropdown on my page page scroll to top. I was tested in Mozilla Firefox and latest Google Chrome. Hown I can solve that problem?

etharshrouf:
This problem happened with me too

alexweissman:
Could you please check the open PRs to see if a patch has already been submitted?  If not, could one of you please create a jsfiddle to demonstrate this behavior?

lk77:
the entire page or just the dropdown ?

etharshrouf:
when click in selectbox first time the page scroll to top, but when scroll down to selectbox and click on selectbox page not scroll, but when make refresh and click to select the page scroll again

alexweissman:
Please create a fully-functioning demonstration of this behavior (jsfiddle is fine).

etharshrouf:
https://jsfiddle.net/fL9fj2sL/12/

alexweissman:
Could you try adding some linebreaks or something to this fiddle?  As it is, this page doesn't seem to scroll at all.

etharshrouf:
http://ts.com.ps/toof/toof/buyer-individual.html

etharshrouf:
http://ts.com.ps/toof/toof/buyer-institute.html

alexweissman:
Thanks. This appears to be a duplicate of #4810.  I've tested your page and can confirm this behavior on your page.

It's hard to say what's going on here - I don't see any event handlers other than select2 and jQuery attached to that element, so I don't think it would be interference from another plugin.  I do notice that you're using jQuery 3, as is the case in #4810.  Could you try (temporarily) downgrading to jQuery 2.x, and see if the problem persists?

etharshrouf:
i downgrading jquery version to 2.x , and problem not solved

OlegKi:
I wanted just make the demo https://jsfiddle.net/fL9fj2sL/12/ created by **etharshrouf** (see [the comment](https://github.com/select2/select2/issues/5022#issuecomment-326696661))  more clear.

[...]

PR #5150, opened 2017-11-29
Add scrollAfterSelect as a configurable option for multiselect dropdowns to allow toggling of highlightFirstItem() behaviour

boweihan:
Resolves issue for multiselects using closeOnSelect: false that caused the list of results to scroll to the first selection after each select/unselect. This behaviour was intentional to deal with infinite scroll UI issues but it created an issue with multiselect dropdown boxes of fixed length. This pull request adds a configurable option to toggle between these two desirable behaviours.

This pull request includes a

- [x] Bug fix
- [ ] New feature
- [ ] Translation

The following changes were made

- Added configurable option "scrollOnSelect"
- If set to TRUE multiselect dropdowns will scroll to the first item on the list after "select" or "unselect" events (i.e. for infinite scroll case)
- If set to FALSE highlight will remain on selected item.

I ran into this issue when implementing select2 multiselect dropdowns with a fixed item length. It was confusing to have the focus move to the first item after each select/unselect event.

If this is related to an existing ticket, include a link to it as well.
https://github.com/select2/select2/pull/4869 - is tagged needs more work. I noticed it hasn't been touched for a while so I'm following up on this.
https://github.com/select2/select2/issues/4417 - discussion for where this problem originated. (this particular commit - https://github.com/select2/select2/commit/9f581285d88128b29a01fc1e5fd2d445d610b553)

PR #5183, opened 2018-01-04
Removing highlightFirstItem calls to fix scroll

lk77:
Hello,

i'm proposing this change to correct the scroll bug when selecting and unselecting, the scrollbar always return to the first selected item.

it's related to issue : https://github.com/select2/select2/issues/5022

i have removed the calls to highlightFirstItem in select and unselect events in results.js
these calls where added after in v 4.0.0
i have left the call in results:all because it seems it cause a problem with the enter key i don't know why.

tell me what you think of it.

thanks.

### DIFF
--- a/src/js/select2/results.js
+++ b/src/js/select2/results.js
@@ -276,7 +276,6 @@ define([
       }
 
       self.setClasses();
-      self.highlightFirstItem();
     });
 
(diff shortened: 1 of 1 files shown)


pedrofurtado:
Already resolved in https://github.com/select2/select2/pull/5150

PR #5198, opened 2018-01-19
fix: undefined validation, jquery3 compatible

waitingsong:
This pull request includes a

- [X ] Bug fix
- [ ] New feature
- [ ] Translation

-  fix: Cannot read property 'toString' of undefined
- fix: error JQMIGRATE: jQuery.fn.offset() requires an element connected to a document for jQuery3
-


pedrofurtado:
@waitingsong Add some unit test. After it, the PR is ok for merge

Issue #3214, opened 2015-03-31
Get the current search text


anne-sophie-balestra:
Hi, 
I am working with an empty select2 that I would like to populate according to the search term (I know we can do this with an ajax function but I always get the error : data is undefined) 
So I am trying to do it on my own, giving the search term on keyUp to an ajax function (that will return my JSON)

Anyway, I would like to know if there is anyway to get the search term (on key up, while typing anyway) from my select2 ?

Thanks in advance for your answers,
Anne-Sophie

kevin-brown:
This appears to be more of a usage question than a bug report or common feature request.  As we continue to keep the Select2 issue tracker geared towards these two areas, bug reports and common feature requests, we are redirecting usage questions to other communities such as the [mailing list][ml], [IRC][irc], or [Stack Overflow][so].

[ml]: https://select2.github.io/community.html
[irc]: https://select2.github.io/community.html
[so]: http://stackoverflow.com/questions/tagged/jquery-select2?sort=votes

PR #5245, opened 2018-03-07
Replace Deprecated .blur() and .focus() Calls

davidcochrum:
This pull request includes a

- [x] Bug fix
- [ ] New feature
- [ ] Translation

All event shorthand methods have been deprecated as of jQuery 3.3.0 ([#3214](https://github.com/jquery/jquery/issues/3214), [022b69a4](https://github.com/jquery/jquery/commit/022b69a44e42684bdd0029dd456bedb3b495cc24))

- Replaced calls to `.blur()` with `.trigger('blur')`
- Replaced calls to `.focus()` with `.trigger('focus')`

The change is fully backwards compatible across all versions of jQuery

### DIFF
--- a/src/js/select2/dropdown/search.js
+++ b/src/js/select2/dropdown/search.js
@@ -60,7 +60,7 @@ define([
       self.$search.attr('tabindex', -1);
 
       self.$search.val('');
-      self.$search.blur();
+      self.$search.trigger('blur');
     });
 
     container.on('focus', function () {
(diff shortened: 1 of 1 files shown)


cosmoKenney:
I just upgraded a project from select2 4.0.6-rc.0 to 4.0.47. And the typescript compiler is complaining about calls like this:
$("#someId").select2( "focus" );
Is that method calling style no longer supported?
The error doesn't make any sense to me:
`TS2559: Type '"focus"' has no properties in common with type 'Options<DataFormat | GroupedDataFormat, any>'.`

Issue #5249, opened 2018-03-14
Clicking outside of the body doesn't close the select


nils703:
When you open the select and click somewhere outside the body of the page it doesn't close the select.

To reproduce the issue just click below the select in the following example:
https://jsfiddle.net/sxpa6p69/5/

pedrofurtado:
Discussion will continue here -> https://github.com/select2/select2/pull/5250

PR #5250, opened 2018-03-14
Changing the close event from body to document

nils703:
This pull request includes a

- [X] Bug fix
- [ ] New feature
- [ ] Translation

- Fixing issue #5249 by changing the close event from document.body to document


taoqf:
I prefer a mask rather than a global body event.

PR #5283, opened 2018-04-28
Fix typo in uk locale

ilyar:
This pull request includes a

- [x] Typo fix
- [x] Translation


kevin-brown:
Hi @ilyar,

Unfortunately we are no longer accepting patches for the 3.5.x line of Select2, since it is considered out of support. We would highly recommend upgrading to the 4.0.x line of Select2, which is currently supported.

As far as I can tell, these typos have been corrected in [the latest version of these translations](https://github.com/select2/select2/blob/3fb9e4fceb43d4520f7a6fc7d842abeceb8ddf50/src/js/select2/i18n/uk.js).

Issue #4530, opened 2016-08-12
Add tooltip (title) to the 'remove all' 'X' icon


Loadmaster:
## Enhancement request

Could we please add a `title` attribute to the "remove all tags" '×' span element? In the `AllowClear.prototype.update` function definition, something along the lines of:

```
  AllowClear.prototype.update = function (decorated, data) {
    ...
    var $remove = $(
      '<span class="select2-selection__clear" title="Remove all items">' +    //<<modified line
        '&times;' +
      '</span>'
    );
    ...
  };
```

(I would do it myself (for v4.0.3), but I don't have the GIT infrastructure on my PC here.)

If we really wanted to get fancy, this could be (yet another) configuration options setting:

```
var select2Opts = { ..., removeAllTip: "Remove all items", ... };
```

alexweissman:
This appears to be a duplicate of #4456.

alexweissman:
Actually, #4456 is about the clear icon for individual selections, whereas this request is for the "clear all" icon when using `allowClear`.

nisha-kaushik:
I would like to pick this up as my initial contribution.

PR #5291, opened 2018-05-14
#4530 Add tooltip (title) to the 'remove all' 'X' icon.

nisha-kaushik:
This pull request includes a

- [ ] Bug fix
- [ ] New feature
- [ ] Translation

-
-
-

### DIFF
--- a/src/js/select2/selection/allowClear.js
+++ b/src/js/select2/selection/allowClear.js
@@ -98,7 +98,7 @@ define([
     }
 
     var $remove = $(
-      '<span class="select2-selection__clear">' +
+      '<span class="select2-selection__clear" title="Remove all items">' +
         '&times;' +
       '</span>'
     );
--- a/tests/selection/allowClear-tests.js
+++ b/tests/selection/allowClear-tests.js
@@ -79,6 +79,26 @@ test('clear is displayed for placeholder', function (assert) {
   );
 });
 
+test('clear icon should have title displayed', function (assert) {
+  var selection = new AllowClearPlaceholder(
+    $('#qunit-fixture .single-with-placeholder'),
+    allowClearOptions
+  );
+
+  var $selection = selection.render();
+
+  selection.update([{
+    id: 'one',
+    test: 'one'
+  }]);
+
+  assert.equal(
+    $selection.find('.select2-selection__clear').attr('title'),
+    'Remove all items',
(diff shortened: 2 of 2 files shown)


pedrofurtado:
@nisha-kaushik This title attribute must show the I18n value, that select2 already provides for other texts. Modify the PR to show the title from i18n.
That's a great PR, but needs support for i18n.

nisha-kaushik:
Hi, 

I am new to open source and i18n, do we need to add the text for every language in all those files?

e.g: es.js and en.js etc?

pedrofurtado:
@nisha-kaushik Wow, Good job! My only one point is, please, to remove all files from `dist` folder of PR. They are generated automatically during build process. So, maintain only the changes in `src` folder.
After it, the PR is ok for merge 🎉

nisha-kaushik:
There are 3 TO Dos , I couldn't find conversion in Lower Sorbian., Upper Sorbian and turkmen. Can you pleas ehelp me with that.

pedrofurtado:
@nisha-kaushik There is no problem, we can maintain the default as english string. Other users from community (that know this languages) will fix this missing messages, if necessary.

pedrofurtado:
@nisha-kaushik Just only one more point: Add just one test that has the same code you did in PR, but testing in another locale. Then, we will be safe that this feature is tested in default language (english) and the others languages.

nisha-kaushik:
Ok, I have reverted dist folder, Let me know for further changes.

pedrofurtado:
@nisha-kaushik It is something like this below:

```javascript
test('clear icon should have title displayed in another locale', function (assert) {
  var selection = new AllowClearPlaceholder(
    $('#qunit-fixture .single-with-placeholder'),
    new Options({
      placeholder: {
        id: 'placeholder',
        text: 'This is the placeholder'
      },
      allowClear: true,
      language: 'ANY_LOCALE'
    });
  );
   var $selection = selection.render();
   selection.update([{
    id: 'one',
    test: 'one'
  }]);
   assert.equal(
    $selection.find('.select2-selection__clear').attr('title'),
    'TEXT OF THE LOCALE',
    'The clear icon should have title displayed'
  );
});
```

nisha-kaushik:
I tried the same,  
with language : 'af', but it still renders with english :(

Issue #5294, opened 2018-05-18
[Feature Request] Add "nepali" localization


samundra:
**Feature Request**

~- Add Nepali (np) localization support for the select2~
- Add Nepali (ne) localization support for the select2

PR #5295, opened 2018-05-18
Add Nepali (ne) localization

samundra:
This pull request includes a

- [ ] Bug fix
- [ ] New feature
- [x] Translation

~- Added Nepali (np) localization support~
- Added Nepali (ne) localization support

This Pull Request fixes the https://github.com/select2/select2/issues/5294 issue.


kevin-brown:
Hi @samundra,

Thanks for submitting a translation to Select2! I've reached out to some contacts who are familiar with Nepali to see if they can verify the translations here.

Before we can accept this translation though, we noticed that the incorrect ISO 639 language code appears to be used. The [language code for the Nepali language appears to be `ne`](http://www.localeplanet.com/icu/iso639.html#ne), but these translations were submitted under `np` (which appears to be unassigned). In order for browsers to automatically pick up the language, and for many systems to indicate the right language, we use the ISO 639 language code to name our translations.

If you can rename the file to `ne.js`, this will help speed up the process for accepting this pull request. Thank you!

samundra:
Hi @kevin-brown 

I have renamed the file name. Please, let me know if there is anything I need to address.

Regards,
Samundra

PR #5296, opened 2018-05-18
Fix #3604: Options programmatically added/removed to/from optgroups are not rendered

### LINKED ISSUE #3604: Options programmatically added/removed to/from optgroups are not rendered
In [this](http://jsfiddle.net/z0hhkbg0/) fiddle I have two selects: one with just an option and another with an option within an optgroup.

After running the fiddle, the select with just an option has three options, as expected, but the select with the optgroup only has two. It seems as if the select with the optgroup does not render any newly appended options once the select has been opened.

Am I doing something unintended or is this a bug?

LucaRainone:
solution is don't cache optgroup data.

This pull request includes a

- [x] Bug fix

- data optgroups is not cached anymore

Fixing issue https://github.com/select2/select2/issues/3604:


kevin-brown:

LucaRainone:
@kevin-brown  I've opened another PR with modifications only in src, here: https://github.com/select2/select2/pull/5465

Issue #5300, opened 2018-05-24
The default matcher should trim the term


EvgenyOrekhov:
In most cases it is desired to trim the search term before matching, therefore it should be done by default. The trimming would be especially useful when using tags in "multiple" mode, since it would prevent the adding of duplicate entries.

Steps to reproduce:

1. Go to https://jsbin.com/kinotukika/edit?html,js,output
2. Type "test " (with a space at the end) into the input

Expected: "test " should match "test".
Actual: nothing matches, "No results found" message is shown.

Select2 version: any
Browser version: any
OS version: any

nikhilbarar:
Can i pick this up?

nikhilbarar:
Any update on this? I'd like to send a PR for this

pedrofurtado:
This discussion will continue inside PR -> https://github.com/select2/select2/pull/5318

PR #5318, opened 2018-06-23
#5300 - Fixed

nikhilbarar:
This pull request includes a

[5300] Bug fix

Trimmed the term for default and matcher

Fixed Issue #5300


pedrofurtado:
@nikhilbarar Please, remove all files in this PR from `dist` folder. They are automatically generated during build process.

src/js/select2/compat/matcher.js
```diff
@@ -8,6 +8,9 @@ define([
       if (params.term == null || $.trim(params.term) === '') {
         return match;
       }
+      else {
```
hakan0xFF: the `params.term` is trimmed using a jQuery function so you could do

```js
params.term = $.trim(params.term)
if (params.term === '') {
  return match;
}
```

there is no need of an `else`

src/js/select2/defaults.js
```diff
@@ -311,6 +311,9 @@ define([
       if ($.trim(params.term) === '') {
         return data;
       }
+      else {
```
hakan0xFF: see comment above

src/js/select2/compat/matcher.js
pedrofurtado: @hakan0xFF But we have to be aware if `$.trim` works well when parameter is `null`. The result of `$.trim` is always a string, even blank? Or is there some case that it returns null? This should be included in the code.

hakan0xFF: if the value/parameter is null the trim function returns an empty string. so it returns always a string. see https://github.com/select2/select2/blob/develop/vendor/jquery-2.1.0.js#L402

PR #5353, opened 2018-08-31
fixes #3663 "closing dropdown which is out of the viewport causes viewport to scroll"

### LINKED ISSUE #3663: closing dropdown which is out of the viewport causes viewport to scroll
Version: Select2 4.0

Example: https://jsfiddle.net/z7L1px15/

To reproduce:
1. In the above fiddle click the select2 dropdown to open it.
2. Scroll the viewport down to the 2nd input and click it to focus on it

What is expected:
The 2nd input will have the focus and be visible.

What happens:
The 2nd input receives the focus, but the viewport scrolls up to the select2 dropdown so the focused input is no longer visible.

This seems like somewhat of an edge case but we see people hit it with some regularity on a longish form where the submit button is at the bottom of the page. Users will decide not to change a select2 enabled field after opening, and scroll down and click the submit button to save other changes, then get confused when the page jumps back up. They have to scroll back down to the bottom and click the submit button again for the form to be submitted.

The line responsible is https://github.com/select2/select2/blob/master/src/js/select2/selection/base.js#L84
That line returns the focus to the element after the dropdown closes. I think what we really want is for this focus event to occur only when a value is selected, not every time the dropdown is closed.

LOK-Soft:
This pull request includes a

- [x] Bug fix

- focus will only be set if the user selected an Element, not when clicking outside the select2 elements

This deals with #3663


pedrofurtado:
@LOK-Soft Please, remove all files of PR from `dist` folder. They are automatically generated during build process. And there are some conflicts :+1:

LOK-Soft:
@pedrofurtado removed all changes in /dist/ now :)

LOK-Soft:
Is there a possibility to merge this request soon?

LOK-Soft:
Any news here?

SerdarMustafa1:
> Any news here?

I was also hoping for an update on this.

Victor-Prodanov:
Hi,

I think the problem still exists for multiple select with "closeOnSelect" set to "false".

Thank you.

kevin-brown:
This is definitely still a bug, but because it wasn't reviewed fast enough there are now merge conflicts (mostly because we changed how focus was being assigned recently).

I'm not entirely sure that the fix here is the right approach, but then again I also don't know a great way of detecting whether or not the close requires the focus to be shifted. Would it be possible for you to create a unit test (along the same idea as the other focusing tests we currently have) so we can confirm that this both fixes the issue and won't become an issue again in the future?

PR #5357, opened 2018-09-05
When the select2 instance is multiple with ajax and you click with mouse twice or more in the field, the focus stay in this field forever.

pedrofurtado:
@ivaynberg @alexweissman 

When the select2 instance is multiple with ajax and you click with mouse twice or more in the field, the focus stay in this field forever.


pedrofurtado:
I forgot to remove the correct line -> https://github.com/select2/select2/commit/933189b92e352d62373b7739a28c266a129f97c9

PR #5371, opened 2018-09-21
Master

MStoehr:
This pull request includes a

- [X] Bug fix
- [ ] New feature
- [ ] Translation

- if the templateResult function returns null, the option will now no longer be added to the select2-results
- therefore the keyboard-navigation will now no longer select elements set to "display: none"


kevin-brown:

PR #5373, opened 2018-09-25
Passed dropdown parent shall be wrapped in $()

mlt:
This pull request includes a

- [X] Bug fix
- [ ] New feature
- [ ] Translation

- Call `$()` on passed *dropdownParent* / *data-dropdown-parent* in *AttachBody*
- It seems generally safe to call `$()` on already a *jQuery* objects


mlt:
ehm...is there anything holding this one that I shall fix? I kind of rely on it and it has a test.

kevin-brown:
I know it's been a few months and we dropped the ball on this one, but this generally looks good and it's my goal to get it merged in.

But before we do that, is there any chance you can split your test changes out so we have two different tests? One for passing in a jQuery object and the other for passing in the selector.

mlt:
I made a separate test case and rebased my branch.

Peter-Potts:
Hi, thanks mlt for proposing this change.
I would also find it valuable - using the 'data-dropdown-parent' attribute (eg ="#myID') actually seems to fail otherwise.
Cheers

mlt:
bump?

PR #5413, opened 2018-11-13
fix result option highlighting for screenreaders

JoshMoreno:
Addresses #3735 and pull request [#3821](https://github.com/select2/select2/pull/3821)

Not very familiar with the codebase and didn't have much time. Hope this helps move the accessibility improvements in the right direction.

From what I can deduce from various examples`aria-activedescendent` should be updated to the id of the current highlighted result. Also, `aria-activedescendent` is on the `input` in the examples.

Examples:
- http://oaa-accessibility.org/example/11/
- https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html

### DIFF
--- a/src/js/select2/results.js
+++ b/src/js/select2/results.js
@@ -308,6 +308,9 @@ define([
       self.$results.attr('aria-expanded', 'false');
       self.$results.attr('aria-hidden', 'true');
       self.$results.removeAttr('aria-activedescendant');
+
+      // remove aria-activedescendant from the search input for screen readers
+      this.dropdown.$search.removeAttr('aria-activedescendant');
     });
 
     container.on('results:toggle', function () {
@@ -360,6 +363,9 @@ define([
 
       var $next = $options.eq(nextIndex);
 
+      // update the search input's aria-activedescendant for screen readers
+	  this.dropdown.$search.attr('aria-activedescendant', $next.attr('id'));
+
       $next.trigger('mouseenter');
 
       var currentOffset = self.$results.offset().top;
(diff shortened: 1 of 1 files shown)


kevin-brown:
Hi there,

Thanks for your contribution to Select2. This is something I'm definitely interested in seeing land in Select2 4, but unfortunately right now there appear to be some failing tests.

It looks like those tests are failing because this is attempting to set the attribute for the search box within the results dropdown. Because Select2 works in both a single select mode (where the search is in the dropdown) and multiselect mode (where the search is in with the selection), we cannot rely on the location of the search. Or even on it existing, since it's possible to disable the search box.

My suggestion to fix this issue would be to look into moving these calls into the `Search.js` files which power the single select and multiselect dropdowns, so we don't accidentally tie the results to the search itself. If you need to pass data between the results and the search box, you can trigger an internal event by calling `.trigger()` and this will propagate to the other areas of Select2.

Let me know if you have any questions or need any assistance.

ScriptShot:
Hi, is there any progress or news here? It would be great if this were implemented :)

ScriptShot:
Knock knock...

kevin-brown:
Two notes for why this hasn't been accepted yet:

1. It's trying to set properties on the search box from within the results. Select2 enforces a clear separation here, since we have multiple places where a search box can be, and as a result this would need to be modified so these event handlers are present within the search components and not the results component.

2. This is missing unit tests. While we don't require them for all changes, we try to increase the coverage as we go and as a result I end up writing a lot of unit tests that other people chose not to write. Because of this, reviewing and accepting pull requests can go from a few minutes per pull request to over an hour, depending on how much work needs to go into fixing issues discovered through testing.

If someone is willing to submit an alternate pull request, or update this one, that fixes one or both of those issues, I'll gladly re-review it. If not, the Stale bot will show up in 60 days and close this one out.

kevin-brown:
Closing this off in favor of https://github.com/select2/select2/pull/5582, where this is implemented (among other things).

Issue #5378, opened 2018-09-28
I can not add new items to a Select2 list with Tagging support using IE 11


JacoFourie:
Hi all.

We use the Oracle Apex Plugin Version of this component. All works fine accept on IE we are not able to add new items to a list by typing and then hitting enter. I know you don't support the plugin here. I did also log the issue on the plug-in page. But I am hoping that somebody could point me to a fix as the plug-in is based on this code.

Here is a sample app. If you use Internet Explorer you will see you are not able to enter new text in the combo box. In any other browser it works 100%

https://apex.oracle.com/pls/apex/f?p=96087:2

Kind Regards.

Jaco.

Rocky-bg:
Hi all,

Could you please help me with some tip how to fix this issue because we use select2 a lot in our project and this IE issue is hard to understand what is going wrong. You can reproduce it on select2 example page : [](https://select2.org/tagging) with Internet Explorer 11.

Best Regards,
Rosen

JacoFourie:
@vandernorth did you fix the issue ? Sorry I have not been on here for some time.

JacoFourie:
Hi. I can confirm that this fixed the issue on IE. Thanks so much for that.

jbosse:
I can confirm this was either not actually resolved or a regression has occurred. You cannot add a tag in IE11. Tested with 4.0.10 min.js

jianbinwen:
This issue still exists in select2 4.0.12. Wonder why this issue was closed.

vandernorth:
Issue still exists when using IE11 and jquery v3.4.1 and select2 v4.0.13. 
Made a new PR #5821 
@kevin-brown any thoughts on this?

kevin-brown:
I was doing some testing of this issue and this appears to have been fixed as a side-effect of the changes made in https://github.com/select2/select2/pull/5842.

JacoFourie:
Hi. I can confirm that the error is still there on IE with version 4.0.13.
We tried to use the beta but that does funny things with text placement.

kevin-brown:
> We tried to use the beta but that does funny things with text placement.

Please open up a ticket if you are finding issues in 4.1.0-beta.1. Noting that custom themes may not yet have been updated to support the new version, so it is possible that the issues you are seeing may be related to theme bugs.

JacoFourie:
Thanks Kevin I will do so. Will you be fixing the IE bug in version 4.0.13 ?

PR #5453, opened 2019-02-11
Fix select2#5378 ie11 specific focus issue.

vandernorth:
This pull request includes a fix for #5378 

- [X] Bug fix
- [ ] New feature
- [ ] Translation

- Existing IE detection moved up a few lines so they are declared before they are used.
- On the focusout event of the search-input use the existing variable to detect IE
- If IE, and the event comes from the <body> tag, refocuses so we can keep typing

Issue: https://github.com/select2/select2/issues/5378

### DIFF
--- a/src/js/select2/selection/search.js
+++ b/src/js/select2/selection/search.js
@@ -29,6 +29,14 @@ define([
   Search.prototype.bind = function (decorated, container, $container) {
     var self = this;
 
+    // Try to detect the IE version should the `documentMode` property that
+    // is stored on the document. This is only implemented in IE and is
+    // slightly cleaner than doing a user agent check.
+    // This property is not available in Edge, but Edge also doesn't have
+    // this bug.
+    var msie = document.documentMode;
+    var disableInputEvents = msie && msie <= 11;
+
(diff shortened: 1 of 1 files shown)


exhtml:
Hi! First of all thanks so much for your superb plugin! When do you think this fix could be ready? Thank you!

LastZactionHero:
Thanks for the fix! This fixes an issue for me, would love to see it merged in.

vandernorth:
@kevin-brown can you merge this request?

kevin-brown:
I think this may actually be fixed in 4.0.8, which was released the other day and corrected some unexpected focus issues.

Is there any chance you can confirm the original issue still exists?

LastZactionHero:
4.0.8 does not resolve the IE11 issue for me. For me, issues manifests as only being able to type a single character. We've been using the changes from this PR in production for a few months.

tpetry:
I can confirm that 4.0.8 does not solve the bug. But the submitted pull-request does not solve the problem completely either:
* For chrome a tag is created after entering the first 2 characters
* For ie11 the focus is lost after entering the first tag

PR #5464, opened 2019-03-14
minor fix (greek omega used has no diacritic)

tgogos:
replaced `\u03C9` (GREEK SMALL LETTER OMEGA: `ω`)
with `\u03CE` (GREEK SMALL LETTER OMEGA WITH TONOS: `ώ`)

This pull request includes a

- [X] Bug fix
- [ ] New feature
- [ ] Translation

- replaced `\u03C9` (GREEK SMALL LETTER OMEGA: `ω`)
with `\u03CE` (GREEK SMALL LETTER OMEGA WITH TONOS: `ώ`)


tgogos:
@kevin-brown have you ever considered replacing the `diacritics.js` file with the use of [`String.prototype.normalize()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/normalize) to remove diacritics?

I was dealing with this issue (regarding the greek language) back in 2014 (see [Javascript - normalize accented greek characters](https://stackoverflow.com/questions/23346506/javascript-normalize-accented-greek-characters)) and I was "manually" doing the following replacements:

```
Greek and Coptic (Range: 0370— 03FF) 
ΆΑά -> α
ΈΕέ -> ε
ΉΗή -> η
ΊΪΙίΐ -> ι
ΌΟό -> ο
ΎΫΥΰϋύ -> υ
ΏΩώ -> ω

Greek Extended (Range: 1F00— 1FFF)
ἀἁἂἃἄἅἆἇὰάᾀᾁᾂᾃᾄᾅᾆᾇᾰᾱᾲᾳᾴᾶᾷἈἉἊἋἌἍἎἏᾈᾉᾊᾋᾌᾍᾎᾏᾸᾹᾺΆᾼ -> α
ἐἑἒἓἔἕὲέἘἙἚἛἜἝῈΈ -> ε
ἠἡἢἣἤἥἦἧὴήᾐᾑᾒᾓᾔᾕᾖᾗῂῃῄῆῇἨἩἪἫἬἭἮἯᾘᾙᾚᾛᾜᾝᾞᾟῊΉῌ -> η
ἰἱἲἳἴἵἶἷὶίῐῑῒΐῖῗἸἹἺἻἼἽἾἿῘῙῚΊ -> ι
ὀὁὂὃὄὅὸόὈὉὊὋὌὍῸΌ -> ο
ὐὑὒὓὔὕὖὗὺύῠῡῢΰῦῧὙὛὝὟῨῩῪΎ -> υ
ὠὡὢὣὤὥὦὧὼώᾠᾡᾢᾣᾤᾥᾦᾧῲῳῴῶῷὨὩὪὫὬὭὮὯᾨᾩᾪᾫᾬᾭᾮᾯῺΏῼ -> ω
ῤῥῬ -> ρ
```

I believe the complexity gets worse as you add more languages... Some years later, I found a way to do the same thing with the `normalize()` function and without keeping lists of characters.

## Example:

``` javascript
"ἀἁἂἃἄἅἆἇὰάᾀᾁᾂᾃᾄᾅᾆᾇᾰᾱᾲᾳᾴᾶᾷἈἉἊἋἌἍἎἏᾈᾉᾊᾋᾌᾍᾎᾏᾸᾹᾺΆᾼ".normalize('NFD').replace(/[\u0300-\u036f]/g, "");
"αααααααααααααααααααααααααΑΑΑΑΑΑΑΑΑΑΑΑΑΑΑΑΑΑΑΑΑ"
```

### How it works:

With `.normalize('NFD')` where `NFD` means *Normalization Form Canonical Decomposition* accented characters are decomposed to:

 -  the character itself
 -  followed by the equivalent Combining Diacritical Mark (see: range [0300-036f])
 -  which can then be easily removed with `.replace(/[\u0300-\u036f]/g, "")`

kevin-brown:
The problem that we run into with `String.prototype.normalize()` is that it is not supported by IE, which unfortunately Select2 currently supports.

When we eventually decide to drop support for IE and older versions of browsers, this is one of the first things I'm looking to replace with something native.

PR #5467, opened 2019-03-20
Feature/enable separating tokens in entries when quoted

imadx:
This pull request includes a

- [ ] Bug fix
- [X] New feature
- [ ] Translation

- updated tokenizer to accommodate quoted strings
- enabled tokenizer to be customized with specific characters
- added test cases


imadx:
@kevin-brown let me know if I need to get rid of the `package-lock.json`

kevin-brown:
> let me know if I need to get rid of the `package-lock.json`

Unfortunately, you do need to get rid of it.

---

Thanks for sending in this feature. I am going to take a shot at reviewing it in the near future, but right now I can't guarantee a whole lot. Because this is a feature request, and recently I've been focusing on fixing bugs, this takes a slightly lower priority.

But the good news is that since you included tests for this, it has a much higher chance of being accepted once we are reviewing new features.

src/js/select2/data/tokenizer.js
```diff
@@ -70,6 +70,12 @@ define([
   Tokenizer.prototype.tokenizer = function (_, params, options, callback) {
     var separators = options.get('tokenSeparators') || [];
     var term = params.term;
+
+    var enclosureOptions = (options.get('tokenEnclosures') || ['\"', '\'']);
```
kevin-brown: It would be preferred to have this default to an empty array instead of defaulting to quotes. This would allow us to better maintain backwards compatibility when this is brought in.

package-lock.json
```diff
@@ -0,0 +1,8015 @@
+{
```
kevin-brown: We don't include the `package-lock.json` since we only cover development dependencies now.

Issue #3575, opened 2015-07-14
set default language in v4.0


azza0091:
In version 3.5.2 I used to initialize the plugin by setting its Language, so that every select would have automatically used the selected Language. I was doing it like this:
$.extend($.fn.select2.defaults, $.fn.select2.locales['it']);

In version 4.0.0 I've seen that I can specify the Language in every call to the select2 plugin by setting the "Language" parameter to the desired Language.
My question is: is there a way, in version 4.0.0, to set the default Language once so that it would be common to all select2 that will be created? Like I was doing in version 3.5.2

kevin-brown:
You can override the default options for select2 using `$.fn.select2.defaults.set()`. So setting the default language would be

```js
$.fn.select2.defaults.set('language', 'it');
```

You can read more in [Setting default options](http://select2.github.io/options.html#defaults) which is hidden away on the [options page](http://select2.github.io/options.html).

bzx:
Even though I have the language file located in "i18n/pl.js" the default setting doesn't work for me - however it does work when I apply the language at the single selector. Is there some special syntax for setting the default language?

kevin-brown:
@bzx Make sure to set the default language before you initialize Select2.

http://jsbin.com/horubaladi/1/edit?html,js,output

bzx:
Yes, I do have that set in the first line of my .js file, like so:

http://jsbin.com/cacinahini/edit?js

--
TC

bzx:
Can you reproduce it?
Is this a new issue?

raydric:
I'm having the same issue.

`$.fn.select2.defaults.set('language', 'it');`
`console.table($.fn.select2.defaults);`

shows language correctly but the changes aren't applied on the UI

it works only with:

    $('.select2').select2({
        language: 'it'
    });

kevin-brown:
@bzx I pulled your JS into the original test case and still wasn't able to reproduce the issue.

> Can you reproduce it?

Unfortunately I wasn't able to reproduce it. Here my test case: http://jsbin.com/yemucofuro/1/edit?html,js,output

If you don't include the language file on the page before initializing Select2, no errors will be raised and Select2 will silently fall back to the English translations. This may be related to your issue.

raydric:
I'm combining and uglifying select2.js and the language file and getting this error:
`Select2: The language file for "./i18n/it" could not be automatically loaded. A fallback will be used instead.`

[...]

PR #5468, opened 2019-03-20
Adjust priority of Select2 language picker mechanism

deltoss:
This pull request includes a
- [X] Translation Fix

I noticed that Select2 internationalisation mechanisms picks up the language in this priority:
1. `language` option
2. `lang` property of the select element
3. Closest parent element with `lang`
4. Defaults set via the code `$.fn.select2.defaults.set('language', '<LANGUAGE>');`

I really think 3 and 4 should be switched around, especially since the HTML language code doesn't perfectly match up with Select2, looking at the list at (W3Schools).[https://www.w3schools.com/tags/ref_language_codes.asp].

For example, Select2 has **zh-CN**, which means traditional chinese, but in the above language code, it's actually **zh-Hans**. Thus, it's a good idea to switch the order of `3` and `4` around so it'll look like:
1. `language` option
2. `lang` property of the select element
3. Defaults set via the code `$.fn.select2.defaults.set('language', '<LANGUAGE>');`
4. Closest parent element with `lang`

For more information, see:
[Select2 Issue post](https://github.com/select2/select2/issues/3575#issuecomment-203550862)

Also... I can't seem to do an `npm install` on select2... so I couldn't compile, minify and test according to the `CONTRIBUTING.md` guide. I get the error below.

image

Thus, all I checked in was the changes to the source file. I did test it through trying it out on the browser with examples however.


kevin-brown:
I'm marking this as "needs additional work" since I'd prefer to see a test for this before accepting it. But the code change looks correct and I can add the test before I merge this if needed.

deltoss:
bump

src/js/select2/options.js
```diff
@@ -37,7 +37,8 @@ define([
     if (this.options.language == null) {
       if ($e.prop('lang')) {
         this.options.language = $e.prop('lang').toLowerCase();
-      } else if ($e.closest('[lang]').prop('lang')) {
+      } else if (typeof $.fn.select2.defaults.defaults.language !== 'string' &&
```
kevin-brown: This should actually be

```js
typeof Defaults.defaults.language !== 'string'`
```

Since `Defaults` is the singleton that is imported into this class, and it's the one that we assign to `$.fn.select2.defaults`. While these should be the same, we intentionally don't enforce that `$.fn.select2.defaults` must exist.

deltoss: Ok, thanks for the feedback. I'll make that change soon 👍.

kevin-brown:
Thanks for making those changes! This pull request is on track to land in 4.0.9.

If you're feeling adventurous, this would benefit from the creation of a test case or two to ensure that we don't break this in the future. You should be able to use the `data-*` tests as an example.

https://github.com/select2/select2/blob/2fce8ae6c4937b88c332ad3e30e135cf24504cf1/tests/options/data-tests.js

deltoss:
I think I'll try to attempt that sometime this week. Got the latest updated develop branch, and it looks like I can now run `npm install` and other `grunt` commands in `CONTRIBUTING.md` without a hitch now!

deltoss:
Added the tests as per recommendation. Please review, any feedback is appreciated. 👍

kevin-brown:
Thanks for adding the tests! While reading and reviewing this pull request a week ago, I came to the conclusion that the translation system for Select2 was broken and should really be fixed. Unfortunately that meant that this pull request was basically replaced by https://github.com/select2/select2/pull/5602, so I'm going to have to close it off.

The good news though is that the bug originally described in this pull request has been fixed and will be included in the 4.0.9 release.

I'm sorry for making you go through the work of correcting this pull request based on the feedback that was given. Especially considering the long amount of time that it took to initially review this in the first place.

PR #5507, opened 2019-05-07
Release Select2 4.0.7

kevin-brown:


nmfzone:
@kevin-brown When this release available in npmjs? Would need to update since 4.0.6 has an issue. (downgrade not an option).

PR #5512, opened 2019-05-10
updated instalation section

royopa:
This pull request includes a update on section Instalation from README.md, to inform the user how to install using bower, npm and yarn.


kevin-brown:
I have chosen not to include the JS package managers in the README because we currently support something like 10 different package managers because Select2 is used in more than just JS projects. I would rather not list them all out.

PR #5513, opened 2019-05-16
Feed back the keypress code that was responsible for the 'close' even…

6pac:
Feed back the keypress code that was responsible for the 'close' event as part of the event parameters. This is a non-breaking enhancement that doesn't affect any existing behaviour.

### DIFF
--- a/src/js/select2/core.js
+++ b/src/js/select2/core.js
@@ -325,7 +325,7 @@ define([
       if (self.isOpen()) {
         if (key === KEYS.ESC || key === KEYS.TAB ||
             (key === KEYS.UP && evt.altKey)) {
-          self.close(key);
+          self.close(evt);
 
           evt.preventDefault();
         } else if (key === KEYS.ENTER) {
@@ -468,12 +468,12 @@ define([
     this.trigger('query', {});
   };
 
-  Select2.prototype.close = function (key) {
+  Select2.prototype.close = function (evt) {
     if (!this.isOpen()) {
       return;
     }
 
-    this.trigger('close', { key: key });
+    this.trigger('close', { originalEvent : evt });
   };
 
   Select2.prototype.isOpen = function () {
--- a/src/js/select2/core.js
+++ b/src/js/select2/core.js
@@ -325,7 +325,7 @@ define([
       if (self.isOpen()) {
         if (key === KEYS.ESC || key === KEYS.TAB ||
             (key === KEYS.UP && evt.altKey)) {
-          self.close();
+          self.close(key);
 
           evt.preventDefault();
         } else if (key === KEYS.ENTER) {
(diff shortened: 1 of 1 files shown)


6pac:
this would be *really* handy for a lot of use cases, folks! Very safe and small. Please consider.

6pac:
bump

kevin-brown:
Just realized I never reviewed this one.

Generally I'm okay with things like this being done (providing event context when we do things), but in the past we've done this by passing `originalEvent` pointing to the original event which caused this action. This would be compatible with what you're looking for (getting the triggered keypress from the close event) while also maintaining consistency

https://github.com/select2/select2/blob/efbfd14414bd749143fbac386ff163cf5c5be8c8/src/js/select2/selection/single.js#L48
https://github.com/select2/select2/blob/efbfd14414bd749143fbac386ff163cf5c5be8c8/src/js/select2/selection/multiple.js#L50
https://github.com/select2/select2/blob/e5131d0cc8dffd98ba36a68f3d027bf79e763cb4/src/js/select2/results.js#L464

6pac:
OK, makes sense, changes made...

kevin-brown:
Thanks a bunch for making that change!

This PR has been accepted for the next release of Select2 (likely 4.0.12) and will be merged in the near future once the most recent release has stabilized.

6pac:
Thanks, that's great!

6pac:
bump

joepetrakovich:
This is a great start, but I think it would make sense to include the original event on all triggers of the `close` event.  

This one only works if the select closes from that small set of keyboard commands, but if it's closed via things like `selectOnClose ` or the many other calls to `self.close` or` trigger('close')`, the original event is missing.

This feature is nice because I can observe the tab or shift-tab key being pressed to close the dropdown, but it doesn't work if `selectOnClose `is enabled, and there is still no way to see if the enter key was pressed.  Also, if the user selects the original value that was already selected, the `close` event is trigger but the original event is missing.

6pac:
I was simply using this for keyboard navigation in a grid. For example, I need to know if it was closed with an Enter (focus remains or go to next control), Tab (go to next control) or Shift-Tab (go to last control).

joepetrakovich:
That's the same use case for me actually.

Tab and shift+tab for forward and backward, but in my case enter should go
forward as well.  And on top of that, both enter and tab should select the
highlighted option, which means I have to use SelectOnClose.

This enhancement covers some of these scenarios but I still can't get the
keycode to know when enter is pressed.

kevin-brown:
As always, we encourage further pull requests that expand this functionality out to more use cases.

6pac:
@joepetrakovich which grid? perhaps we should establish a simple test page? I'm using SlickGrid, but I don't suppose the grid really matters that much.

joepetrakovich:
@6pac yeah it probably doesn't matter.  I'm using DevExpress's ASPxGridView in batch edit mode (https://demos.devexpress.com/ASPxGridViewDemos/GridEditing/BatchEditing.aspx). 

They have their own built in autocomplete dropdown but it's slow and clunky compared to Select2.  If you add custom controls into the grid cells you have to re-implement the keyboard navigation.  This commit made it _a lot_ easier but like I said since the keypress event is lost in most other actions on the dropdown, I haven't gotten it fully fleshed out without weird non-compliant hacks.

I may try my hand at a pull request.

PR #5514, opened 2019-05-16
Update base.js

Miloage:
allow focus to other input box when the select2 opened.
（Originally, two clicks were needed to focus on the new input box.）


kevin-brown:
Just testing around on https://select2.org/appearance, I can't actually reproduce the issue that this claims to be fixing. If you could make a bug ticket or expand upon the issue in this pull request, I'll gladly reconsider this change (and re-open the PR).

PR #5518, opened 2019-05-23
Fix wrong search input width calculation when empty selection

### LINKED ISSUE #5517: Wrong width calculation of empty search input
Now search input width calculate as parent `innerWidth()`:
```
    if (this.$search.attr('placeholder') !== '') {
      width = this.$selection.find('.select2-selection__rendered').innerWidth();
```
This is wrong, because `ul.select2-selection__rendered` has left&right padding. 
So search input makes select2 container grows up.
To fix that it just need to calculate width by `width()`, that not count paddings.

JS fiddle: https://jsfiddle.net/karakum/2o6xzupe/

Empty select:
изображение

Not empty:
изображение

karakum:
This pull request includes a

- [X] Bug fix
- [ ] New feature
- [ ] Translation

This PR fix #5517

### DIFF
--- a/src/js/select2/selection/search.js
+++ b/src/js/select2/selection/search.js
@@ -214,7 +214,7 @@ define([
     var width = '';
 
     if (this.$search.attr('placeholder') !== '') {
-      width = this.$selection.find('.select2-selection__rendered').innerWidth();
+      width = this.$selection.find('.select2-selection__rendered').width();
     } else {
       var minimumWidth = this.$search.val().length + 1;
 


kevin-brown:
This makes sense as a change (based on the details in #5517). It would be solid if we could also get a test which covers this ([these might be a good starting example](https://github.com/select2/select2/blob/04fce559673684af8ec2912457a9d6380ea90314/tests/dropdown/positioning-tests.js)), just so we don't run into this again in the future.

Issue #5485, opened 2019-04-18
select2-4.0.5 doesn't seem to work well with jquery-3.4.0


kikonen:
Tried select2-4.0.5 with jquery-3.4.0 and started to encountered strange bug with focus logic. Especially multiselect widget was behaving very oddly (input focus disappeared, keyboard up/down started to change selected value in the field (no popup selector shown)).

Reverted back to jquery-3.3.1 and strange behaviours seemed to fix themselves. So I concluded that there might be some compatibility problem going on.

herbalite:
There are at least 2 bugs in jquery 3.4.0. Probably best to wait until 3.4.1
https://github.com/jquery/jquery/issues

JamoCA:
This is a pretty bad bug... Every advanced integration on my form worked perfectly (muliple data sources, ajax, etc) except for the "tag" function.  We hadn't rolled 3.4.0 out fully in production, so I was the only one encountering the issue. It was frustrating trying to troubleshoot and I finally figured I'd check the issues reported on GitHub.  Thanks for reporting this!

kevin-brown:
I can confirm that if I run the tests under 3.4.0, it throws a ton of errors. Unfortunately those errors appeared to have been introduced in 3.2/3.3 and make it a bit difficult to trace the exact issues at play.

Any chance someone can generate a quick jsbin which demonstrates the problems under 3.4.0?

neerone:
Same thing with focus logic. jQuery 3.4.1

fydelio:
Closing my other issue and moved it here. Here is a phenomenon I encountered using jQuery 3.4.0 and the latest select2 4.0.7

For my select2 box I have multiple and tagging enabled. Interestingly with this setup I can only create new tags with max 2 characters. Selecting an existing option with a longer text-value no problem. If I revert to jQuery. 3.3.1 the problem disappears!

Please see the JSFiddle [here](https://jsfiddle.net/rqyzupjv/2/)

Is this a bug or am I doing something wrong?

devkral:
is there a way to hack a workaround?

mariuszkrzaczkowski:
When can we expect the error to be fixed?

zero24x:
hello, jquery 3.3.1 has 1 medium level security vulnerability, please check out https://snyk.io/test/npm/jquery/3.3.1 

[...]

PR #5524, opened 2019-05-31
Fix compatibility with jquery-3.4.0

borisson:
See issue #5485

This pull request includes a

- [x] Bug fix
- [ ] New feature
- [ ] Translation

https://github.com/select2/select2/issues/5485

### DIFF
--- a/src/js/select2/selection/search.js
+++ b/src/js/select2/selection/search.js
@@ -178,7 +178,7 @@ define([
       var isTagInput = this.$element.find('[data-select2-tag]').length;
       if (isTagInput) {
         // fix IE11 bug where tag input lost focus
-        this.$element.focus();
+        this.$element[0].focus();
       } else {
         this.$search.focus();
       }


mkutlu:
This is not solved problem for IE, it just work for chrome.

blikblum:
> 
> 
> This was not solved problem for IE, it just work for chrome.

I tested with IE 11 and is working

kevin-brown:
Hi there, and thanks for the pull request to Select2.

Since this is fixing something which has already broken, we prefer it if the pull request includes a failing test case. With a failing test case, we don't need to worry about breaking this specific case again in the future, which saves us all a lot of time and trouble.

borisson:
@kevin-brown I'm sorry - I won't be able to include a testcase. We've since switched to a different library so I won't get company time and I maintain my own project(s) so most of my limited free time goes to that. I understand why you'd want a test - sorry I won't be able to provide that.

kevin-brown:
Should ultimately be fixed by https://github.com/select2/select2/pull/5558.

PR #5526, opened 2019-06-01
Update dev dependencies

blikblum:
This pull request includes a

- [ ] Bug fix
- [ ] New feature
- [ ] Translation
- [x] Dependency update

- Update most of the dev dependencies
- grunt-contrib-qunit is not updated to latest version because it conflicts with almond. The newest version compatible with almond loader is 0.5. This avoids to update grunt
- Update travis CI node version to 8 (oldest supported node)

This removes most of the build warnings / dependency vulnerabilities


kevin-brown:
> * grunt-contrib-qunit is not updated to latest version because it conflicts with almond

Do you happen to have a link to somewhere that states this? I believe you, I just didn't find anything when I did a quick search.

blikblum:
This is the affecting commit https://github.com/gruntjs/grunt-contrib-qunit/commit/db075a2b0a5e0f00390f39538cf6d18ff098673a

The select2 loader will try to load qunit module and fail

blikblum:
Not a problem of almond at all, but a conflict between how the qunit-grunt runs and select2 loader

kevin-brown:
This was merged in through https://github.com/select2/select2/pull/5529. I recompiled the dist files there and double checked that the resulting files are compatible (they are).

Issue #3860, opened 2015-10-22
Limiting the number of selections does not work using CTRL


FredericoSFerreira:
The maximum number of selection does not work when you use the CTRL key to perform the multiple selection

Decagon:
:+1:

What browser/OS have you tested this bug with? Since this is an older bug, does the newest version of select (4.0.1-rc.1 at the time of this writing) work?

kevin-brown:
You can reproduce this in 4.0.1-rc.1 by looking at the examples page.

https://select2.github.io/examples.html#multiple-max

This is because the maximum selection limit is enforced whenever the dropdown is opened, instead of when the selection is made.

urkle:
This issue still happens in the latest 4.x release (new URL to page showing the issue https://select2.org/selections#limiting-the-number-of-selections )

urkle:
This is still occurring in the latest code.

kevin-brown:
This is a duplicate of https://github.com/select2/select2/issues/3514.

Issue #3514, opened 2015-06-23
maximumSelectionLength is overridden by other configuration options


fkranenburg:
The property "maximumSelectionLength" does not work when "closeOnSelect" is set to false. I can select more items without any effective limit.

Check out this fiddle where both situations are shown.
http://jsfiddle.net/f9jszen1/

sparkhee93:
Is there a way to work around this bug?

jmiguel-localvox:
I'm also having this issue.

Candyffm:
Here is a workaround (depending on your exact needs). Enable closeOnSelect and let the select2 dropdown reopen automatically each time the selection has changed:
`
       var poiCatSelection = $('#poi_cats');
`

	poiCatSelection
		.select2({
			multiple               : true,
			closeOnSelect          : true,
			maximumSelectionLength : 3})
		.on('select2:selecting', function(){
			setTimeout(function() {
               poiCatSelection.select2('open');
    	    }, 0);
		});
`

grunmin:
@Candyffm not really a perfect solution, because container would flash and options' position  may change

alexweissman:
@mayashavin has identified that this is an issue with a few other configuration options as well.  From her issue (#4900):

Some of the following fields will override `maximumSelectionLength`:
1. `query`
2. `tokenSeperators`
3. `closeOnSelect` (set it to false with cause `maximumSelectionLength`) not working.

Libraries
- jQuery version:1.11.1
- Select2 version: 4.0.3

Browsers: Google Chrome
Operating System: Windows/Mac OS

Investigation: The dataAdapter prototypes get decorated with methods when maximumSelectionLength first, and with tokenSeperators later. Unfortunately, they both have same method name "query", and one will override the other in dataAdapter respectively.

Demo: https://jsfiddle.net/dpnminh/76shLmt2/

RedaMakhchan:
Having same issue fireFox 65

fkranenburg:
Yes. This issue is closed but not fixed.

RedaMakhchan:
Ok I found a workarround for this with little Js on select2.

We need to know that **maximumSelectionLength** is only not allowing user to open the dropdown, it doesn't limit the select inside dropdown if the dropdown kept opened with **closeOnSelect**.


So as a workarround I just added a listener on my select2 and dispatch a close once it attempts the value of limit :

I used this two options on my select 2 :
`
                closeOnSelect          : false,
                maximumSelectionLength: 3
`

[...]

PR #5533, opened 2019-06-05
Prevent too many items.

JacobLuckhardt:
This pull request includes a

- [x] Bug fix
- [ ] New feature
- [ ] Translation

- When selecting a new item, the eventRelay will trigger the maximumSelected message if the maximumSelectionLength is reached. This will prevent the user from adding more than maximumSelectionLength items, in case he is holding ctrl or closeOnSelect is false.

This will fix these issues:
#3514  
#3860


kevin-brown:
Fixed with tests in https://github.com/select2/select2/pull/5581.

Issue #5185, opened 2018-01-05
Select2 loses focus after selecting Item


PieterjanDeClippel:
Issue apparently already happened earlier: https://github.com/select2/select2/issues/4971.
However I haven't been bothered by it yet. All the time I've been using Select2, it's been working fine [v4.0.4 works fine](https://lyricdb.tk/song/37/edit).

Issue is available at the [very demo page](https://select2.org/getting-started/basic-usage#multi-select-boxes-pillbox). Just start typing something, and select a suggested item. Notice that the Select2 loses focus. My version is <b>v4.0.6-rc.1</b>

alexweissman:
I have confirmed that this problem is reintroduced in 4.0.6-rc.1.  See this [bin](http://jsbin.com/nuwuyiduru/edit?html,js,console,output).

@PieterjanDeClippel in the future, you need to create a [minimal, complete, and verifiable](https://stackoverflow.com/help/mcve) example using our [JSBin template](http://jsbin.com/hojajiwoye/edit?html,js,output) so that we can easily confirm these regressions.

alexweissman:
The only change I can think of that might have introduced this bug, is the addition of `blur` in [this commit](https://github.com/select2/select2/commit/ef2ef8b685c1ade1f5bf5cd691ce94a8b8e55437).  This was merged from #4680 without extensive testing.  Could you please confirm if this is the culprit?

tallon1252:
Hej :)
@alexweissman, the referenced commit seems not to be the issue.
I removed the line to test it out, but the behavior didn't change.

kzkn:
Probably [this commit](https://github.com/select2/select2/commit/45ee7340e54259156b3b0ed38786f6a730810587). See this [bin](http://jsbin.com/rimeqac/edit?html,js,console,output).

The jsbin has inlined select2 (4.0.6-rc1) code without `window.setTimeout(function() { self.$selection.focus(); }, 0);`, marked with `// REMOVE THIS`.

tallon1252:
@kzkn This fix doesn't seem to work for me…
But it works in your JSBin example. This one really starts to confuse me.

kzkn:
@tallon1252 The commit edits `dist/js/select2.js` directly, and use this from the jsbin. Perhaps, you built it yourself?

tallon1252:
@kzkn I just removed the lines you mentioned in directly in the select2.js…

kzkn:
[...]

PR #5552, opened 2019-07-01
Set focus to selection synchronously when closing

### LINKED ISSUE #5532: 4.0.5 -> 4.0.7 select2.close event regresssion
Upgrading from 4.0.5 to 4.0.7 and ran into an issue where this code no longer works:

```javascript
$(document).ready(function() {
  $("#select1").select2().select2("focus");
  $("select").on("select2:close", function (e) {
	 $("#checkbox1").focus();
	});
});
```

The goal is to auto-select the select2 on page load, then after the select2 is closed, focus on the next element. With 4.0.5 it works, with 4.0.7, the select2 element remains focused.

Example w/4.0.5 which works:
https://jsfiddle.net/bvL2wejz/

Example w/4.0.7 which doesn't work:
https://jsfiddle.net/4b0zy1hd/

blikblum:
This pull request includes a

- [x] Bug fix
- [ ] New feature
- [ ] Translation

- Set focus to selection synchronously in close event (remove setTimeout call)
- Call native focus method instead of jquery one because the later fails with jquery v1

Fixes #5532

### DIFF
--- a/src/js/select2/selection/base.js
+++ b/src/js/select2/selection/base.js
@@ -81,10 +81,8 @@ define([
       self.$selection.removeAttr('aria-activedescendant');
       self.$selection.removeAttr('aria-owns');
 
-      window.setTimeout(function () {
-        self.$selection.focus();
-      }, 0);
-    
+      self.$selection[0].focus();
+
       self._detachCloseHandler(container);
     });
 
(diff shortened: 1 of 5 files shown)


blikblum:
Probably fixes #5185 also

PR #5554, opened 2019-07-04
Remove unused variables

blikblum:
This pull request includes a

- [ ] Bug fix
- [ ] New feature
- [ ] Translation
- [x] Cleanup

- Remove some variables declared but not used
-
-

### DIFF
--- a/src/js/select2/selection/base.js
+++ b/src/js/select2/selection/base.js
@@ -38,7 +38,6 @@ define([
   BaseSelection.prototype.bind = function (container, $container) {
     var self = this;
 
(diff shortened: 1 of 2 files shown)


src/js/select2/selection/base.js
```diff
@@ -38,7 +38,6 @@ define([
   BaseSelection.prototype.bind = function (container, $container) {
```
kevin-j-morse: `$container` isn't used either....

kevin-j-morse: although maybe this is implemented somewhere else... so nvm

PR #5584, opened 2019-07-23
Check whether element is in document before calling jQuery.fn.offset()

kevin-j-morse:
This pull request includes a

- [X] Bug fix
- [ ] New feature
- [ ] Translation

**The following changes were made**

As of jQuery 3.0, jQuery.fn.offset() requires an element connected to a document. In earlier versions of jQuery, the .offset() method would return a value of `{ top: 0, left: 0 }` for some cases of invalid input. jQuery 3.0 throws errors in some of these cases. 

This PR checks whether the element is within the document. If not, it sets the offset to `{ top: 0, left: 0 }` which is what was happening before. I believe this covers all valid scenarios but perhaps someone else can comment?

See the jQuery migrate documentation for more info. https://github.com/jquery/jquery-migrate/blob/master/warnings.md#jqmigrate-jqueryfnoffset-requires-an-element-connected-to-a-document


kevin-brown:
I know this might be tough, since our positioning logic is not the easiest to read and test, but would it be possible for you to create a test for this?

It wouldn't be a whole lot different from

https://github.com/select2/select2/blob/2fce8ae6c4937b88c332ad3e30e135cf24504cf1/tests/dropdown/positioning-tests.js#L61-L119

Except you don't need to play around with the CSS and you don't need to attach it to the QUnit fixture (since that would add it into the document).

kevin-j-morse:
I will see what I can do.

kevin-j-morse:
@kevin-brown should this test live within `select2/tests/dropdown/positioning-tests.js`

kevin-brown:
> should this test live within `select2/tests/dropdown/positioning-tests.js`

That sounds like a good place for it, yes.

kevin-brown:
In the interest of getting 4.0.9 out the door without procrastinating even further, I'm moving this PR to land in 4.0.10.

kevin-j-morse:
Sorry. I have been swamped with work lately.

I will try to get to this soon.

PR #5588, opened 2019-07-27
Updated the qunit version to drop phantomjs

paxnovem:
Tests are currently being run with phantomjs which the maintainer
recomends switching to puppeteer and chrome

This pull request includes a

- [ ] Bug fix
- [ ] New feature
- [ ] Translation
- [x] Misc

- Upgrading grunt-qunit-contrib
- This converts us from phantomjs to headless chrome

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
     "grunt-contrib-concat": "^1.0.1",
     "grunt-contrib-connect": "^2.0.0",
     "grunt-contrib-jshint": "^1.1.0",
-    "grunt-contrib-qunit": "^1.3.0",
+    "grunt-contrib-qunit": "^3.1.0",
     "grunt-contrib-requirejs": "^1.0.0",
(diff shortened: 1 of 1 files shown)


kevin-brown:
One thing to note here is that there appears to be a bug in grunt-contrib-qunit where it is not consistently executing all of the tests. I have seen this locally, and you can observe it in the Travis CI output for this pull request.

https://travis-ci.org/select2/select2/builds/564490611

What's slightly worse is that I suspect it is executing the integration tests multiple times. There might be a change to the QUnit test harness that we use for integrating it which needs to be updated.

paxnovem:
I will look to see if there were changes to how tests were collected since our last version. It is likely there are changes since we are going up a couple major versions.

paxnovem:
Going to see if I can get phantomJS to  play well with running in a docker container (where I run my dev environment)

PR #5611, opened 2019-08-16
Add smooth scrolling to results on iOS Devices

bjoernwibben:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

This CSS fix only enables "smooth" scrolling on the results on iOS Devices.

From the Docs (https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-overflow-scrolling):

"Use momentum-based scrolling, where the content continues to scroll for a while after finishing the scroll gesture and removing your finger from the touchscreen. The speed and duration of the continued scrolling is proportional to how vigorous the scroll gesture was. "

It should be now easier to scroll long result lists on a iOS mobile Device.

Cheers,
Bjoern

### DIFF
--- a/src/scss/_dropdown.scss
+++ b/src/scss/_dropdown.scss
@@ -17,6 +17,7 @@
 }
 
 .select2-results {
+  -webkit-overflow-scrolling: touch;
   display: block;
 }
 


src/scss/_dropdown.scss
```diff
@@ -17,6 +17,7 @@
 }
 
 .select2-results {
```
andreivictor: Why adding to .select2-results?
Actually the .select2-results__options is the scrolled element (max-height: 200px; overflow-y: auto;)

Issue #5616, opened 2019-08-22
Select2 single choice - wrong clear title / tooltip


florianlacreuse:
Hi,

Since PR #4530, the clear icon to remove all the selected items in Select2 multiple choices has a title / tooltip. However, this new title is now also applied to the clear icon for Select2 single choice, whereas it makes much less sens to have 'Remove **all** item**s**' as a tooltip in the case of a single / unique choice component.

Reduce test case: https://codepen.io/florianlacreuse/pen/jONVogy

adrichey:
I am going to hop on this and get a PR in this weekend. :)

florianlacreuse:
There is a PR in progress. See #5671.

florianlacreuse:
There is a PR in progress. See #5671.

kevin-brown:
Cloning what I put in https://github.com/select2/select2/pull/5671#issuecomment-546722934

> Thank you for the pull request! I've been thinking about this one for a couple of days, trying to decide how to best handle this knowing that there are some changes in the pipeline for handling accessibility in future versions of Select2. And I'm honestly still thinking.
> 
> 1. We're looking to add title text for the individual "remove" buttons on selected items for a multiple select. This would also likely apply to the "clear" button for a single select, but we'll see when it gets implemented. The translation we would try to target is `removeItem` and we would be injecting the option text into the translated message (so it says "Remove Alaska" instead of "Remove item")
> 2. I'd also like to modify the `removeAllItems` translation so it either reads out the number of items it's removing or the individual titles. Ideally it would pass in the data as a parameter so people could override this on their own for greater flexibility.
> 3. Because the plan is to allow user-submitted data within title attributes, we need to move away from injecting the title attribute by string concatenation and move towards injecting it using `.setAttribute('title', value)`/`.title = value`, to avoid future XSS issues.

dmmishchenko:
Hi, what is the current status of this issue?

florianlacreuse:
Still open, a PR had been made by another developer but was not merged (closed). I still haven't had any feedback from maintainers.

(AFAIK)

PR #5671, opened 2019-10-20
Select2 single choice - wrong clear title / tooltip

adrichey:
This pull request includes a

- [ ] Bug fix
- [X] New feature
- [ ] Translation

- This changes the allowClear "Remove all items" tooltip to say "Remove item" when the `<select>` is a single select.
- Includes localization and tests.

- Addresses issue https://github.com/select2/select2/issues/5616.


kevin-brown:
Thank you for the pull request! I've been thinking about this one for a couple of days, trying to decide how to best handle this knowing that there are some changes in the pipeline for handling accessibility in future versions of Select2. And I'm honestly still thinking.

1. We're looking to add title text for the individual "remove" buttons on selected items for a multiple select. This would also likely apply to the "clear" button for a single select, but we'll see when it gets implemented. The translation we would try to target is `removeItem` and we would be injecting the option text into the translated message (so it says "Remove Alaska" instead of "Remove item")
2. I'd also like to modify the `removeAllItems` translation so it either reads out the number of items it's removing or the individual titles. Ideally it would pass in the data as a parameter so people could override this on their own for greater flexibility.
3. Because the plan is to allow user-submitted data within title attributes, we need to move away from injecting the title attribute by string concatenation and move towards injecting it using `.setAttribute('title', value)`/`.title = value`, to avoid future XSS issues.

PR #5690, opened 2019-10-31
Docs: Fix Selections templating example

swrobel:
This pull request includes a

- [x] Bug fix
- [ ] New feature
- [ ] Translation

The [docs page for selections](https://select2.org/selections) currently does not render the flags as intended in the example. This makes the code the same as is used on the [dropdown page](https://select2.org/dropdown), which does work
Screen Shot 2019-10-31 at 1 21 22 PM
Screen Shot 2019-10-31 at 1 21 05 PM


swrobel:
I now realize this was a misunderstanding of the difference between templateSelection & templateResult.
Screen Shot 2019-10-31 at 1 34 03 PM

Issue #5682, opened 2019-10-28
Select2 inside ShadowRoot Broke In 4.0.11


OJMichael:
#5584 breaks the usage of Select2 inside a ShadowRoot, because `$.contains()` returns false, even though the Node is actually connected to the document.

I'd propose changing the following code:
```javascript
    if ($.contains(document.body, $offsetParent[0])) {
      parentOffset = $offsetParent.offset();
    }
```

To this:
```javascript
    if ($offsetParent[0].isConnected) {
      parentOffset = $offsetParent.offset();
    }
```

lazka:
Since 4.0.11 (4.0.10 works) I also get wrong offsets for the popup:

![select2-bug](https://user-images.githubusercontent.com/991986/67678042-e9345b80-f985-11e9-8ce6-38c7bc1df5e4.png)

This is also inside a Shadow DOM.

kevin-brown:
Probably going to switch this to

```js
    if ($.contains(document.body, $offsetParent[0]) || $offsetParent[0].isConnected) {
      parentOffset = $offsetParent.offset();
    }
```

Just to handle IE/Edge/older browsers when they aren't using the shadow DOM (because let's be honest, they won't be using the shadow DOM if they're old enough to not support `isConnected`).

paxnovem:
I have made the change locally, looking into how to get proper test coverage for this

knlshnd:
I am still not able to get it working inside the shadowDOM. I am trying to use it inside an angular element. It gives me an error: Please add Select2 library (js file) to the project. I am using view encapsulation in this. I do not know how to use dropdownParent properly, so have not tried that

lazka:
I have it working with a dropdownParent referencing an element in the shadow dom, if that helps.

knlshnd:
Can you give me a reference as to how exactly to do it?

sergotail:
Hi, issue still persists for multiple shadow DOM subtrees having select2 elements. I have a custom element, and i initialize select2 for elements inside shadow DOM, in connectedCallback. And when i add second, third and so on custom elements, all select2 forms nested in existing custom elements are become broken.
My version of select2 is 4.0.13.

PR #5691, opened 2019-11-01
Fix support for the Shadow DOM when calculating offsets

paxnovem:
This pull request includes a

- [X] Bug fix
- [ ] New feature
- [ ] Translation

- Changed check to support if it is placed inside shadowroot
- 
-

https://github.com/select2/select2/issues/5682

### DIFF
--- a/src/js/select2/dropdown/attachBody.js
+++ b/src/js/select2/dropdown/attachBody.js
@@ -195,7 +195,7 @@ define([
       left: 0
     };
 
-    if ($.contains(document.body, $offsetParent[0])) {
+    if ($.contains(document.body, $offsetParent[0]) || $offsetParent[0].isConnected) {
       parentOffset = $offsetParent.offset();
     }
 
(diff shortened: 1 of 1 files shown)


kevin-brown:
Marking this "needs additional work" since we're going to need tests that verify the behaviour. But also marked "pending PR merge" since I'll do those if I need to.

The goal is to roll a new release on or around Tuesday of this week to fix this issue.

kevin-brown:
After playing around with writing tests for this I've come to the conclusion that it's not worth it to write tests, since they would require setting CSS on the `<html>` root element and that would be tough to clean up.

For context, calling `$.fn.offsetParent()` on an element held within the Shadow DOM returns the `<html>` root element. So the only way to get an offset on it is to adjust the CSS styling of it. Considering how rare this case is, since most people don't apply styling to the `<html>` element, it's not worth it to cover this edge case and I'd rather handle regression as-needed.

PR #5716, opened 2019-12-03
Fix show long text on tag mode

MrJavadAdib:
Hi! This pull request includes a
- [x] Bug fix

- fix on type new tag
- fix on show tag in list
- fix on show selected tag

### DIFF
--- a/src/scss/_dropdown.scss
+++ b/src/scss/_dropdown.scss
@@ -32,6 +32,11 @@
   user-select: none;
   -webkit-user-select: none;
 
+  max-width: 100%;
+  overflow: hidden;
+  text-overflow:ellipsis;
+  white-space:nowrap;
+
   &[aria-selected] {
     cursor: pointer;
   }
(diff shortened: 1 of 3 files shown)


kevin-brown:
I'm going to have to reject this.

We do this for single selects because they try as closely as possible to match the behaviour of a browser-renderer single select, which includes handling text wrapping. We intentionally leave the display of multiple selects as far as this goes to the user, since it tends to be an edge case as-is.

It's the dropdown text wrapping that was the actual cause for the rejection, since we definitely don't want to have `white-space: nowrap` within the results.

I would recommend placing these overrides in your own CSS files.

MrJavadAdib:
Okay!
i try to fix 3 bug. but you reject two of them with reason. as you wish.

Before fix
image

After fix
image

kevin-brown:
For more explanation on the issue with the results one.

Before:

image

After:

image

Forcing `white-space: nowrap` on results gives a worse user experience on long option text for the regular case (natural whitespace breaks).

----

When typing on a multiple select we already follow the browser conventions for handling text overflows

image

----

The remaining change to the display of long multiple select options would be reasonable to apply, though you don't need `white-space: nowrap` (since it doesn't wrap naturally) and it was not consistently applied across themes here.

image

PR #5723, opened 2019-12-09
Fix #5110

### LINKED ISSUE #5110: [Fixed] Bootstrap Modal with Select2
Hi I just find the way to fix the compatible between Bootstrap and Select2, when the select2 cannot present, cannot clickable and when it scroll get problem.

Like this problem: https://jsfiddle.net/eL9d2pdL/3/
You cannot click into it.

First you need to do the dropdownparent:
This gonna allow you to do the input in the select.
https://jsfiddle.net/eL9d2pdL/4/

However if the modal is too high then like this the select2 will run away:
https://jsfiddle.net/eL9d2pdL/5/

And how I fixed it:
https://jsfiddle.net/eL9d2pdL/6/

First you need to find: 

`container.bottom` in `AttachBody.prototype._positionDropdown`

Original:
Line 4274:
`container.bottom = offset.top + container.height`
Line 4319:
`css.top = container.top - parentOffset.top - dropdown.height
`
Changed to:
4274:
`container.bottom = offset.top + container.height + $($(this)[0].$dropdownParent[0]).scrollTop();`
4319:
`css.top = container.top - parentOffset.top - dropdown.height + $($(this)[0].$dropdownParent[0]).scrollTop();
`

The new code doesn't affect the normal one.
https://jsfiddle.net/rLmztr2d/276/

:) Hope you enjoy :)

tlebreton:
This pull request includes a

- [x] Bug fix
- [ ] New feature
- [ ] Translation

This pull request Fix a bug with bootstrap modal.
If we click on a select2 then scroll then reclick the dropdown doesn't update his position.
This is in relation with #5110

### DIFF
--- a/src/js/select2/dropdown/attachBody.js
+++ b/src/js/select2/dropdown/attachBody.js
@@ -162,7 +162,8 @@ define([
     };
 
     container.top = offset.top;
-    container.bottom = offset.top + container.height + $($(this)[0].$dropdownParent[0]).scrollTop();
+    container.bottom = offset.top + container.height + 
+      $($(this)[0].$dropdownParent[0]).scrollTop();
 
     var dropdown = {
       height: this.$dropdown.outerHeight(false)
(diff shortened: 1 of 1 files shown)


tlebreton:
@kevin-brown Why is my PR has been staled.

It's a confirm bug (see the issue) and this pr solved it.

kevin-brown:
I'm going to reopen this for another stale cycle, just to see if we can get this into a shape where it can be accepted.

In order for it to be fully accepted we are going to [need a test that demonstrates the original issue](https://github.com/select2/select2/blob/284f6bb02850eae283add5fcaf8dc576714f85f2/tests/dropdown/positioning-tests.js), so that we can ensure this issue does not come back up in the future. If you are not able to write a unit test then a jsbin that reproduces the exact issue would also work.

PR #5734, opened 2019-12-17
Only trigger 'selection:update' once on DOM change events

dbramwell:
This pull request includes a

- [X] Bug fix
- [ ] New feature
- [ ] Translation

- Rather than triggering DOM updates for every mutation within the select, only do it once if at least one mutation is detected.

I've had an issue when using select2 with datatables editor. I have a dropdown that has 4000+ elements, loaded via datatables editor and ajax. Everytime I filter/sort/change page on the table a new request fires to get the data the correct page, but also (annoyingly and seemingly unnecessarily) the list of options for the dropdown gets returned again... This then causes the page to become unresponsive as select2 goes through and refreshes the DOM (I think, it's certainly doing something along those lines) 4000+ times for each option being removed, and 4000+ times for each option being added.

This issue didn't exist before 4.0.3, and the below changes fix it. All tests are still passing, but I could very easily have overlooked something important though.

### DIFF
--- a/src/js/select2/core.js
+++ b/src/js/select2/core.js
@@ -208,8 +208,8 @@ define([
 
     if (observer != null) {
       this._observer = new observer(function (mutations) {
-        $.each(mutations, self._syncA);
-        $.each(mutations, self._syncS);
+        self._syncA();
+        self._syncS(null, mutations);
       });
       this._observer.observe(this.$element[0], {
         attributes: true,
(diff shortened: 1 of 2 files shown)


kevin-brown:
Is it possible to add a test that covers this change? It would probably be similar to the existing DOM integration tests, but for handling multiple changes.

dbramwell:
Happy to give it a try, yes. Should I just be testing that `selection:update` only happens once? Kind of tricky without making the test "wait" for a little bit to check it doesn't get called more than once.

dbramwell:
The test I've added fails before my change, but I'm not sure how useful the test case is... Let me know your thoughts

PR #5741, opened 2020-01-09
Fix Contribute link

VSPPedro:
This pull request includes a

- [x] Bug fix

Hi, the Contribute link returned for me Page Not Found. This is a fix for it.

Thank you all for this awesome library!

PS:  Another thing, this link is returning not found too: `https://opencollective.com/select2/contributors.svg?width=890&button=false`. It appears at the side of the Contribute link. I'm not sure how to solve this, so I'm just saying.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,7 +130,7 @@ The license is available within the repository in the [LICENSE][license] file.
 
 ### Code Contributors
 
-This project exists thanks to all the people who contribute. [[Contribute](CONTRIBUTING.md)].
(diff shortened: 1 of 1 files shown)


kevin-brown:
> PS: Another thing, this link is returning not found too:

I think that was a configuration bug within Open Collective, where we were pointing to the code organization instead of the repository. I've updated that now, so I'm going to give it some time before I reach out to figure out what is wrong there.

PR #5746, opened 2020-01-17
Update LICENSE.md for 2020

Amorymeltzer:
Last updated in 2017


kevin-brown:
The LICENSE file doesn't need to be updated every year. In fact, I would rather remove the ending year than keep updating it.

Copyright is granted when the work is first created. The end year there doesn't really matter.

PR #5752, opened 2020-01-25
Add new option for multiselects to remove element on backspace

siddharthvp:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

- Add a new boolean option `clearTermOnBackspace`, which if set true, for a multiselect menu, causes backspace key on an empty search input field to remove the last selected option rather than converting it to text.
- Added unit test for the above.

Solves #3354. PR was requested in https://github.com/select2/select2/issues/3354#issuecomment-422209357

### DIFF
--- a/src/js/select2/selection/search.js
+++ b/src/js/select2/selection/search.js
@@ -212,7 +212,7 @@ define([
       data: item
     });
 
-    this.$search.val(item.text);
+    this.$search.val(this.options.get('clearTermOnBackspace') ? '' : item.text);
     this.handleSearch();
   };
 
--- a/tests/selection/search-tests.js
+++ b/tests/selection/search-tests.js
@@ -90,6 +90,56 @@ test('backspace will set the search text', function (assert) {
   assert.equal($search.val(), 'One', 'The search text was set');
 });
 
+test('backspace will remove the search text too with clearTermOnBackspace',
+  function (assert) {
+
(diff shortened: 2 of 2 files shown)


siddharthvp:
Any plan on merging this @kevin-brown ?

Nick-Riggs:
👍  I would very nice like to see this merged as well.

kevin-brown:
As noted in https://github.com/select2/select2/issues/3354#issuecomment-624310154, I'm going to hold off on changing around this behaviour until there is anything resembling a consensus around how accessibility should be handled for removing elements in components like this. I understand that this is inconvenient and annoying for right now, but unfortunately accepting this PR is just going to cause more churn later down the line.

Mihai-github:
Hi, any update because I would really need this future now :)))

bobjase:
Another upvote. The current functionality is totally unexpected from a user standpoint and is very annoying

martinkrcho:
I would love to see this merged into the library as well. @kevin-brown , if this is configurable, it shouldn't cause any more church. Whoever is happy with the existing behaviour will see no change. People who want the change, will simply change how they configure the library.

The same goes for the accessibility aspect of this. You would be leaving the choice to the users of the library.

Would you re-consider adding this to the next version of library? Thanks

PR #5758, opened 2020-01-30
Including the original event on enter key press and passing it along to final triggers

joepetrakovich:
This is a continuation of the work done for #5513 by @6pac.

The original event is included if the user tabs off the dropdown, but not if the user uses the enter key to select the item.  This change adds the original event to the enter key press and then makes sure it is passed along to both the `close `and `select `triggers inside the the `results:select` handler.

The existing container `select` handler already includes the original event if it is present, so this is all that was needed to get that to work.

### DIFF
--- a/src/js/select2/core.js
+++ b/src/js/select2/core.js
@@ -335,7 +335,9 @@ define([
 
           evt.preventDefault();
         } else if (key === KEYS.ENTER) {
-          self.trigger('results:select', evt);
+          self.trigger('results:select', {
+            originalEvent: evt
+          });
 
           evt.preventDefault();
         } else if ((key === KEYS.SPACE && evt.ctrlKey)) {
--- a/src/js/select2/core.js
+++ b/src/js/select2/core.js
@@ -335,7 +335,7 @@ define([
 
           evt.preventDefault();
         } else if (key === KEYS.ENTER) {
-          self.trigger('results:select', {});
+          self.trigger('results:select', evt);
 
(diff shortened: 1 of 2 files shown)


6pac:
Cool. Since Slickgrid is open source, it might be easier to put up a public example for that. I just think it's good to have a concrete example to tease out the subtleties of navigation. I'll try to do that in the next week.

src/js/select2/core.js
```diff
@@ -335,7 +335,7 @@ define([
 
           evt.preventDefault();
         } else if (key === KEYS.ENTER) {
-          self.trigger('results:select', {});
+          self.trigger('results:select', evt);
```
kevin-brown: This should be wrapped as `originalEvent` similar to how we do it for other events.

joepetrakovich: Ah, okay.  I wrapped it like that in `results.js` but thought it would be better to be consistent here with the call to `select.close` 4 lines above this, and to be consistent with the other handlers in `results.js`  (like the `mouseup` handler on line 448).   

Otherwise the `results:select` handler will have two `originalEvent: **evt**.originalEvent` wrappings which are inconsistent with the existing wrapping in the `mouseup `handler.  Seems like trading one inconsistency for two others.

Happy to make the change as long as you're okay with that.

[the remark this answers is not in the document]
joepetrakovich: Ok cool.  That plan sounds great actually.

Issue #4820, opened 2017-03-06
TAB should select a result


jlrolin:
This was previously in issue #4578 and then closed and referred to #3472 but I do not see any code change behind it in the source.

When a user opens Select2 and TABs, can we have the result selected? TAB seems like it should select, not close.

axelosorio:
yes we have the same problem? I tried with the events but cannot make to work.

Copywright:
I figured out a workaround:

`
$(document).ready(function () {
    ...
    var $selects = $('select');
    if ($selects.length > 0) {
        for (var i = 0; i < $selects.length; i++) {
            $($selects[i]).parents('.form-group').attr('tabindex', '0');
        }
    }
});

`

jlrolin:
OR... you could just add TAB to the results.select like the previous issues stated would occur.

alexweissman:
This makes sense to me, since this is how a native Select element behaves.  It is unclear how this would translate to multi-select, however.

ahmedali5530:
Even it does not work in official demo at [https://select2.org/getting-started/basic-usage](https://select2.org/getting-started/basic-usage)

pedrofurtado:
We don't have immediate plans to provide this. We are focused now to fix UI bugs. But if you open a PR with unit tests, I will be glad to review and approve if everything is ok.

paolobroccardo:
> We don't have immediate plans to provide this. We are focused now to fix UI bugs. But if you open a PR with unit tests, I will be glad to review and approve if everything is ok.

Would you mind explaining why? It seems like a perfectly valid request, especially considering that normal selects work this way, and it shouldn't be a big fix to add.

jakeNiemiec:
Paolo, see: https://github.com/select2/select2/issues/5367

RLHawk1:
This seems like a critical problem and makes stock select2 unusable in my product. For data entry, people expect to be able to start typing, then hit tab to select the highlighted result.

paxnovem:
I am not sure why this was originally closed. I am going to ping @kevin-brown  to get his opinion.

kevin-brown:
I just confirmed that the behaviour of Chrome when you hit tab is to select the currently highlighted option and close. Right now we just close, but we do not select the currently highlighted option.

https://github.com/select2/select2/blob/48775d027c3e301b0bd2c4bfefbf1e1536e94cee/src/js/select2/core.js#L332-L334

Tab should behave more like the enter key rather than the esc key.

[...]

PR #5759, opened 2020-01-30
Changed tab key press to behave the same as enter key and select the highlighted option

joepetrakovich:
Changed tab key press to behave the same as enter key and select the highlighted option.  This is consistent with modern browser behavior for `<select>` elements.

#4820

### DIFF
--- a/src/js/select2/core.js
+++ b/src/js/select2/core.js
@@ -329,12 +329,11 @@ define([
       var key = evt.which;
 
       if (self.isOpen()) {
-        if (key === KEYS.ESC || key === KEYS.TAB ||
-            (key === KEYS.UP && evt.altKey)) {
+        if (key === KEYS.ESC || (key === KEYS.UP && evt.altKey)) {
           self.close(evt);
 
           evt.preventDefault();
-        } else if (key === KEYS.ENTER) {
+        } else if (key === KEYS.ENTER || key === KEYS.TAB) {
           self.trigger('results:select', {});
 
           evt.preventDefault();


joepetrakovich:
I just realized this may be a breaking change for a lot of people.  If they've gone this long using select2 with the assumption that TAB should _not_ select, changing it may mess their surrounding navigation code up.

One scenario I found is when the user tabs to the dropdown, I have the dropdown automatically open up so they can start typing, but when you call open, the first option is already highlighted, so if you hit TAB again (with the desire to skip the field and leave it blank), you can't do it because it will now select that first highlighted option.  You'd be forced to use the mouse to then clear it.

This reminds me of the other issues that mention the desire for a blank first option.

kevin-brown:
> If they've gone this long using select2 with the assumption that TAB should _not_ select, changing it may mess their surrounding navigation code up.

We have a minor release coming up soon which will feature some of these small-but-important somewhat-breaking changes so I'm comfortable knowing that.

> I have the dropdown automatically open up so they can start typing

Custom UX behaviours aren't our concern.

> You'd be forced to use the mouse to then clear it.

Good news! This is one of the places where Select2 is accessible through a keyboard.

https://github.com/select2/select2/blob/51be56a6910288ea1b98b4191141c8904e2aa5d6/src/js/select2/selection/allowClear.js#L87-L89

> you can't do it because it will now select that first highlighted option

A fun feature request which could help here is to provide the keyboard event data to the `select` event, so you could detect that case and prevent the selection.

BKSpurgeon:
> Changed tab key press to behave the same as enter key and select the highlighted option. This is consistent with modern browser behavior for `<select>` elements.
> 
> #4820

@joepetrakovich thx i appreciate the PR. question: is there a way of adding the minor code changes in your PR so that it overrides the existing functionality in `src/js/select2/core.js`? any pointers would be much appreciated.

joepetrakovich:
@BKSpurgeon if you use the non-minified version of the library you can just do a text search for some of the lines I changed (like `key === KEYS.ESC`) and make the same change yourself.  Is that what you are asking?

BKSpurgeon:
Hi @joepetrakovich appreciate the response.  Was wanting to know if there was a way to quickly override the original core functionality provided by select2: (i'm getting it from a yarn / npm package so presumably that package is not minified): if that is true, then is there a way to retain the original package and specifically override those 1 or 2 lines of code?

BKSpurgeon:
Hi @joepetrakovich appreciate the response.  Was wanting to know if there was a way to quickly override the original core functionality provided by select2: (i'm getting it from a yarn / npm package so presumably that package is not minified): if that is true, then is there a way to retain the original package and specifically override those 1 or 2 lines of code?

BKSpurgeon:
> @BKSpurgeon In the intermediate, I've been using using the events system:
> 
> ```js
> mySelect2Widget.on('select2:select', function (e) {
>   $(this).focus();
> });
> ```

@devkinetic thx - do you have the context / larger code snippet of where to place the code? Not sure if i'm doing the right thing, but i couldn't get the above to work.

devkinetic:
@BKSpurgeon Sorry, i don't check github as often as I should. To follow up from my example `mySelect2Widget` is selector for an element with `select2()` applied to it. 

```javascript
var mySelect2Widget  = $('.css-class-to-target');
mySelect2Widget.on('select2:select', function (e) {
        $(this).focus();
        setTimeout($.proxy(triggerSubmit, this.form), 500);
      });
```

kevin-brown:
Reopening this for another stale cycle before the next release. Need to double check on ARIA for this before reviewing it.

kevin-brown:
ARIA has zero guidance on this front so I'm going to default back to selectWoo and how the browser implements it. So this PR looks good and will land for the 4.1.0 release candidate.

AndrijaX:
I noticed that if you navigate through web form with TAB keys (ARIA behaviour) and you are landing on select2 (multiple items) you get stacked in select2. TAB doesn't allow you to get to the next form field. This shouldn't happen. I'm missing something?

PR #5775, opened 2020-02-16
JQuery "find" replaced with more efficient "querySelectorAll" in SelectAdapter

vyshkant:
This pull request includes a

- [x] Bug fix
- [x] New feature
- [ ] Translation

- in `SelectAdapter`'s `current` method replaced expensive call to `jQuery.find` with more efficient call to `DOMElement.querySelectorAll`

According to [W3C Recommendation 06 November 2018](https://www.w3.org/TR/selectors-3/#checked), 

> the :checked pseudo-class initially applies to such elements that have the HTML4 selected and checked attributes

which is applicable for our case.

Here is an example illustrating the problems of jQuery.find method: https://jsfiddle.net/vyshkant/nzr4kp8u/12/

**CAUTION**: an execution of the example might take about 2 minutes (which actually illustrates the degradation of jQuery.find with increasing the number of `option` tags.

In my personal case I have a middle-size db which includes about 20 000 of options and it is impossible to use Select2 because of the timing.

### DIFF
--- a/src/js/select2/data/select.js
+++ b/src/js/select2/data/select.js
@@ -14,15 +14,19 @@ define([
 
   SelectAdapter.prototype.current = function (callback) {
     var data = [];
-    var self = this;
 
-    this.$element.find(':selected').each(function () {
-      var $option = $(this);
+    var domElement = this.$element[0];
 
-      var option = self.item($option);
+    var selectedElements = domElement.querySelectorAll(':checked');
+
+    for (var index = 0; index < selectedElements.length; index++) {
+      var selectedElement = selectedElements[index];
+      var $option = $(selectedElement);
+
+      var option = this.item($option);
 
       data.push(option);
-    });
+    }
 
     callback(data);
   };
--- a/src/js/select2/data/select.js
+++ b/src/js/select2/data/select.js
@@ -13,20 +13,14 @@ define([
   Utils.Extend(SelectAdapter, BaseAdapter);
 
   SelectAdapter.prototype.current = function (callback) {
-    var data = [];
+    var self = this;
 
-    var domElement = this.$element[0];
-
-    var selectedElements = domElement.querySelectorAll(':checked');
-
-    for (var index = 0; index < selectedElements.length; index++) {
-      var selectedElement = selectedElements[index];
-      var $option = $(selectedElement);
-
-      var option = this.item($option);
-
-      data.push(option);
-    }
+    var data = Array.prototype.map.call(
+      this.$element[0].querySelectorAll(':checked'),
+      function (selectedElement) {
+        return self.item($(selectedElement));
+      }
+    );
 
     callback(data);
   };


vyshkant:
If you agree with my arguments and consider the PR appropriate, I'll gladly fix the tests.

kevin-brown:
This is something I'm willing to entertain, and the timing is great since we're about to start dropping legacy support for IE. The tests need to be fixed, as you've noted, but this looks like a solid start.

vyshkant:
@kevin-brown do you have any ideas why did the tests fail?

I have the same message on my local computer:
```
Running "qunit:all" (qunit) task
Testing http://localhost:9999/tests/integration-jq1.html F
>> PhantomJS timed out, possibly due to:
>> - QUnit is not loaded correctly.
>> - A missing QUnit start() call.
>> - Or, a misconfiguration of this task.
```

Meanwhile at the `develop` branch everything is OK, even locally.

Maybe I need to (re)generate some files? Could you please help me understanding your workflow?)

vyshkant:
And when I open QUnit in browser I can see tests are passed:
screenshot-localhost_9999-2020 02 24-02_43_20

src/js/select2/data/select.js
```diff
@@ -14,15 +14,23 @@ define([
...
+    if (this.$element.length > 0) {
+      var domElement = this.$element[0];
 
-      var option = self.item($option);
+      var selectedElements = domElement.querySelectorAll(':checked');
 
-      data.push(option);
-    });
+      for (var index in selectedElements) {
+        if (selectedElements.hasOwnProperty(index)) {
```
kevin-brown: I suspect this is allowing non-DOMElement properties of the `NodeList` to make their way through in tests which is causing the test failures we are seeing in CI. I would recommend falling back to a traditional index-based `for` loop which should be fine for this situation and should ensure that we only ever try to call `.item` on a DOM element.

vyshkant:
@kevin-brown you were absolutely right about iteration over `NodeList`, thanks a lot! Looks like the tests got fixed.

src/js/select2/data/select.js
```diff
@@ -14,15 +14,21 @@ define([
 
   SelectAdapter.prototype.current = function (callback) {
     var data = [];
-    var self = this;
 
-    this.$element.find(':selected').each(function () {
-      var $option = $(this);
+    if (this.$element.length > 0) {
```
kevin-brown: I thought `$element` should always be present here? Do we have any tests where this is not the case?

vyshkant: @kevin-brown basicly it was an equivalent of `$.find()` funcion call (it won't fail even if it actually doen't contain a single element). Also I don't see inside `SelectAdapter` any explicit check of whether `this.$element` contain any dom element or it is just an emply wrapper.

If you think this check is redundant, I can remove it.

vyshkant: @kevin-brown what do you think? Should I remove this check?

kevin-brown: Sorry, lost track of this PR. I think the check can be safely removed, since Select2 will not initialize itself unless an element is present. So you can assume that `this.$element` references a DOM element.

This can be verified in the Select2 jQuery adapter file where we call `each()` on the element set to ensure that Select2 is initialized once per element and also is only initialized when there are elements.

vyshkant:
@kevin-brown could you please take a look again? I've removed dist changes and modified `current` method of `SelectAdapter` (removed unnecessary check).

kevin-brown:
This looks good and it'll land in the next release (4.1.0) which is coming at some yet-to-be-determined date.

Right now the status checks are failing but that appears to be because of a GitHub Actions issue. I'll look into that later, but it shouldn't impact this.

Issue #5778, opened 2020-02-22
 data-* attribute for pre-selected values on data-data array


icaroscherma:
Select2 v4.0.5
`FEATURE REQUEST`

I'm currently using `data-data` attribute with a JSON array of objects but this data doesn't come with the pre-selected values, I got it elsewhere and I don't want to manually put on a `<script>` tag like `$('#mySelect').val(arrayOfSelectedValues);`, is there hidden feature to an attribute like `data-selected="[1,2,3]"` to run this `$(this).val($(this).data('selected'));` ?

Is it a wanted feature to somebody else? If so I will take a look to make a PR.

icaroscherma:
On `Options.prototype.fromElement = function ($e) {`, before `return this;` (line 5035 on v4.0.5), add the lines:

```Javascript
if ($e.data('selected')) {
  let selected = $e.data('selected');
  if (!Array.isArray(selected)) {
    selected = [selected];
  }
  selected.forEach(s => {
    const index = this.options.data.findIndex(x => x.id === s);
    if (index !== false) {
      this.options.data[index].selected = true;
    }
  });
}
```

Any negative effects that you guys are aware of?

kevin-brown:
Closing this off as wontfix, unfortunately we do not plan on adding another option to support this behaviour. Luckily this should be something you can do on your own with a little bit of code after you initialize Select2.

```js
$('.js-select2').select2();
$('.js-select2[data-selected]').each(function () {
  var selected = $(this).data('selected');
  if (!Array.isArray(selected)) {
    selected = [selected];
  }
  $(this).val(selected);
});
```

All of this works because Select2 automatically converts array data into `<option>` elements that can be treated like any other option in your select. Which means you can call `val()` on the select to set those as the expected options.

PR #5782, opened 2020-02-26
Add the option to pre-select in case of data array

icaroscherma:
You can use `data-selected` attribute with a single id (`"5"`) or an array of multiple ids (`"[1,3,5]"`), useful in case you're getting `data-data` straight from your database but your selected options are comming from a different location.

This pull request includes a __New feature__ .

- Before returning the scope of `Options.prototype.fromElement`, some`options.data` are manipulated to add `.selected = true`

This is related the existing ticket: #5778

### DIFF
--- a/src/js/select2/options.js
+++ b/src/js/select2/options.js
@@ -127,6 +127,19 @@ define([
       }
     }
 
+    if ($e.data('selected')) {
+      let selected = $e.data('selected');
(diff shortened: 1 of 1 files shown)


kevin-brown:
Closing this off per https://github.com/select2/select2/issues/5778#issuecomment-615999524.

PR #5811, opened 2020-03-25
Accounted for load more ajax errors

MTStevensW:
Fixed "Uncaught TypeError: Cannot read property 'length' of undefined" when the initial ajax load works fine but subsequent ajax loads of further results error, thus causing a further error in the rendering of results

This pull request includes a

- [X ] Bug fix
- [ ] New feature
- [ ] Translation

- Ajax error element added (once) when rendering more results with an error object
- Ajax error element removed when successfully rendering more results

### DIFF
--- a/select2.js
+++ b/select2.js
@@ -1832,6 +1832,15 @@ the specific language governing permissions and limitations under the Apache Lic
                     // ignore a response if the select2 has been closed before it was received
                     if (!self.opened()) return;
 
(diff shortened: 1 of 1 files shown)


kevin-brown:
Unfortunately we are no longer accepting pull requests to the 3.5.x line of Select2. It's ancient at this point and should not be used.

PR #5812, opened 2020-03-25
Update highlightFirstItem to take tags in to consideration

martincarlin87:
This pull request includes a

- [x] Bug fix
- [ ] New feature
- [ ] Translation

- Update `Results.prototype.highlightFirstItem` to check for unselected options and if the `tags` option is set. If so, highlight this first instead of a previously selected option.

This works at the moment when no tags exist, but not when there are existing selected tags.

__Fix__

image

__Current__

image

### DIFF
--- a/src/js/select2/data/tags.js
+++ b/src/js/select2/data/tags.js
@@ -78,7 +78,7 @@ define([
 
       if (tag != null) {
         var $option = self.option(tag);
-        $option.attr('data-select2-tag', true);
+        $option.attr('data-select2-tag', 'true');
 
         self.addOptions([$option]);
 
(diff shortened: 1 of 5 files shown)


kevin-brown:
I've been thinking about this one for the last 9 days and I think the solution I'm aiming to go for is instead of modifying `Result` to check for `tags`, is to add a new decorator that is injected when `tags` is enabled that overrides this method to provide this extra functionality.

martincarlin87:
No problem, sounds like a neater way of doing it 👍

kevin-brown:
Reopening as a reminder about the decorator plan since I don't see an attached ticket.

PR #5658, opened 2019-10-08
Fix select2#5378 ie11 specific focus issue with select2#5453

jbosse:
This pull request includes a

- [ X] Bug fix
- [ ] New feature
- [ ] Translation

The following changes were made

Resolved merge conflicts from https://github.com/select2/select2/pull/5453 developed by @vandernorth

There is bug with tags in IE11. This fixes it. Two previous Issues were closed without a fix. Please do not close unless merged.

PR #5821, opened 2020-04-01
Fix select2#5378 ie11 specific focus issue.

vandernorth:
This pull request includes a
- [X] Bug fix
- [ ] New feature
- [ ] Translation

 - Existing IE detection moved up a few lines so they are declared before they are used.
 - On the focusout event of the search-input use the existing variable to detect IE
 - If IE, and the event comes from the tag, refocuses so we can keep typing
 - Couldn't add an automatic IE11 test but added a html file where you can manually test this issue

Original issue #5378 
Original PR #5453 / second attempt PR #5658 
Probably fixes the same issue as was attempted by the removed code from https://github.com/select2/select2/commit/a83be86e9a2ecf0965ffa20965efe855783e0c24 
I think the removed code no longer fixes the issue and can stay removed.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -33,7 +33,7 @@ module.exports = function (grunt) {
     cwd: 'src/js'
   }, 'select2/i18n/*.js');
 
-  var testFiles = grunt.file.expand('tests/**/*.html');
+  var testFiles = grunt.file.expand('tests/**/!(manual)*.html');
   var testUrls = testFiles.map(function (filePath) {
     return 'http://localhost:9999/' + filePath;
   });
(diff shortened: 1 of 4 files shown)


vandernorth:
@ivaynberg / @kevin-brown any thoughts?

kevin-brown:
I was doing some testing of this issue and this appears to have been fixed as a side-effect of the changes made in https://github.com/select2/select2/pull/5842.

vandernorth:
@kevin-brown seems to be fixed in the 4.1.x-beta branch. I'll test again once it's stable. Thanks!

PR #5824, opened 2020-04-06
Add aria-label 'Search' to search input when dropdown opens.

adityatoshniwal:
This pull request includes a

- [ ] Bug fix
- [X] New feature
- [X] Translation

I've added an aria-label to the search box when dropdown opens. I've added a translation function for the label so that it can be set as per the language.


kevin-brown:
I'm going to evaluate the need for this soon as we are making some accessibility changes in the upcoming 4.1.0 release which would be a solid time for this pull request. Unfortunately it looks like this pull request includes changes to the `dist` directory which need to be reverted. We regenerate the `dist` periodically and including changes to that directory in pull requests causes the directory to quickly become out of sync with any code changes.

adityatoshniwal:
Got it. Generally, other packages required the dist files to be included in the pull.
I'll make the changes.

adityatoshniwal:
Oh wait. This got closed somehow. I messed up my fork and did hard reset it to this branch. I've committed the changes.

kevin-brown:
I'm pretty sure we need this for the dropdown search, and for consistency it probably makes sense to have it for the selection search as well (when the dropdown is open). The main changes that need to happen are:

* Set the attribute using `setAttribute` or `jQuery.attr()` (depending on context) instead of injecting it using string concatenation
  * This avoids potential XSS issues further down the line and is in line with how we do other translation injection
* Unit tests that ensure the attribute is being set properly in the right cases

adityatoshniwal:
Hey @kevin-brown, I've made the changes as mentioned. I've also added the test cases.
Please review.

PR #5831, opened 2020-04-18
Added i18n: Punjabi language

adsingh14:
This pull request includes a

- [x] Translation

- Added new i18n **Punjabi** language file, **pb.js**, in `src/js/select2/i18n/`


kevin-brown:
Hi there and thanks for contributing this translation to Select2. Unfortunately it looks like the `pb` language code was used for this translation, which doesn't align with the ISO 639 language code of `pa` which was assigned to the Punjabi language. If the file could be renamed to `pa.js` that would make it easier for us when we come around to merging this in for the upcoming release.

PR #5841, opened 2020-04-26
Mark highlighted results with aria-selected

### LINKED ISSUE #4349: Safari and VoiceOver, and the usage of aria-selected

kevin-brown:
This pull request includes a

- [x] Bug fix
- [ ] New feature
- [ ] Translation

- `select2-results__option--selectable` class was introduced to mark options that are selectable within results
- `select2-results__option--selected` class was introduced to mark options that are selected
- `select2-results__option--disabled` class was introduced to mark options that are disabled within results
- `[aria-selected]` is used to indicate the highlighted option instead of indicating that an option is selected

Fixes #4349

### DIFF
--- a/src/js/select2/results.js
+++ b/src/js/select2/results.js
@@ -174,8 +174,7 @@ define([
     option.classList.add('select2-results__option--selectable');
 
     var attrs = {
-      'role': 'option',
-      'aria-selected': 'false'
+      'role': 'option'
     };
 
     var matches = window.Element.prototype.matches ||
@@ -184,7 +183,6 @@ define([
 
     if ((data.element != null && matches.call(data.element, ':disabled')) ||
         (data.element == null && data.disabled)) {
-      delete attrs['aria-selected'];
       attrs['aria-disabled'] = 'true';
 
       option.classList.remove('select2-results__option--selectable');
@@ -192,7 +190,6 @@ define([
     }
 
     if (data.id == null) {
-      delete attrs['aria-selected'];
       option.classList.remove('select2-results__option--selectable');
     }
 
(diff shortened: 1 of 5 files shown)


manueltomasir:
@kevin-brown - I've updated the version of Select2 that i'm using to include your changes for accessibility and screen readers (many of my users are disabled)...After the update, it doesn't seem like keyboard navigation of the dropdowns works at all...I can open it using the spacebar but am unable to use the arrow keys to go through the options...Are you able to help me on this? Or am I just too early on updating to your beta and you already have this on your radar?

PR #5842, opened 2020-04-26
Make the multi-select selection area more accessible

### LINKED ISSUE #4418: Input with items already selected reads blank when focused with NVDA

### LINKED ISSUE #5585: Placeholder not showing
The placeholder text is not showing when the select element is not initially visible AND it is a "multiple" type select AND using jquery 3.4.1

See the fiddle for example. 

Click the button to show the select2 to reproduce.
https://jsfiddle.net/tmw4z7eb/1/

kevin-brown:
This pull request includes a

- [x] Bug fix
- [ ] New feature
- [ ] Translation

- Switch from floating the selections to using inline-block elements
- Move the search field outside of the selection list
- Move the clear button outside of the selection list
- Made the clear button use the `<button>` tag
- Multiple select search is always full width when the placeholder is present

Fixes #4418
Fixes #5585
Closes #5571

### DIFF
--- a/src/js/select2/i18n/en.js
+++ b/src/js/select2/i18n/en.js
@@ -42,6 +42,9 @@ define(function () {
     },
     removeAllItems: function () {
       return 'Remove all items';
+    },
+    removeItem: function () {
+      return 'Remove item';
     }
   };
 });
--- a/src/js/select2/selection/allowClear.js
+++ b/src/js/select2/selection/allowClear.js
@@ -102,7 +102,7 @@ define([
     var removeAll = this.options.get('translations').get('removeAllItems');
 
     var $remove = $(
-      '<button class="select2-selection__clear" tabindex="-1">' +
+      '<button type="button" class="select2-selection__clear" tabindex="-1">' +
         '&times;' +
       '</button>'
     );
--- a/src/js/select2/selection/allowClear.js
+++ b/src/js/select2/selection/allowClear.js
@@ -106,7 +106,7 @@ define([
 
     var $remove = $(
       '<button type="button" class="select2-selection__clear" tabindex="-1">' +
-        '&times;' +
+        '<span aria-hidden="true">&times;</span>' +
       '</button>'
     );
     $remove.attr('title', removeAll());
--- a/src/js/select2/selection/allowClear.js
+++ b/src/js/select2/selection/allowClear.js
@@ -92,6 +92,8 @@ define([
   AllowClear.prototype.update = function (decorated, data) {
     decorated.call(this, data);
 
+    this.$selection.find('.select2-selection__clear').remove();
+
     if (this.$selection.find('.select2-selection__placeholder').length > 0 ||
         data.length === 0) {
       return;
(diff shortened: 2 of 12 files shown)


kevin-brown:
There are still a few things left to do in this pull request

- [ ] Surround the `&times;` in remove buttons with an `aria-hidden="true"` element (probably a `<span>`)
- [ ] Add a new translation for "Remove item" which can be used on the individual "remove" buttons for choices

Hosch250:
Sorry for being a bit late with this, but I propose something like the following structure for the remove-item button:

```
<button type="button">
    <span aria-hidden="true">x</span>
    <span class="sr-only">Remove {visible label}</span>  <!-- bootstrap class; you'll want to implement your version that's basically a copy/paste implementation -->
</button>
```

This allows the user to access the button with keyboard focus, and correctly announces which item they are going to delete. It also hides the 'x' glyph you are using for the visual, which gets announced weird in at least some common screen readers. I have used this (or a similar) format myself successfully in a project using the powerful templating support you have; the only thing I couldn't fix using that was the announcement of "blank" when keyboarding/hovering through the dropdown of selectable items.

kevin-brown:
I believe the current generated HTML for the remove button is (unrelated attributes not included)

```html
<button type="button" aria-label="Remove item" title="Remove item" aria-describedby="id-of-item-text">
  <span aria-hidden="true">x</span>
</button>
```

The justification for doing this being that the "x" doesn't need to be made visible to screen readers (since, as you said, it gets announced weird in screen readers). As a replacement, the `aria-label` and `aria-describedby` are used with the goal of having it announced as "Remove item {visible label}" when focused by a screen reader.

Is there any chance you can confirm that this is what is read to screen readers using Select2 4.1.0-beta.1?

Hosch250:
I just tested this, and the only problem is the `tabindex="-1"` prevents it from being focused, so the user can't remove an item with the keyboard. If I remove that attribute, the selected item(s) delete buttons get selected before the control.

It also doesn't announce the label tied to the multi-select field I connected it to; instead, it says something like "item has autocomplete remove item test blank". Otherwise, everything works really well.

kevin-brown:
@Hosch250 while I've got those noted down on my side as things to fix in the next beta release, is there any chance you can make a ticket or two for capturing those bugs? That would make it easier to track the issues and know when they have been resolved.

Thank you for testing this for us!

Hosch250:
Done!

nishantyadav151:
@kevin-brown and @Hosch250 is this resolve?
Issue: announcment of "Blank" by NVDA while focussing on the selected value.
Steps to reproduce:
The issue is in with the select box
Steps to reproduce:
1.Turn on NVDA.
2. Go to the selection box.
3. tab over to the state field(the State field is selctable) and tab out of it.
4. tab back to state field and it reads "State open combo box editable blank"

 the expected behavior is not to say blank in the end.

blank items

PR #5859, opened 2020-05-17
Telugu (te) language translation

AtRaoN:
Created Telugu language translation.

This pull request is Translation only.

- Created Telugu Messages.


kevin-brown:
The changes here look good except the English translation file should not be deleted. That is what is causing the build failures that were automatically reported.

PR #5873, opened 2020-06-02
Added aria-labels to the Clear and Remove Selection buttons

matthewgeer:
This pull request includes a

- [X] Bug fix
- [ ] New feature
- [ ] Translation

Added aria-label attributes to the Clear and Remove Selection buttons in single and multi select, respectively.


ivaynberg:
applied in 44faf2b5, thanks!

PR #5879, opened 2020-06-11
Respect disabled attribute inherited from containing fieldsets

### LINKED ISSUE #2030: Fieldset disabled no longer disables Select2
Field set disabled no longer disables select 2. This worked in the prior version 3.2.

fire-birdie:
This pull request includes a

- [x] Bug fix
- [ ] New feature
- [ ] Translation

- Sets up a `MutationObserver` to monitor the `disabled` attribute of containing `frameset` elements (A second observer was used to avoid calling `_syncSubtree`, which shouldn't be necessary for the purposes of this fix)
- Extends the `isDisabled` abstraction to use `$element.is(':disabled')` in order to detect the inherited state

Resolves #2030

### DIFF
--- a/src/js/select2/core.js
+++ b/src/js/select2/core.js
@@ -353,7 +353,8 @@ define([
   };
 
   Select2.prototype._syncAttributes = function () {
-    this.options.set('disabled', this.$element.prop('disabled') || Utils.hasDisabledAncestors(this.$element));
+    this.options.set('disabled', this.$element.prop('disabled') ||
+      Utils.hasDisabledAncestors(this.$element));
 
     if (this.isDisabled()) {
       if (this.isOpen()) {
--- a/src/js/select2/core.js
+++ b/src/js/select2/core.js
@@ -353,7 +353,7 @@ define([
   };
 
   Select2.prototype._syncAttributes = function () {
-    this.options.set('disabled', this.$element.prop('disabled') || this.$element.is(':disabled'));
+    this.options.set('disabled', this.$element.prop('disabled') || Utils.hasDisabledAncestors(this.$element));
 
     if (this.isDisabled()) {
       if (this.isOpen()) {
--- a/src/js/select2/core.js
+++ b/src/js/select2/core.js
@@ -219,7 +219,7 @@ define([
 
         self._ancestorObserver.observe(this, {
           attributeFilter: ['disabled'],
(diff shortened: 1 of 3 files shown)


src/js/select2/core.js
```diff
@@ -473,7 +490,7 @@ define([
    * @return {false} if the disabled option is false.
    */
   Select2.prototype.isDisabled = function () {
-    return this.options.get('disabled');
+    return this.options.get('disabled') || this.$element.is(':disabled');
```
kevin-brown, line 493: Two things here.

1. The expected behaviour is to sync the "disabled" option with the state of the element. We do that already but we only check the `disabled` property.

  https://github.com/select2/select2/blob/ef33f4be110f5154c11b49d73e7b15921e143398/src/js/select2/core.js#L339

2. We do not currently use the `:disabled` jQuery selector because it relies on the Sizzle engine. If you are able to replace the `.is()` call with a vanilla JS call then we are more likely to be able to work that fix into Select2.

src/js/select2/core.js
```diff
@@ -206,6 +206,23 @@ define([
       childList: true,
       subtree: false
     });
+
+    var ancestors = this.$element.parents('fieldset');
```
kevin-brown, line 210: Only concern here is that we only try to figure this out _once_, which works for 90% of use cases but fails for a common one: dynamically added elements. It's common for Select2 to be initialized on an element before that element is added to the DOM.

I'm not sure of a great way for us to detect that and properly apply the observer.

fire-birdie: I actually considered that while creating the original patch, but it was 5am and the only solution I could think of was "attach an observer to the document root and monitor subtree changes", but that would obviously have enormous performance implications.

I'll give it some more thought, but honestly every moment feels like 5am these days...

kevin-brown: > I'll give it some more thought, but honestly every moment feels like 5am these days...

I 100% feel that. Not expecting any urgency on a fix for this or for you to be the one to provide it. We're all in this together.

---

I think this can be done with another set of mutation observers, but I would need to look more into it. It's tricky because most of the docs on things like this are for how to detect nodes being inserted into the current element, not for how to detect when the current element is inserted into the DOM. I'm sure we'll find an answer other there.

PR #5942, opened 2020-10-26
Add aria-controls attribute for combobox to fix accesibility issue

4rturj:
This pull request includes a

- [x] Bug fix

- aria-controls attribute added for combobox 

### DIFF
--- a/src/js/select2/selection/single.js
+++ b/src/js/select2/selection/single.js
@@ -37,6 +37,7 @@ define([
       .attr('role', 'textbox')
       .attr('aria-readonly', 'true');
     this.$selection.attr('aria-labelledby', id);
+    this.$selection.attr('aria-controls', id);
 
     this.$selection.on('mousedown', function (evt) {
       // Only respond to left clicks


eckmo:
I believe that this fix _does_ remove the "Element with role attribute has required states and properties" flag in a11y checkers, but should the id associated with `aria-controls` on the `combobox` instead be attached tot he corresponding `listbox` element? [This MDN doc](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/combobox_role#:~:text=When%20a%20combobox%27s%20popup%20is%20displayed%2C%20ensure%20the%20aria%2Dcontrols%20attribute%20on%20the%20combobox%20element%20is%20set%20to%20the%20id%20of%20the%20popup%20listbox%2C%20tree%2C%20grid%2C%20or%20dialog%20element.%20This%20is%20how%20the%20relationship%20between%20the%20element%20with%20the%20combobox%20role%20and%20the%20popup%20it%20controls%20is%20indicated.) seems to suggest that's the correct use of `aria-controls` in the case of `combobox`. So in the case of this `combobox`, the id associated with the `aria-controls` would be `select2-edit-field-topic-target-id-results` instead of `select2-edit-field-topic-target-id-container`.

PR #5961, opened 2020-12-28
show/hide select2

UziTech:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

- `$el.select2("hide")` will hide the select2
- `$el.select2("show")` will show the select2 when hidden

### DIFF
--- a/src/js/select2/core.js
+++ b/src/js/select2/core.js
@@ -553,6 +553,14 @@ define([
     this.$element.val(newVal).trigger('input').trigger('change');
   };
 
+  Select2.prototype.hide = function () {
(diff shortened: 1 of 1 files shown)


UziTech:
Should we also hide `this.$dropdown` on `hide`?

kevin-brown:
After thinking this feature request over for a week, we have decided to not accept it. It does not align with the feature set we are looking to bring to Select2 currently, but we recognize that it can be achieved with small modifications to the code.

UziTech:
What modifications?

PR #5981, opened 2021-02-11
Fix error callback with custom transport

KrzysztofBogdan:
If custom transport does not return any object 
then calling error callback would result in `Uncaught TypeError: right-hand side of 'in' should be an object, got undefined`.
This patch adds check if $request evaluates to something.

This pull request includes a

- [X] Bug fix

- Check if transport method returned data

### DIFF
--- a/src/js/select2/data/ajax.js
+++ b/src/js/select2/data/ajax.js
@@ -82,7 +82,7 @@ define([
       }, function () {
         // Attempt to detect if a request was aborted
         // Only works if the transport exposes a status property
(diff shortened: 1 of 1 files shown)


KrzysztofBogdan:
As documentation example shows in
https://select2.org/data-sources/ajax

This should work:
```
  ajax: {
    transport: function (params, success, failure) {
      var request = new AjaxRequest(params.url, params);
      request.on('success', success);
      request.on('failure', failure);
    }
 ```

but there will be error while calling `failure` callback

PR #5983, opened 2021-02-18
Add i18n extend addendum

job3dot5:
Documentation addendum how to extend a translation

### DIFF
--- a/docs/pages/13.i18n/docs.md
+++ b/docs/pages/13.i18n/docs.md
@@ -57,6 +57,23 @@ language: {
 
 >>> Translations are handled by the `select2/translation` module.
 
+### Extend a default translation
+```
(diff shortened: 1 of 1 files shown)


kevin-brown:
Is there any chance you can make a ticket about how to properly extend translations? This is definitely not the recommended way.

The recommended way involves relying on the [`language` fallback chain](https://github.com/select2/select2/pull/5602) to define your translation overrides instead of modifying the internals of Select2.

```js
var languageFr = [
  {
    noResults: function() { // Custom overrides for translation keys
      return 'Pas de bras pas de chocolat';
    }
  },
  'fr' // Fallback language for anything not overridden
];

$('select').select2({}
  language: languageFr
});
```

PR #6010, opened 2021-03-29
Fix Tokenizer error when an unknown option value is typed in, when `tags` are not enabled

### LINKED ISSUE #4747: "TypeError: g._removeOldTags is not a function" when deleting selections with backspace in multi-select
I am using the latest version of select2 from this repo.
I have a simple situation with a field which gets its data using ajax from a php page. Almost identical to the example "Loading remote data".
tags option is set to false and needs to be set to false in my situation.
multiple: true
firefox 50

Steps:
1. select an option from the list
2. use backspace to delete the option
3. get the error in console: TypeError: g._removeOldTags is not a function while deleting the option character by character with backspace

Nothing breaks, everything continues to work as it should, but it would be nice to avoid such errors. They may be an indication of more profound code problems. 

Thanks!

devmonkey22:
Fix Tokenizer error (`TypeError: g._removeOldTags is not a function`) when an unknown option value is typed in, when `tags` are not enabled.  Unknown options are ignored in this case.

This pull request includes a

- [X] Bug fix
- [ ] New feature
- [ ] Translation

- Update `Tokenizer` to check if `tags` are enabled before trying to create tags via `Tags` decorator methods.
- Add test to reproduce/validate fix

Resolves https://github.com/select2/select2/issues/4747

### DIFF
--- a/src/js/select2/data/tokenizer.js
+++ b/src/js/select2/data/tokenizer.js
@@ -33,11 +33,17 @@ define([
 
       // If an existing option wasn't found for it, create the option
       if (!$existingOptions.length) {
-        var $option = self.option(item);
-        $option.attr('data-select2-tag', true);
+        // If tags, allow creating new options
+        if (self.options.get('tags')) {
+          var $option = self.option(item);
+          $option.attr('data-select2-tag', true);
 
-        self._removeOldTags();
-        self.addOptions([$option]);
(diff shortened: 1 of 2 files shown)


thenitai:
Just wanted to say "Thank you" as this solved the issue

jonahgreenthal:

PR #6014, opened 2021-04-01
Add `SelectionOrder` DataAdapter decorator and related options

### LINKED ISSUE #3106: In multi-select, selections do not appear in the order in which they were selected
On this page: https://select2.github.io/examples.html

...All the tags appended to the end show up in an arbitrary place in the Select2 box. There seems to be no rhyme or reason where they will appear, neither alphabetical nor sorting by list matching. It just randomly sorts each time. This seems to be the reverse of issue #2283, so I'm not sure how to actually force the tags to append.

devmonkey22:
Add `SelectionOrder` DataAdapter decorator and related options as a proposal to resolve https://github.com/select2/select2/issues/3106 with some limitations for manual selection changes (as noted).

- This class has to keep both display and selection ordering for all options in order to manage selection order, while maintaining the default results/query ordering (when `sorter` is not used, etc).
- Notable but unfortunate limitation: Setting the control value using `jQuery.val([...])` with multiple items or manually toggling option `selected` (without a `change` event) still cannot preserve the value ordering.
  - The partial workaround for this is to manually set your selected options with the `data-selection-order` attribute before setting val() if possible.
  - It may also be possible to enhance the decorator to watch for manual changes via MutationObserver. TBD.

New options:
- `keepSelectionOrder`
- `trackManualSelectionOrder`

Update docs.
Add tests.

This pull request includes a

- [X] Bug fix
- [X] New feature
- [ ] Translation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,9 @@
 * Selected results in the dropdown should now be properly announced to screen readers (#5841)
 * Significant improvements were made to make the selection area accessible (#5824, #5842, #5916, #5942, #5973)
 * Allow pasting multiple lines into the search field for tokenization (#5806)
+* Added optional `SelectionOrder` DataAdapter decorator to support tracking value ordering for multiple selects (#3106)
+  * Added `keepSelectionOrder` option
+  * Added `trackManualSelectionOrder` option
 
 ### Bug fixes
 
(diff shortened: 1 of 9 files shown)


devmonkey22:
So I played with MutationObserver a little more. Unfortunately it doesn't fire for property changes (`option.selected` etc), so there is still no consistent way to watch for `jQuery.val` sets (without modifying `jQuery.valHooks` which I think is pretty discouraged.  So this is still a limitation of the solution without a different workaround like:

- Setting option selections individually with a change event each like https://github.com/select2/select2/blob/5fbd45fc10aacd26c25ffce8b1e0219d51f30036/tests/data/selectionOrder-tests.js#L555
- Manually setting the `data-selection-order` attribute of each option before setting `jQuery.val([...])`

devmonkey22:
I figured I'd throw this out here. This is a 6 year old problem with an imperfect solution, but I don't think there ever will be (without MutationObserver changes perhaps) given the constraints of using a real `<select>` and how HTML/JS behaves natively.

PR #6022, opened 2021-04-14
Clear search field after selection (w/ CloseOnSelect: false)

k00ka:
This pull request includes a

- [ x] Bug fix
- [ ] New feature
- [ ] Translation

- On results:select, and with tagging and closeOnSelect: false, the search field is not cleared. The "old" search text remains and the tag is created, resulting in the attached image.
- Solution is to clear the search field on results:select and not rely on the 'trigger' action to clear it.
- There is a question of whether the test is actually required. Without it, the search text may be cleared twice, but that doesn't seem like an issue.

Screen Shot 2021-04-14 at 10 32 50 AM


k00ka:
[Commit fe75f28] Clear selection also when *clicking* the item (formerly: only on enter keypress)

edegaudenzi:
Not sure why this was closed, issue still persists in 2025.
As a workaround this little snippet works ([source](https://stackoverflow.com/questions/27961256/clear-search-term-and-reload-all-items-of-select2-input-after-selecting-any-item)):
```
$(test).change(function() {
  $(this).parent().find('.select2-search__field').val("");
})
```
Tested in Select2 v4.0.13 (shipped with [Select2 Yii2 Extension](https://demos.krajee.com/widget-details/select2))

PR #6051, opened 2021-06-15
Completely hide the clear button on single select when allowClear = false

matthewgeer:
This pull request includes a

- [x] Bug fix
- [ ] New feature
- [ ] Translation

Applied the display: none CSS rule to the clear button on the single select control when allowClear setting is false.  This removes it from the tab order and prevents screen readers from reading any associated labeling.


control/src/select25.scss
```diff
@@ -293,6 +293,10 @@ $dropdown-padding: 8px 10px;
         width: 30px;
     }
 
+    &.s25-offscreen {
+        display: none;
```
ivaynberg: I dont think this is correct, it overloads the meaning of offscreen to mean hidden. I think the code should add the correct s25-hidden class when the control needs to be hidden.

PR #6068, opened 2021-07-10
Update docs.md

chris-kent:
Using select for an ajax call just doesn't work, looks like it needs to be an input element to work.
Many users are getting this error: Option 'ajax' is not allowed for Select2 when attached to a select element

This pull request includes a

- [ ] Bug fix
- [ ] New feature
- [ ] Translation

-
-
-


kevin-brown:
I'm very confident you're using a version of Select2 that is at least 6 years old, if not more. Please upgrade.

PR #6071, opened 2021-07-24
Set width for hidden placeholder text

hexmode:
A hidden field has no width

See https://phabricator.wikimedia.org/T284176
See #3817


jonahgreenthal:

PR #6076, opened 2021-08-09
Create .gitattributes

pzzd:
This pull request includes a .gitattributes file to exclude tests, src, and other directories from the archive files available under tags.

I'm following the advice at https://stackoverflow.com/questions/32475266/how-to-remove-all-tests-in-composer-php:

> It's worth noting that the Composer maintainers stopped promoting .gitattributes as a way to exclude stuff from packages. That option still exists due to the way Composer usually fetches distribution packages from Git repos, but it has been labeled to be "out of scope" for Composer itself. – Sven Jul 6 '16 at 19:43

The reason for this change is that the docs directory includes PHP files. So when this package is installed in a PHP site, there are unwanted PHP files in the web root.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,4 @@
+/.github export-ignore
+/docs export-ignore
+/src export-ignore
+/tests export-ignore


kevin-brown:
Based on what I've read, this will have the side effect of also removing these files from the zipped archives generated by our mirrors which is the opposite of what we want. The benefits to the PHP community do not outweigh that drawback unfortunately.

kevin-brown:
On a side note, we're thinking about removing those PHP files anyway which is another solution here.

pzzd:
Removing the PHP files is a good solution. Thanks for your review.

PR #6079, opened 2021-08-16
Adding "originalEvent" to "close" trigger args.

AlphaSolutions-jbj:


jonahgreenthal:

kevin-brown (approved):

PR #6082, opened 2021-08-19
Add language uz

ghost:
This pull request includes a

- [x] Translation

Uzbek translation was added to src/js/select2/i18n


jonahgreenthal:
I don't see any reason to close this (though apparently it fails lint, so that should be fixed).

PR #6086, opened 2021-09-03
Update docs.md

JoanneFsadni1998:
This pull request includes a

- [x] New feature

Live Search Section in Searching


JoanneFsadni1998:
This has been a headache to do on my own... i think it should be added in the select 2 docs

JoanneFsadni1998 (commented):
Live search with Ajax

jonahgreenthal:

PR #6096, opened 2021-09-17
Add feature that allows reuse of options

### LINKED ISSUE #4472: Reuse items in a multi-select (multiple) box
Similar to Issue #1301

I need the multiple select list (the side by side with boxes) to allow for the same value multiple times

for example

list would be Apple, Orange, Pear

I want to select 3 Oranges and 1 Apple

example  Orange, Orange, Orange, Apple

Issue #1301 says just create them in the <option>tags, but I don't want a long list of options

another example would be Red,Green,Blue

and for a security system my code might be Red,Red,Blue,Red so I need to enter it in this way to pass security, so no auto sort if this is possible.

sphuber:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

- Add the possibility to reuse existing tags for select widgets that allow multiple tags. When clicking on a tag from the selection container that is already selected, instead of deselecting it (as is the current behavior) the option is cloned and selected as well. Deselecting tags can now only be done by directly clicking on the cross mark of the tag which will also remove the corresponding option from the selection container.

Fixes #4472 

This is my first contribution to `select2` and I am not a javascript programmer in general. It is quite possible that this implementation is therefore far from ideal. I haven't added unit tests yet because I am not sure how, but testing this manually seems to accomplish what I needed for this feature. Happy to add tests once I know how and add documentation if the feature is approved and finalized.

### DIFF
--- a/src/js/select2/data/select.js
+++ b/src/js/select2/data/select.js
@@ -27,6 +27,21 @@ define([
 
   SelectAdapter.prototype.select = function (data) {
     var self = this;
+    var current = this.$element.val();
+    var options = self.options;
+
+    // If the selected data was already selected, clone the option and select it
+    // as well.
+    if (options && options.get('reuse') && current.includes(data.id)) {
+      // Clone the option removing the id such that a new one is generated.
+      var $option = this.option(data);
+      $option[0].removeAttribute('data-select2-id');
+      this.addOptions($option);
+
+      current.push(data.id);
+      this.$element.val(current);
+      this.$element.trigger('input').trigger('change');
+    }
 
     data.selected = true;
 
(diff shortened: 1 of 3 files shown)


jonahgreenthal:

sphuber:
I have been using this solution in our own deployment. I think there should be a generic use-case. I am not sure of the quality of my implementation, however, since I am not a JS programmer. Happy to improve if I can get pointers where needed.

rvandernoort:
I think this is a great idea! For my own project I was looking for something like this, but then instead of cloning the option, I was looking to count how many times the option was chosen. (use case is a money splitting with factors) I tried to add +1 and -1 buttons into the chosen options, ~~but the container seems to block any onclick actions for these buttons~~, which seems to work now, but is a bit sketchy implementation of this. Would be great if this could be further investigated in relation with this feature.

MZeeshanZafar:
@sphuber 
I want to use this in my code file. I have a dropdown to select the products. and now I want to order same product multiple times so I like this idea. How can I use this?

kevin-brown:
Unfortunately I'm going to reject this (5 years later) as something we don't want to add first-party support for within the library. While I see the use case here for it, it's not common enough to warrant the increased maintenance burden that we would receive from this.

PR #6111, opened 2021-10-18
Update docs.md

billynoah:
This pull request includes a

- [x ] Bug fix
- [ ] New feature
- [ ] Translation

wrong option name in docs


kevin-brown:
I'm pretty sure this is a change in 4.1.x which correctly reflects the option being changed.

billynoah:
@kevin-brown - Ok thanks for explaining. I was going based on the current release. I think It's nonsensical to have the live docs reflective of changes in an upcoming release.  Anyone looking at them and using the newest version will only be confused.

PR #6171, opened 2022-06-30
Allow using the module as a dependency in firefox extensions

pustovitDmytro:
This pull request includes a

- [x] Bug fix
- [ ] New feature
- [ ] Translation

Thank you for the nice plugin!

I'm developing a firefox extension, and want to use select2. But unfortunately [addon validator](https://addons.mozilla.org/en-US/developers/addon/validate) treats this dependency as insecure.  jQuery, underscore, react and the same famous libraries are white-listed by the validator, but select2 is not. 
Screenshot from 2022-06-30 21-57-14

So I changed innerHTML to jquery $().html. 
I understand it is unlikely a security fix, but at least it will allow using select2 in addons. And easier pass some static security audits.


kevin-brown:
I need to close this off because we're moving away from jQuery, not to jQuery. Unfortunately this means this library can't be used in Firefox plugins.

PR #6179, opened 2022-07-21
Unselect events: cache lookup key fix

### LINKED ISSUE #6128: Cannot unselect by clicking selected when using dataAdapter
When using select2 with `dataAdapter`, the selected option cannot be deselected by clicking it again on list.
If no `dataAdapter` option is used, it works fine, option can be deselected by clicking it second time.

Here's example with a bug I described, you can select any option and then try to deselect it by clicking it again.
http://jsfiddle.net/o53fLqum/

jayaddison:
This pull request includes a

- [x] Bug fix
- [ ] New feature
- [ ] Translation

- During `unselect` events, retrieve the HTML `option` element related to the unselection and use that as a cache key to retrieve item data

Resolves #6128.

### DIFF
--- a/src/js/select2/data/select.js
+++ b/src/js/select2/data/select.js
@@ -110,9 +110,7 @@ define([
     this.container = container;
 
     container.on('select', function (params) {
-      var option = container.$element.find('option')[0];
-      var data = Utils.GetData(option, 'data');
-      self.select(data);
+      self.select(params.data);
     });
 
     container.on('unselect', function (params) {
--- a/src/js/select2/data/select.js
+++ b/src/js/select2/data/select.js
@@ -114,7 +114,8 @@ define([
     });
 
     container.on('unselect', function (params) {
-      container.$element.find('option[value=' + params.data.id + ']').each(function () {
+      var unselected = 'option[value=' + params.data.id + ']';
+      container.$element.find(unselected).each(function () {
         var data = Utils.GetData(this, 'data');
         self.unselect(data);
       });
--- a/src/js/select2/data/select.js
+++ b/src/js/select2/data/select.js
@@ -110,11 +110,15 @@ define([
     this.container = container;
 
     container.on('select', function (params) {
-      self.select(params.data);
+      var option = container.$element.find('option')[0];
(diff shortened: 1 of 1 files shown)


src/js/select2/data/select.js
```diff
@@ -114,7 +114,12 @@ define([
     });
 
     container.on('unselect', function (params) {
-      self.unselect(params.data);
+      container.$element.find('option').each(function () {
+        if ($(this).val() == params.data.id) {
```
jayaddison, line 118: Note: this is somewhat intentionally not a complete-strict equality test (in other words: it is expected that this is `==` instead of `===`)

This follows a similar equality pattern check in [`array.js` here](https://github.com/select2/select2/blob/5c38f186c98d793c7d396442986191ca05634d0a/src/js/select2/data/array.js#L47).

If both option-value and data-id are guaranteed to be of type `string`, then this could be upgraded to a strict equality test.

src/js/select2/data/select.js
```diff
@@ -114,7 +114,12 @@ define([
     });
 
     container.on('unselect', function (params) {
-      self.unselect(params.data);
+      container.$element.find('option').each(function () {
```
jayaddison, line 117: @kevin-brown I'm feeling a bit uncertain about whether this change is a good idea, after looking at it again.

This extra `for` loop feels like it could quite easily introduce performance regressions.

I'll find time to take another look at this soon, but if in any doubt, please feel free to revert this change.

jayaddison: (update: re-inspecting this changeset currently)

jayaddison: Yep, a slightly stronger statement after re-reading the changeset and related issues: I think I'd prefer for this pull request to be reverted (and #6128 re-opened), if that's OK.

I didn't add any test coverage as part of the changes, which was unwise, because that makes it much harder to determine whether the modifications truly resolve the issue identified in the linked ticket.

jayaddison: (revert pull request opened as #6217)

PR #6180, opened 2022-08-04
Search pop-up preventing while `remove` btn clicking

slovoiztrehbukv:


kevin-brown:
Solution presented in https://github.com/select2/select2/issues/3209#issuecomment-149663474 is the currently accepted method of doing this. This PR unfortunately will not be accepted because it can be done using the linked method.

PR #6194, opened 2022-10-05
increased dropdown z-index

trustedcoder:
This pull request includes a

- [x] Bug fix

- Increased the dropdown z-index so that it can be more visible when using select on a bootstrap Modal


kevin-brown:
I'm going to have to close this one off unfortunately, at this point the value has been set to this for so long that changing it will hurt more people than it helps. You can still override this in your own CSS though.

PR #6195, opened 2022-10-07
Change results mouseup to click event to avoid accidental selection after clear

### LINKED ISSUE #6187: Select2 automatically selects a value
**Use Case Scenario**
So the scenario is I have a dropdown that I need to have a VALUE for. If it's empty, then I will need to show a specific element. The problem is, currently when I press the select2 dropdown clear button. It does show the elements I need that are previously hidden, but it automatically selects a value.

View the [working replication](https://jsbin.com/busujipuve/edit?html,js,output).

**Reproduction Steps via Clone**
1. Open the link.
2. Press the clear button in the select dropdown.
3. You will see that this automatically selects a value.

**Browsers Tested**
All of these were tested on Windows 10 10.0.19043 Build 19043 and Windows 11 Pro 10.0.22000 Build 22000.
1. Chrome Browser 105.0.5195.102 (Official Build) (64-bit) (cohort: 105_Win_102)
2. Mozilla Firefox 104.0.2 (64-bit)

I'm using the latest version of Select2 which is 4.0.13.

Brainshaker95:
This pull request includes a

- [x] Bug fix / UX improvement
- [ ] New feature
- [ ] Translation

The event for selecting results was changed from mouseup to click.  
This prevents the undesirable behavior that options can be clicked by accident after clearing the select.  
This is only relevant when clearing triggers a layout shift but I think since this is quite a commen use case this issue should be adressed somehow.

Resolves #6187

### DIFF
--- a/src/js/select2/results.js
+++ b/src/js/select2/results.js
@@ -328,7 +328,7 @@ define([
         return;
       }
 
-      $highlighted.trigger('mouseup');
+      $highlighted.trigger('click');
     });
 
     container.on('results:select', function () {
@@ -448,7 +448,7 @@ define([
       });
     }
 
-    this.$results.on('mouseup', '.select2-results__option--selectable',
+    this.$results.on('click', '.select2-results__option--selectable',
       function (evt) {
       var $this = $(this);
 


kevin-brown:
I'm closing this (3.5 years later) since this is not the fix we're going to target for this. We intentionally capture the `mouseup` event for the results to allow since click open+selections in dropdowns. We will likely address this on the clear button side so it does not propagate the mouseup to the results.

Issue #6196, opened 2022-10-12
Feature Request: Removal of jQuery as Dependency


RacerDelux:
To start, I know this was asked a while back. The reason I am reposting this feature request is that we are now in 2022 and the use of jQuery simply isn't required in many/most modern frameworks.

While I know this would be a large refactor, the good news is that JavaScript has a 1:1 replacement for pretty much anything you would do in jQuery. And since this project has type definitions, if done in TypeScript it would help quickly find and update all of the jQuery references anyway.

Sadly there are not too many good alternatives to select2 that do not include jQuery. I know many of us would be very glad to see the drop of a dependency.

Apart from the dependency removal, another bonus is that this will remove any bugs that occur from using different jQuery versions.

As a flip side, if this refactor is done, I am not sure if IE11 needs to be supported. Microsoft has dropped support in their own applications (last year I think) and its finally being left in the past. The only reason I mention this is that there are many powerful features that are unlocked the second IE11 is no longer in the picture.

ajiho:
Yes, bootstrap 5 is already doing this, and it has completely abandoned its support for ie11. I personally suggest that select2 should do the same.

kevin-brown:
I would absolutely love to get rid of the dependency on jQuery! But unfortunately there's a lot of ground to cover and only one of me, so I'm willing to accept pull requests tagged to this ticket.

RacerDelux:
> 

I have not looked at source yet, but is this built on TypeScript? If so it would make the switch over process pretty straight forward, just tedious.

If I get some extra time here I will look into doing some PRs. More than likely removing jQuery would be best in steps right? Or would you prefer it be in one PR?

kevin-brown:
Select2 is unfortunately not built on TypeScript (honestly I wish it was), but instead of AMD-compatible JavaScript. And yes, I prefer smaller PRs to make reviews a bit easier and quicker. It also greatly reduces the chances that someone gets half way through it and gives up, making all of their work go to waste.

stratboy:
Yes, jquery should be abandoned here.

duanemoody:
Looking over the 6500 lines of select2.full.js, it's mostly leveraging jQ for events (96 calls to `.on()`) and some property getters/setters, but ironically there's only one complex `$()` query selector in the code with the other 45 wrapping variable names. There is afaik no chaining in the codebase.

[...]

PR #6233, opened 2023-03-03
Remove jQuery in Utils and Translation

sabas:
This pull request includes a

- [ ] Bug fix
- [ ] New feature
- [ ] Translation

- I tried to remove jQuery from two files to see if I can help somehow
- I ran ```grunt test``` and all passed


src/js/select2/translation.js
```diff
@@ -15,7 +14,7 @@ define([
   };
 
   Translation.prototype.extend = function (translation) {
-    this.dict = $.extend({}, translation.all(), this.dict);
+    this.dict = Object.assign(translation.all(), this.dict);
```
kevin-brown: Doesn't this modify the translation dictionary with the already defined translations? We were doing the extend with the empty object before to get around modifying the translations and to ensure it was always a new object.

sabas: Ok sorry, I checked again the documentation and it can be done as in jquery, I add another commit

sabas:
#6196

src/js/select2/utils.js
```diff
@@ -230,8 +228,14 @@ define([
...
-    return ($el.innerHeight() < el.scrollHeight ||
-      $el.innerWidth() < el.scrollWidth);
+    var computed = window.getComputedStyle(el);
+    var paddingHeight = parseInt(computed.paddingTop, 10);
+        paddingHeight += parseInt(computed.paddingBottom, 10);
+    var paddingWidth = parseInt(computed.paddingLeft, 10);
+        paddingWidth += parseInt(computed.paddingRight, 10);
+  
+    return ((el.clientHeight - paddingHeight) < el.scrollHeight ||
+      (el.clientWidth - paddingWidth) < el.scrollWidth);
```
jayaddison, line 238: The `jQuery` approach seems more elegant and adaptable here - the API doesn't expose parsing of text values (such as `6px`) obtained from the computed style info.

Does this code follow a wide-recognized pattern for determining an element's inner dimensions, and if so is there a hyperlink to some documentation that explains it?

sabas: @jayaddison actually I don't remember the rationale, probably I followed some ideas I found researching

jayaddison: No problem.  I think it would be a significant benefit to remove the jQuery dependency; however I also remember why jQuery was originally popular -- to solve quirks and behaviours that can vary across browsers.  That means that I am sometimes a bit cautious about removing code that is expressed fairly clearly using jQuery -- it provides a single place to identify and solve those problems.

dtrucs, line 238: Why not just use the `width` and `height` from `getComputedStyle`?

```suggestion
    var computedEl = window.getComputedStyle(el);
 
    return (parseFloat(computedEl.height) < el.scrollHeight ||
      parseFloat(computedEl.width) < el.scrollWidth);
```

PR #6239, opened 2023-03-30
Fixup: convert optgroup child data identifiers to string datatype

### LINKED ISSUE #3882: Cannot deselect items in grouped multiple select when data ids are not strings

jayaddison:
This pull request includes

- [x] Bug fix

- Add a unit test case intended to cover behaviour initially reported in #3882.
- Ensure that optgroup (children) data identifiers are converted to `string`.

Resolves #3882.


jayaddison:
Argh, I don't think this is testing the correct behaviour.

Currently the code here tests clicks on the 'remove' button on each tag within the multi-select.

Re-attempting the demo from #3882, that behaviour works fine.  What is buggy is selecting an item from the drop-down, and then attempting to de-select it, but again by using the dropdown.

Hopefully that is a small change to the test code..

tests/integration/select2-methods.js
```diff
@@ -137,3 +137,70 @@ test('multiple value matches the jquery value', function (assert) {
...
+  var $selections = $('.select2-results__option[aria-selected=true]');
+
+  assert.equal(
+    $selections.length,
+    1,
+    'One item should remain selected'
+  );
+
+  // Remove the second selection by clicking on the item in the dropdown
+  $selections.first().trigger('click');
```
jayaddison: The idea with this line is to simulate the user clicking on the dropdown menu item, so that we can check that the selection is removed afterwards. It doesn't appear to invoke a relevant event handler though.

tests/integration/select2-methods.js
```diff
@@ -137,3 +137,70 @@ test('multiple value matches the jquery value', function (assert) {
...
+
+  var $remove = null;
+
+  // Remove the first item
+  $remove = $container.find('.select2-selection__choice__remove');
+  $remove.trigger('click');
+
+  // Open the dropdown menu
+  $container.find('.select2-selection').trigger('click');
+  var $selections = $('.select2-results__option[aria-selected=true]');
```
jayaddison: I'm not sure that this `jquery` query (that starts from the global '$' object -- instead of `$container` or anything local to this function) is returning the expected objects.

What I'd like it to find is the selected items in the dropdown for the multi-select created within this test case.

I have a sense that it's retrieving elements from another, unrelated integration test.

tests/integration/select2-methods.js
jayaddison: Changing the event from `click` to `mouseup` -- to correspond with [the event handler here](https://github.com/openculinary/select2/blob/db929cb6b835b30a31af72932335179fafb61f8d/src/js/select2/results.js#L451) seems to make a difference.  It causes the `dom-changes.js` integration test to break, though:

```
>> select2(val) - multiple selection and clearing of grouped options
>> Message: Too many calls to the `assert.async` callback
>> Actual: null
>> Expected: undefined
>> http://localhost:9999/tests/integration/dom-changes.js:284:14
```

jayaddison:
Ok, the test coverage here for #3882 _seems_ to also validate the fix attempted originally in #6179 (subsequently reverted due to lack of test coverage).

Even so: I don't think that the fix from #6179 is ideal.  There could be a better approach possible.  I'll spend some time to investigate that while looking at this.

kevin-brown:
I'm wondering if it would make sense to apply this fix within [`SelectAdapter._normalizeItem`](https://github.com/select2/select2/blob/4de1d236e602149a5c78d35899b49b3989b94107/src/js/select2/data/select.js#L252), likely by calling `_normalizeItem` on all of the children recursively if `data.children` is present (something I didn't realize we didn't do). I suspect that might allow this to be applied a little more consistently, so it works even if the results dropdown has not been loaded.

jayaddison:
Yep, that sounds sensible - I'll take another look at the code and see how to refactor it like that.

jayaddison:
(I guess that's not really refactoring; more like a better fix.  anyway, taking a look into that now)

src/js/select2/data/select.js
```diff
@@ -278,6 +278,10 @@ define([
       item._resultId = this.generateResultId(this.container, item);
     }
 
+    if (item.children) {
+        item.children = item.children.map(this._normalizeItem);
```
jayaddison: Would it make sense to limit the recursion depth here?

jayaddison: It's probably overly-pedantic, but it does seem that [HTML `optgroup` elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/optgroup) don't support nesting, so there could be a small safety improvement by preventing recursive calls here. 

It's unlikely that circular references or multiple levels of nesting would appear in input data often on the web, but unless there's a downside to preventing recursive calls (other than slightly less-clean code), I figure it could make sense to limit the recursion to a single level.

kevin-brown: Not particularly worried about infinite recursion happening here, we allow as many levels of nesting as people want in a few other places without depth checks. For our AJAX use cases we have to allow multiple levels here, so we can skip the recursion checks.

jayaddison:
@kevin-brown I think this is ready for (re)review now; it's also been rebased against the latest `develop` branch changes.

kevin-brown (approved):
LGTM, the recursive data normalization approach makes this quite a bit cleaner.

jayaddison:
Thank you @kevin-brown!

PR #6241, opened 2023-04-05
Fixup: unselection of items from AJAX data sources with non-string identifiers

### LINKED ISSUE #6128: Cannot unselect by clicking selected when using dataAdapter

jayaddison:
This pull request includes a

- [x] Bug fix

- Adds the [`jquery-mockjax`](https://github.com/jakerella/jquery-mockjax/) plugin as a vendored integration test dependency.
- Provides integration test coverage for selection-then-unselection of items based on an AJAX data source.
- Performs normalization of result items received from AJAX sources.

Resolves #6128.

### DIFF
--- a/src/js/select2/data/ajax.js
+++ b/src/js/select2/data/ajax.js
@@ -66,19 +66,16 @@ define([
 
     function request () {
       var $request = options.transport(options, function (data) {
-        if (data.results) {
-          data.results = data.results.map(AjaxAdapter.prototype._normalizeItem);
-        }
         var results = self.processResults(data, params);
 
-        if (self.options.get('debug') && window.console && console.error) {
+        if (results && results.results && Array.isArray(results.results)) {
+          results.results = results.results.map(AjaxAdapter.prototype._normalizeItem);
+        } else if (self.options.get('debug') && window.console && console.error) {
           // Check to make sure that the response included a `results` key.
-          if (!results || !results.results || !Array.isArray(results.results)) {
-            console.error(
-              'Select2: The AJAX results did not return an array in the ' +
-              '`results` key of the response.'
-            );
-          }
+          console.error(
+            'Select2: The AJAX results did not return an array in the ' +
+            '`results` key of the response.'
+          );
         }
 
         callback(results);
--- a/src/js/select2/data/ajax.js
+++ b/src/js/select2/data/ajax.js
@@ -69,13 +69,17 @@ define([
         var results = self.processResults(data, params);
 
         if (results && results.results && Array.isArray(results.results)) {
-          results.results = results.results.map(AjaxAdapter.prototype._normalizeItem);
-        } else if (self.options.get('debug') && window.console && console.error) {
-          // Check to make sure that the response included a `results` key.
-          console.error(
-            'Select2: The AJAX results did not return an array in the ' +
-            '`results` key of the response.'
+          results.results = results.results.map(
+            AjaxAdapter.prototype._normalizeItem
           );
+        } else {
+          if (self.options.get('debug') && window.console && console.error) {
+            // Check to make sure that the response included a `results` key.
+            console.error(
+              'Select2: The AJAX results did not return an array in the ' +
+              '`results` key of the response.'
+            );
+          }
         }
 
         callback(results);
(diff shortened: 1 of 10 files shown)


kevin-brown:
> * Maybe-controversial: removes some special-case handling for multi-select controls

Yeah those are too controversial to roll in here. We intentionally don't have single-select elements issue `unselect` because that'd always be paired with a `select`.

src/js/select2/data/ajax.js
```diff
@@ -66,6 +66,9 @@ define([
 
     function request () {
       var $request = options.transport(options, function (data) {
+        if (data.results) {
+          data.results = data.results.map(AjaxAdapter.prototype._normalizeItem);
+        }
         var results = self.processResults(data, params);
```
kevin-brown:
```suggestion
        var results = self.processResults(data, params);
        if (results) {
          results = results.map(AjaxAdapter.prototype._normalizeItem);
        }
```

I'm thinking this should happen after the results processing, since in many cases people bring the results into compliance through that method already (as well as doing things like mapping `id` from other fields).

So this would actually get moved down ahead of the callback and would also need to be guarded by `Array.isArray()` based on the debug catch which we have below this block. Sometimes people don't return an array properly so we have to kind of just roll with it and let it happen.

kevin-brown (changes requested):

kevin-brown:
I am likely going to spin a new `ajax-tests.js` module out of this and move this test into it, since we don't currently have good tests around AJAX handling and this would be a solid first test to add. More than likely I would switch this to just be testing the `AjaxAdapter` itself and create a new test around the weirdness with creating `<option>`s on select events.

src/js/select2/data/ajax.js
jayaddison: Thanks, ok - that makes sense :+1:

jayaddison:
Thanks @kevin-brown - how would you like to co-ordinate the creation of that test file?  (should I attempt to create it here, or would you prefer to do that separately and move these changes into it?)

jayaddison:
> Yeah those are too controversial to roll in here. We intentionally don't have single-select elements issue `unselect` because that'd always be paired with a `select`.

Sounds good; the single-select `unselect` event change here has been reverted.

src/js/select2/data/ajax.js
jayaddison: (`Array.isArray` guard added, and a single conditional clause is now used to check the post-processed results in the correct-data path, and for debug error reporting in the malformed-data path)

jayaddison:
Also a note: the `jquery-mockjax` component is available under dual MIT-or-GPLv2 licensing terms (it's up to the authors of the codebase where it is included to decide on the applicable terms).  I think I did check the license compatibility before choosing it for this branch, but wanted to double-check that again anyway.

kevin-brown:
Good call on checking the license, thank you for switching back the unselect changes. This looks good to go from a high level, so I'll mark it for my next sweep.

I'll take care of the testing changes independently, the tests you wrote get the job done for this change set. Anything I add would likely be in addition to the ones here and aren't required.

jayaddison:
Hey @kevin-brown - I'm afraid I'm joining the chorus of 'any updates on this' here - is there anything I can do to help this and/or the 4.1.0 milestone along?

jayaddison:
Note, partly since licensing was discussed here previously: `jquery-mockjax` v2.6.1 has consolidated from dual MIT-and-GPLv2 licensing to a single MIT license (ref: https://github.com/jakerella/jquery-mockjax/commit/8183c940d8b41faaaec39657c4e931273645eab9 and the [associated changelog entry](https://github.com/jakerella/jquery-mockjax/blob/4ca7ace3daf8cfb54134555ee91c4b1cff671065/CHANGELOG.md?plain=1#L2)).

jayaddison:
Hi @kevin-brown - I've been browsing through a few old pull requests to clear the decks - any chance this might be nearish-acceptable?  (I've one or two others open also, but this one seemed to be most-progressed).  No problem if not.

kevin-brown (approved):

kevin-brown:
Just realized I merged this in _right after_ we upgraded QUnit so the tests are now failing. Should be a quick fix to update the calls.

Issue #6262, opened 2023-07-18
HTML markup appears inside input when markup escaping disabled


zkelo:
# Info
Select2 version: `4.0.13`
JSBin: https://jsbin.com/wofuxebipe/1/edit?html,js,output

I made `select2` with enabled `multiple` option and some items _(in this examples is books)_. Each book have its keywords that shouldn't appear in `<option>` tags but should exist to make them searchable.

![1689695393](https://github.com/select2/select2/assets/13527447/1012ce7e-7b05-42a7-a020-f5d4561574c9)

- First, I changed `escapeMarkup` option to my function that doesn't escape markup. This works good except one case - when user chooses at least one item and pressing up `Backspace`, widget inserts into input all data including tag with hidden keywords;
- Second, I also tried change `templateResult` and `templateSelection` options to hide markup inside results and selection and it also doesn't work.

I looked on documentation about this and not found ways to prevent this. May be I lost something?

# Steps to reproduce

1. Choose one item and focus on input;
    ![1689697109](https://github.com/select2/select2/assets/13527447/82cf197c-00dc-45ec-9637-b02684fc687d)
2. Press `Backspace` and you'll see markup.
    ![1689697151](https://github.com/select2/select2/assets/13527447/4241573f-1024-4c9c-9e77-815b910b773e)

kevin-brown:
This appears to be more of a usage question than a bug report or common feature request.  As we continue to keep the Select2 issue tracker geared towards these two areas, bug reports and common feature requests, we are redirecting usage questions to other communities such as the [mailing list][ml], [IRC][irc], or [Stack Overflow][so].

[ml]: https://select2.org/troubleshooting/getting-help
[irc]: https://select2.org/troubleshooting/getting-help
[so]: http://stackoverflow.com/questions/tagged/jquery-select2?sort=votes

PR #6263, opened 2023-07-19
More options

zkelo:
This pull request includes:

- [ ] Bug fix
- [x] New feature
- [ ] Translation

- Added new option `searchDecorator`;
- Added documentation for new option.

This PR helps to solve my issue #6262 by adding special option for specifying search decorator. Using this I can define custom search decorator and easy pass it to widget.

This is how I did it with new option:
```js
// Decorator definition
$.fn.select2.amd.define('CustomSearch', [
    'select2/selection/search',
    'select2/utils'
], function (Search, Utils) {
    function CustomSearch (decorated, $element, options) {
        decorated.call(this, $element, options);
    }

    Utils.Extend(CustomSearch, Search);

    CustomSearch.prototype.searchRemoveChoice = function (decorated, item) {
        this.trigger('unselect', {
          data: item
        });

        this.$search.val(item.text.substr(0, item.text.indexOf('<') - 1));
        this.handleSearch();
    };

    return CustomSearch;
});

// Init select2
$('#books').select2({
    width: 500,
    multiple: true,
    data: [
        {
            id: 1,
            text: 'First book <span style="display: none;">one 1</span>'
        },
        {
            id: 2,
            text: 'Second book <span style="display: none;">two 2</span>'
        },
        {
            id: 3,
            text: 'Third book <span style="display: none;">three 3</span>'
        }
    ],
    escapeMarkup(data) {
        return data;
    },
    searchDecorator: $.fn.select2.amd.require('CustomSearch')
});
```

I think it will be useful not only for me so I made this PR. Also I can add more similar options and documentation for them if this will be useful.


kevin-brown:
Unfortunately I need to close this PR (3 years later) since it's not one which is likely to land with first-party support within Select2. The plugin strategy via decorators made sense originally but never really took off, so we're evaluating them for future major updates.

PR #6279, opened 2023-10-13
fix: Removed ... suffix from nl.js file

Raheman-08:
This pull request includes a

- [ ] Bug fix
- [ ] New feature
- [x] Translation

-
-
-


kevin-brown:
I believe this aligns with the English translations (having the ellipsis) so I'm going to have to close this unless someone teaches me that the elipsis is implied in this language file after so many years.

PR #6282, opened 2023-10-18
ajax datasources: allow search result selection to refresh previously-chosen option labels

### LINKED ISSUE #5391: Ajax remote data : outdated text displayed on re-selection
Hello,

I have a select2 element with an ajax datasource. If a user selects a value, then deselects it and finally reselects it, while text has changed server-side for the same id, the old version of text is displayed instead of the new one.

I've been able to get around it by deleting options at selection, but of course this is not appropriate if select2 allows multiple selection (previous selected elements will be deleted too) : 

```javascript
if (!is_multiple) {
    $select2_element.on('select2:selecting', function (event) {
        $select.find('option').remove();
    });
}
```

Is there a way to disable or get around this caching behavior ? Or, at least, can I refresh by myself options after ajax response ? 

Here is the classical select2 config : 

```javascript
var select2Config = {
    ajax: {
        url: ...,
        data: function (params) {
            return {...};
        },
        dataType: 'json',
        delay: 250,
        processResults: function (data) {...},
        cache: true
    },
    escapeMarkup: function (markup) {...},
    placeholder: '',
    width: 205,
    language: cur_locale
    multiple : true
};
```

jayaddison:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

- Remove locally-cached HTML `<option>` elements from the DOM when an AJAX-sourced item is selected.
- This should allow the data received from AJAX-based (re)selections to be refreshed.

May resolve #5391.
Based upon #6241.

Edit 20240913: s/unselected/selected


tests/integration/select2-methods.js
```diff
@@ -137,3 +137,60 @@ test('multiple value matches the jquery value', function (assert) {
     'The values should match the jquery values'
   );
 });
+
+test('selection and clearing of data from ajax source', function (assert) {
```
jayaddison: To prove that this branch fixes #5391, a second response from the AJAX API should be mocked-out -- with updated, differing metadata for one of the returned items -- and an assertion should be added to check that selecting the updated item is reflected on the control.

tests/integration/select2-methods.js
```diff
@@ -137,3 +137,87 @@ test('multiple value matches the jquery value', function (assert) {
...
+
+  var select = new Select2($select, {ajax: {url: dataURL}, multiple: true});
+
+  assert.equal(
+    $select.find(':selected').length,
+    0,
+    'No items should be selected'
+  );
+
+  var queryTerms = ['firstQuery', 'secondQuery', 'thirdQuery', 'fourthQuery'];
```
jayaddison, line 165: This seems like a really smelly way to achieve multiple visits and actions on the `select2` control.  My JavaScript-fu isn't great, so there's almost certainly a better way to do this.

tests/integration/select2-methods.js
```diff
@@ -137,3 +137,87 @@ test('multiple value matches the jquery value', function (assert) {
...
+      assert.equal(
+        $select.find('option').length,
+        1,
+        'An HTML option element should have been created for the item'
+      );
+
+      // Trigger a fourth call
+      select.selection.trigger('query', {term: 'fourthResult'});
+
+    // Third call: assert that the updated text is in place
```
jayaddison: Oops: this should be `Fourth call: ...`.  I've prepared that commit locally, but won't push it until after the rest of the changeset has had some review.

jayaddison:
Moving this pull request into a 'draft' status; option groups (nested results) need to be handled too.

jayaddison:
> Moving this pull request into a 'draft' status; option groups (nested results) need to be handled too.

From testing locally: option groups _do not_ result in the creation of nested HTML elements in the DOM-cached data; a flattened set of `<option>` elements are stored, the same as for flat datasources.

As a result: I don't think that any special-case handling is required for option groups, and I'll put this pull request back into a 'ready for review' state.

The testing methodology I used was to create a small Python script to generate dummy nested JSON data in the [`select2` data format](https://select2.org/data-sources/formats) and to run that every few seconds by using `watch -n 2 python write_data.py` to (re)write JSON data to file.

In combination with a basic test HTML page (loading JQuery and Select2 JS and CSS files) with a `<select multiple>` and some `select2({ajax: ...})` configuration, and `python -m http.server -b 127.0.0.1` to serve the data and HTML statically, that was enough to test some dynamic grouped result querying, selection, and unselection.

tests/integration/select2-methods.js
jayaddison: (it's been a while, so I went ahead and applied this change)

src/js/select2/data/select.js
```diff
@@ -70,6 +70,14 @@ define([
   SelectAdapter.prototype.unselect = function (data) {
     var self = this;
 
+    // Remote-origin (AJAX) items: remove matching option elements from the DOM
```
kevin-brown: Would it make sense to move this to the AJAX adapter so that way we know for sure that AJAX is enabled? It may require us to add the `unselect` method and call `super` there, but that should be easy since it's happening before the rest of the logic.

jayaddison: That does make sense, yep - and migrating the fix there (without a `super` call) works.

However: adding the `super` call seems to result in test failures:

```
Testing http://localhost:9999/tests/integration-jq1.html ...Failed to load resource: the server responded with a status of 404 (Not Found)
..........
>> TypeError: Cannot read properties of undefined (reading 'prop')
>>     at SelectAdapter.unselect (http://localhost:9999/dist/js/select2.full.js:3373:24)
```

jayaddison:
Although tests pass, I think I should re-test this manually.  Moving into draft status until that is done.

jayaddison:
No, this doesn't quite function the way I'd like it to in practice yet.

When selecting and unselecting items from an AJAX multi-select, it functions correctly.  But the following situations don't work particularly well:

  * Re-selecting an item in a single-select control (the label of the selected item does not update).
  * Re-selecting an item in a multi-select control (the selected item disappears).

[select2-issue-6282-testing.webm](https://github.com/user-attachments/assets/49a59557-0f50-4d9a-990c-b9cbd971408f)

<details>

<summary>Test Python API Code</summary>

```python
from random import randint

from flask import Flask, jsonify
from flask_cors import CORS

app = Flask(__name__)
CORS(app)


@app.route("/ungrouped")
def ungrouped():
    return jsonify({
        "results": [
            {
                "id": 1,
                "text": "Option 0"
            },
            {
                "id": 2,
                "text": f"Option {randint(1, 9)}"
            },
            {
                "id": randint(3, 9),
                "text": "Option n"
            },
        ]
    })

@app.route("/grouped")
def grouped():
    return jsonify({
        "results": [
            {
                "text": "Group 1",
                "children" : [
                    {
                            "id": 1,
                            "text": "Option 1.0"
                    },
                    {
                            "id": 2,
                            "text": f"Option 1.{randint(1, 9)}"
                    }
                ]
            },
            {
                "text": "Group 2",
                "children" : [
                    {
                            "id": 3,
                            "text": "Option 2.0"
                    },
                    {
                            "id": 4,
                            "text": f"Option 2.{randint(1, 9)}"
                    }
                ]
            },
            {
                "text": "Group 3",
                "children" : [
                    {
                            "id": randint(5, 9),
                            "text": "Option n"
                    }
                ]
            }
        ]
    })
```
</details>

<details>

<summary>Test HTML User Interface</summary>

```html
<!DOCTYPE html>
<html>
  <head>
    <link rel="stylesheet" href="dist/css/select2.css" />
    <script src="tests/vendor/jquery-3.4.1.js"></script>
    <script src="dist/js/select2.full.js"></script>
  </head>
  <body>
    <h1>Single-selects</h1>
    <div><select class="ungrouped" style="width: 200px"></select></div>
    <div><select class="grouped" style="width: 200px"></select></div>
    <h1>Multi-selects</h1>
    <div><select class="ungrouped" style="width: 200px" multiple="multiple"></select></div>
    <div><select class="grouped" style="width: 200px" multiple="multiple"></select></div>
    <script>
      $('select.ungrouped').select2({ajax: {url: 'http://127.0.0.1:5000/ungrouped', dataType: 'json'}});
      $('select.grouped').select2({ajax: {url: 'http://127.0.0.1:5000/grouped', dataType: 'json'}});
    </script>
  </body>
</html>
```

</details>

jayaddison:
>  * Re-selecting an item in a single-select control (the label of the selected item does not update).

I think that the label should be updated in this case: an example would be a situation where the user is selecting city-names, and has a separate language-selection control, either in the browser or in the web application.  Let's assume that city identifiers are translated into multiple languages.

So a user could select `London` after searching in English, then search for `Londra` in Italian; both items have identifier `LON`.  The selected option in the control should remain labelled as `London` -- it's what the user requested at the time they made their initial selection -- until/unless they subsequently click on `Londra` from the search results, at which point the label should be updated (because the user took an action that re-selected the same label), but the selected ID value for the control should not change (it remains `LON`).

>  * Re-selecting an item in a multi-select control (the selected item disappears).

Although not entirely intuitive in my example video -- perhaps because of the randomized format of the labels -- this does in fact make sense.  In the video I initially selected an item with ID:2 but labeled `Option 8`, and then when I opened the search for a second time, an item labelled as `Option 1` -- but also in fact with ID:2 -- was subsequently clicked.  It is momentarily apparent from the grayed CSS styling in the dropdown menu that in fact it was already selected.  And since an already-selected item was chosen again, the inference in the control's logic is that the user wants to de-select that item -- and so it is removed from the selection and the tag disappears from the control.

jayaddison:
Given what I've learned about this: I think that this is becoming more of a feature implementation ("allow AJAX responses to dynamically update selected option labels") than a bugfix.

jayaddison:
I think the approach I've been following in this pull request is incorrect; the goal shouldn't be to remove the cached entries, but instead to invalidate and refresh them.

The test coverage is therefore too-precise; it dictates that previously-selected `<option>` elements must be removed, for example -- but I'm not convinced that that is strictly necessary.

I plan to open a feature request with a clearer motivating example relating to selection and re-selection of internationalized labels from AJAX results.

jayaddison:
No.. it's still feasible with this approach.  I just wonder whether there might be a cleaner approach.  I'll push some changes that provide a working -- but maybe inelegant -- solution in a few moments.

src/js/select2/data/ajax.js
```diff
@@ -110,5 +110,19 @@ define([
     }
   };
 
+  AjaxAdapter.prototype.unselect = function (data) {
+    var self = this;
+
+    // Remove matching option elements from the DOM
+    this.current(function (currentData) {
```
jayaddison: This feels like a misuse of the `unselect` event -- because we're ignoring the `data` argument -- which ideally would provide all the information we need -- and instead accessing the control's currently-selected items to unselect everything.

jayaddison: Resolved: moved the cache invalidation to the `select` function on the AJAX adapter instead.

jayaddison:
Ok; I think the updated behaviour and logic is clearer.

  * The labels of AJAX items from remote data may change; local state to determine whether they are selected or not continues to be keyed solely on the ID of each item.
  * When the user selects a remote AJAX item, the label will be updated with the most-recent label from the remote source.

Internally this is achieved by removing cached HTML `<option>` elements, if they exist, when a selection is made on a control that uses AJAX.

**Video demo**

[select2-issue-6282-testing-singleselect-updates.webm](https://github.com/user-attachments/assets/250cb2d4-364d-407b-94cb-66a7f50bf57b)

<details>

<summary>Test Python API Code</summary>

```python
from random import randint

from flask import Flask, jsonify
from flask_cors import CORS

app = Flask(__name__)
CORS(app)


@app.route("/ungrouped")
def ungrouped():
    return jsonify({
        "results": [
            {
                "id": 1,
                "text": "Option 0"
            },
            {
                "id": 2,
                "text": f"Option {randint(1, 9)}"
            },
            {
                "id": randint(3, 9),
                "text": "Option n"
            },
        ]
    })

@app.route("/grouped")
def grouped():
    return jsonify({
        "results": [
            {
                "text": "Group 1",
                "children" : [
                    {
                            "id": 1,
                            "text": "Option 1.0"
                    },
                    {
                            "id": 2,
                            "text": f"Option 1.{randint(1, 9)}"
                    }
                ]
            },
            {
                "text": "Group 2",
                "children" : [
                    {
                            "id": 3,
                            "text": "Option 2.0"
                    },
                    {
                            "id": 4,
                            "text": f"Option 2.{randint(1, 9)}"
                    }
                ]
            },
            {
                "text": "Group 3",
                "children" : [
                    {
                            "id": randint(5, 9),
                            "text": "Option n"
                    }
                ]
            }
        ]
    })
```
</details>

<details>

<summary>Test HTML User Interface</summary>

```html
<!DOCTYPE html>
<html>
  <head>
    <link rel="stylesheet" href="dist/css/select2.css" />
    <script src="tests/vendor/jquery-3.4.1.js"></script>
    <script src="dist/js/select2.full.js"></script>
  </head>
  <body>
    <h1>Single-selects</h1>
    <div><select class="ungrouped" style="width: 200px"></select></div>
    <div><select class="grouped" style="width: 200px"></select></div>
    <h1>Multi-selects</h1>
    <div><select class="ungrouped" style="width: 200px" multiple="multiple"></select></div>
    <div><select class="grouped" style="width: 200px" multiple="multiple"></select></div>
    <script>
      $('select.ungrouped').select2({ajax: {url: 'http://127.0.0.1:5000/ungrouped', dataType: 'json'}});
      $('select.grouped').select2({ajax: {url: 'http://127.0.0.1:5000/grouped', dataType: 'json'}});
    </script>
  </body>
</html>
```

</details>

src/js/select2/data/select.js
jayaddison: Done; note also that this logic has been relocated to the `select` function, with a `super` call in place.

jayaddison:
I'm not completely sure whether it does in fact make sense to remove these locally-cached options.

Yes, it allows AJAX results to update dynamically - but the behaviour that can result from that in a few cases -- for example, when a previously-selected item reappears in the AJAX results with a different name -- are not entirely intuitive.

I'll hold this pull request open for a while longer, but wanted to mention that I'm considering closing it.

jayaddison:
As noted in my previous comment, I think that some of the behaviours that occur in relation to differing AJAX response content can be tricky to reason about, from both a user and also a logical perspective.

As such, and because this feature isn't yet considered essential for functionality I'm working on, nor evident community demand, I think I'll close this PR.  Anyone should feel free to continue on from this branch (`git fetch <remote> pull/6282/head:localbranch` should retrieve this PR at 0b5293bdf1f838f0f6f5fe6b04663114cf5a8739) if the could would be helpful for you.

I failed to document it clearly, but the particular future use-case I had in mind for this involved dynamic change-of-interface language during a web session.  The ability to change from `en` (English) to `es` (Spanish) and then continue interacting with a select control -- which might already contain some selected options -- could be useful.

PR #6304, opened 2024-03-21
Fix bug which causes tags to not work if dataAdapter is set

Nilotaviano:
This pull request includes a

- [x] Bug fix
- [ ] New feature
- [ ] Translation

- I moved a closing bracket to the right place
- This makes it possible to set a dataAdapter while still being able to use `tags` and other options that were affected by this bug

### DIFF
--- a/src/js/select2/defaults.js
+++ b/src/js/select2/defaults.js
@@ -68,38 +68,38 @@ define([
       } else {
         options.dataAdapter = SelectData;
       }
+    }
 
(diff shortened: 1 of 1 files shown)


kevin-brown:
Unfortunately I'm going to have to close this PR (2 years later) since this is actually intended functionality. I recognize that the behaviour is not great, but it's the only way to allow for decorators to be selectively removed instead of requiring them to always be set for these options.

PR #6309, opened 2024-05-01
Adding options for multiple decorators

5andr0:
The current implementation only allows to overwrite the adapters.  
If a user only wants to add a decorator they have to call `Defaults.apply(options)` prior to `select2()` to build the default adapters with its optional decorators based on the given options and they have to import the utils module to build it. 

The problem is, that since you only had the option to change the whole adapter, all the plugins out there build their own adapters, without adding the default decorators.
That way you cannot even combine two whole plugin adapters.

It's better to incentivize devs to actually use the decorator system!

With this new implementation you just have to pass an array of decorators, which are added in order to the default adapters (based on the current options)

---

### This looks like a dead project, so here's a polyfill for those who want to use it while this isn't merged

```js
$.fn.select2.amd.require(['select2/defaults', "select2/utils"],
  function (Defaults, Utils) {
    let __super__apply = Defaults.apply.bind(Defaults);
    const adapterFields = Object.entries(__super__apply({})).reduce((pV,[k,v]) => {
      if (v.prototype && v.prototype.bind) {
        pV.push(k);
      }
      return pV;
    }, 
    []);

    Defaults.apply = function(options) {
      const opt = __super__apply(options);
      for(const adapter of adapterFields) { 
        const decorators = options[adapter + 'Decorators'];
        if(decorators instanceof Array)
        {
          for(const decorator of decorators) {
            opt[adapter] = Utils.Decorate(opt[adapter], decorator);
          }
        }
      }
      return opt;
    }
  }, null, true // forceSync = true
);
```

This pull request includes a

- [ ] Bug fix
- [X] New feature
- [ ] Translation

- added following options to pass an array of decorators:
    -  `dataAdapterDecorators`
    - `resultsAdapterDecorators`
    - `dropdownAdapterDecorators`
    - `selectionAdapterDecorators`

### DIFF
--- a/src/js/select2/defaults.js
+++ b/src/js/select2/defaults.js
@@ -100,6 +100,11 @@ define([
           Tokenizer
         );
       }
+
+      options.dataAdapter = this._decorateAdapter(
+        options.dataAdapter,
+        options.dataAdapterDecorators
+      );
     }
 
     if (options.resultsAdapter == null) {
@@ -132,6 +137,11 @@ define([
           TagsSearchHighlight
         );
       }
+
+      options.resultsAdapter = this._decorateAdapter(
+        options.resultsAdapter,
+        options.resultsAdapterDecorators
+      );
     }
 
     if (options.dropdownAdapter == null) {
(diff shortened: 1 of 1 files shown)


kevin-brown:
Unfortunately I'm going to have to close this PR (2 years later) since it's not something we're likely to accept as-is into Select2. We recognize that the decorator pattern as it is currently implemented is not great and this is a solid idea for how to make it less bad.

Issue #6298, opened 2024-02-07
Support for jQuery 4


marcellov7:
Hi,

I'm an enthusiastic user of Select2 and greatly appreciate its features and effectiveness. However, with the recent announcement of the release of jQuery 4 and its promises to significantly improve library performance, I was wondering if there are plans to ensure Select2 compatibility with this new version of jQuery.
https://blog.jquery.com/2024/02/06/jquery-4-0-0-beta/

I understand that Select2 relies on jQuery and that the upgrade may take time and effort, but considering the potential benefits in terms of performance and future compatibility, it would be extremely helpful for the community to know if there is a commitment to making Select2 compatible with jQuery 4.

Is there a roadmap or defined plan for this? Additionally, if there are opportunities to contribute or support this effort, I would be happy to provide my assistance.

Thank you

jamesleesaunders:
One first observation testing select2 with jQuery4 is:
- jQuery.isArray is deprecated; use Array.isArray

jQuery: 3.7.1
select2: 4.1.0-rc.0
django-select2: 8.1.2

```
        if (self.options.get('debug') && window.console && console.error) {
          // Check to make sure that the response included a `results` key.
          if (!results || !results.results || !$.isArray(results.results)) {             <--- HERE
            console.error(
              'Select2: The AJAX results did not return an array in the ' +
              '`results` key of the response.'
            );
          }
        }
```

```
jquery.js:1 JQMIGRATE: Migrate is installed with logging active, version 3.4.1
jquery.js:1 jQuery Version 3.7.1 Loaded
jquery.js:1 JQMIGRATE: jQuery.isArray is deprecated; use Array.isArray
a @ jquery.js:1
t.<computed> @ jquery.js:1
n._resolveLanguage @ select2.full.min.js:2
n.applyFromElement @ select2.full.min.js:2
e @ select2.full.min.js:2
d @ select2.full.min.js:2
(anonymous) @ select2.full.min.js:2
each @ jquery.js:1
each @ jquery.js:1
r.fn.select2 @ select2.full.min.js:2
initHeavy @ django_select2.js:48
(anonymous) @ django_select2.js:56
each @ jquery.js:1
$.fn.djangoSelect2 @ django_select2.js:53
(anonymous) @ django_select2.js:71
l @ jquery.js:1
c @ jquery.js:1
setTimeout (async)
(anonymous) @ jquery.js:1
u @ jquery.js:1
fireWith @ jquery.js:1
fire @ jquery.js:1
[...]

PR #6332, opened 2024-07-29
Add support for jQuery 4.0.0-beta.2

jayaddison:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

- Prepares support for jQuery 4.0.0 by testing against the second beta of that release (as [announced on the jQuery blog](https://blog.jquery.com/2024/07/17/second-beta-of-jquery-4-0-0/)).
- Builds on work by @marcellov7 to add unit and integration test coverage using [`jquery-4.0.0-beta.2`](https://github.com/jquery/jquery/commit/51fffe9f7395f86fb24c59115c9b98855c39fc07).
- Fixes a small bug in the `$.camelCase` replacement implementation.

Relates to (but does not resolve) #6298.

Edit: remove dependency changes and no-longer-included commits from this description.


docs/themes/site/js/data-fill-from.js
```diff
@@ -5,7 +5,7 @@ $(document).ready(function() {
     var codeContainer = $this.data("fill-from");
     var $container = $(codeContainer);
 
-    var code = $.trim($container.html());
+    var code = $container.html().trim();
```
JCCChris: https://api.jquery.com/jQuery.trim/

$.trim() handles nulls and undefined.  Should this also handle that?

jayaddison: Yep, that seems worthwhile - it does seem that `.html()` can return `undefined` if the container element is not found in the DOM.

.github/workflows/main.yml
```diff
@@ -20,14 +20,19 @@ jobs:
...
     steps:
+      - run: apt update
+      - run: apt install -y chromium-browser
       - uses: actions/checkout@v1
-      - name: Use Node.js 12
+      - name: Use Node.js 18
         uses: actions/setup-node@v1
         with:
-          node-version: 12
+          node-version: 18
```
jayaddison: Two notes about the suggested Node v18 upgrade here:

  * Currently, Node v20 (`iron`) is the long-term-support (LTS) version of Node - so that might be a better choice here.
  * These dependency upgrades should potentially be isolated into a separate pull request (rationale: change one aspect of a project at a time, to identify as much possible breakage -- because then we can resolve that -- as possible).

I'll place this pull request into draft status until the dependencies are updated.

marcellov7 (approved):

jayaddison:
I think these changes are ready for re-review; however I'm also planning to perform a git rebase to rearrange them into a more coherent series of commits.

jayaddison:
One mistaken CSS merge conflict resolution in 4494a73cfe5f5d07cc5cb1ebe84afd931ed44886 - fixing that momentarily..

jayaddison:
cc @kevin-brown - this PR collates some work from #6298 to add support for the beta versions of jQuery 4.0.0 - note that an update to a corresponding beta version of `jquery-mockjax` is also required because it required some compatibility updates too.

kevin-brown (approved):

kevin-brown:
Thanks a bunch for getting this one updated, it's definitely a lot easier to review now that the other QUnit-related pull requests have also been merged. 😄

PR #6334, opened 2024-08-19
Dependencies/CI: updates to prepare for jQuery 4.0.0-beta testing

jayaddison:
This pull request is dependency maintenance.

The following changes were made

- Migrate from the deprecated `node-sass` to `sass` (Dart SASS) library.
  - The CSS `outputStyle` setting is updated accordingly.
- Upgrade to `grunt-contrib-qunit` v10.1.1.
  - This requires a major version update for QUnit; here the latest v2.22.0 is selected.
- Upgrade GitHub Actions workflows to NodeJS v20 `iron` (the current Long-Term-Support release).

This changeset prepares for an upgrade to jQuery 4.0.0-beta2, and extracts unrelated dependency changes from PR #6332.

PR #6335, opened 2024-08-23
Fixups: post-merge updates related to PR #6241

jayaddison:
This pull request includes a

- [x] Bug fix
- [ ] New feature
- [ ] Translation

- Instead of referring to (QUnit) `test` in an integration test added by #6241, use the fully-qualified reference `QUnit.test` instead, as becomes standard after #6334.
- Add some guard conditions to `selection:update` event handlers in the integration tests, because these are not isolated from each other at test runtime, and otherwise may delay/timeout unrelated test cases.
- Regenerate distributable content.
- Regenerate NPM lockfile.

Relates-to / merge-resolution-for #6241 and #6334.

Edit: add additional fixup descriptions.

### DIFF
--- a/dist/js/select2.full.js
+++ b/dist/js/select2.full.js
@@ -3744,9 +3744,13 @@ S2.define('select2/data/ajax',[
       var $request = options.transport(options, function (data) {
         var results = self.processResults(data, params);
 
-        if (self.options.get('debug') && window.console && console.error) {
-          // Check to make sure that the response included a `results` key.
-          if (!results || !results.results || !Array.isArray(results.results)) {
+        if (results && results.results && Array.isArray(results.results)) {
+          results.results = results.results.map(
+            AjaxAdapter.prototype._normalizeItem
+          );
+        } else {
+          if (self.options.get('debug') && window.console && console.error) {
+            // Check to make sure that the response included a `results` key.
             console.error(
               'Select2: The AJAX results did not return an array in the ' +
               '`results` key of the response.'
--- a/dist/js/select2.js
+++ b/dist/js/select2.js
@@ -3744,9 +3744,13 @@ S2.define('select2/data/ajax',[
       var $request = options.transport(options, function (data) {
         var results = self.processResults(data, params);
 
-        if (self.options.get('debug') && window.console && console.error) {
-          // Check to make sure that the response included a `results` key.
-          if (!results || !results.results || !Array.isArray(results.results)) {
+        if (results && results.results && Array.isArray(results.results)) {
+          results.results = results.results.map(
+            AjaxAdapter.prototype._normalizeItem
+          );
+        } else {
+          if (self.options.get('debug') && window.console && console.error) {
+            // Check to make sure that the response included a `results` key.
             console.error(
               'Select2: The AJAX results did not return an array in the ' +
               '`results` key of the response.'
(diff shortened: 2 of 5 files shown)


jayaddison:
Hmm.. I have a feeling that the test failure here (a timeout, essentially) was what I encountered locally before adding a `defer="defer"` attribute to the `<script>` elements in the test HTML files, to guarantee that they are [loaded in-order](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script).

I think that somehow relates to the usage of AJAX in the tests, but I forget exactly why (or perhaps never quite pinpointed the reason).

jayaddison:
> Hmm.. I have a feeling that the test failure here (a timeout, essentially) was what I encountered locally before adding a `defer="defer"` attribute to the `<script>` elements in the test HTML files, to guarantee that they are [loaded in-order](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script).
> 
> I think that somehow relates to the usage of AJAX in the tests, but I forget exactly why (or perhaps never quite pinpointed the reason).

I fully admit that it may seem spammy or unnecessary, but I believe that #6336 should resolve this problem.

jayaddison:
> > Hmm.. I have a feeling that the test failure here (a timeout, essentially) was what I encountered locally before adding a `defer="defer"` attribute to the `<script>` elements in the test HTML files, to guarantee that they are [loaded in-order](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script).
> > I think that somehow relates to the usage of AJAX in the tests, but I forget exactly why (or perhaps never quite pinpointed the reason).
> 
> I fully admit that it may seem spammy or unnecessary, but I believe that #6336 should resolve this problem.

That pull request did not appear to resolve the problem, and so further investigation is required.

jayaddison:
@kevin-brown something about the `searching tags does not loose focus` integration test seems incompatible with #6241 -- I've commented it out here, and that allows the test suite to pass, but it's not an approach I like so I've moved this into draft status until the problem is figured out.

jayaddison:
The problem seems to be that `selection:update` event handling in the integration tests isn't isolated per-test-case.  Some ideas to resolve this:

  * Add some kind of per-test teardown that removes relevant event handling/selection-update resources (listeners, DOM contents, internal state?).
  * Attach once-only event handlers in each of the relevant integration test cases.
  * Add logic to ensure that each event-handler becomes essentially a no-op despite remaining attached during unrelated integration tests.

I attempted to add a [`QUnit.reset` handler from here](https://github.com/qunitjs/qunit/issues/630#issuecomment-52478879) to achieve the first; but this did not seem to work.

The second approach doesn't appear straightforward either; there is no `.once(...)` function available for event-listener attachment on `Select2` instances, nor is there a way to pass `{once: true}` as a parameter to `.on(...)`.

...so I've slightly-reluctantly implemented the third approach.  It doesn't genuinely provide any meaningful runtime isolation between the integration test cases, but it ensure that each event listener should only perform any actions within the scope of its parent test case.

jayaddison:
@kevin-brown it's not ideal code (details in my previous message if you're curious), but I believe this is ready for merge.

kevin-brown (approved):

kevin-brown:
I'll take the "temporary" hack for now. Thanks for looking into this and documenting your findings!

PR #6336, opened 2024-08-23
Tests: add `defer` attribute to all `<script>` elements so that they load in page order

jayaddison:
This pull request includes a

- [x] Bug fix
- [ ] New feature
- [ ] Translation

- Set the `defer="defer"` attribute on all `<script>` elements in the test HTML harnesses, so that scripts are loaded in the same order that they are written in the HTML.

Ref: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script

>  Scripts loaded with the `defer` attribute will load in the order they appear on the page. They won't run until the page content has all loaded, which is useful if your scripts depend on the DOM being in place (e.g. they modify one or more elements on the page).

### DIFF
--- a/tests/integration-jq1.html
+++ b/tests/integration-jq1.html
@@ -8,15 +8,15 @@
     <div id="qunit"></div>
     <div id="qunit-fixture"></div>
 
-    <script src="vendor/qunit-2.22.0.js" type="text/javascript"></script>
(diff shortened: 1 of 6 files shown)


jayaddison:
From local testing: I seem to be mistaken about this.  Including this change does not appear to resolve the timeout problems in AJAX-related tests.

PR #6341, opened 2024-09-26
feat: add selectedData support to AjaxAdapter

saaiful:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

- Introduced `selectedData` option in AjaxAdapter to allow pre-selected options to be programmatically appended and selected during initialization.
- Automatically selects and appends options from `selectedData` if provided.
- Maintains standard AJAX behavior for fetching options while ensuring that selectedData can be pre-loaded and displayed without additional input.
- Added safety checks to ensure selectedData is correctly processed even if not present or in an unexpected format.

### Example Use Case:
```javascript
$('#select').select2({
	ajax: {
		url: '/your-api-endpoint',
		dataType: 'json',
		processResults: function (data) {
			return { results: data.items };
		}
	},
	selectedData: [
		{ id: 1, text: 'val 1' }
	]
});
```

Multiple options can be selected:
```javascript
$('#select').select2({
	ajax: {
		url: '/your-api-endpoint',
		dataType: 'json',
		processResults: function (data) {
			return { results: data.items };
		}
	},
	multiple: true,
	selectedData: [
		{ id: 1, text: 'val 1' },
		{ id: 2, text: 'val 2' },
	]
});
```


kevin-brown:
Unfortunately I have to close this pull request (2 years later) since it's not one we're likely to support within Select2. As you can see with your change here, it's possible to do this before initializing Select2 and it's what we recommend in the docs for pre-selecting AJAX options.

PR #6362, opened 2025-02-18
scrollTop is called on the scroll content and always returns 0, should be on its parent

nielsvanzon:
This pull request includes a

- [x] Bug fix
- [ ] New feature
- [ ] Translation

- installing the scroll event on the parent of the scroll content
- getting the offset from the scroll content parent
- calling the scrollTop from the scroll content parent

Issues at the moment:
- You sometimes cannot scroll up when you have scrolled down.
- When there are more then 50 results the load more message stays and no more results are appended.
- When you reach the end of the list you scroll the page instead of stopping the propagation.

### DIFF
--- a/src/js/select2/dropdown/infiniteScroll.js
+++ b/src/js/select2/dropdown/infiniteScroll.js
@@ -37,7 +37,7 @@ define([
       self.loading = true;
     });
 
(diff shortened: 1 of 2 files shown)


nielsvanzon:
We found the issue in our styling.

Issue #6373, opened 2025-06-23
Whitespaces around tooltip / title.


antoineveldhoven:
I tend to format my options like this:

```TWIG
    <select class="select2">
        {{ foreach option in options }}
        <option value="{{ option.value }}">
            {{ option.text }}
        </option>
        {{ endforeach }}
    </select>
```

This renders into a tooltip like this:

![Image](https://github.com/user-attachments/assets/a489cc7b-fe9b-4c46-bbc5-8bd47a86a724)

To avoid this I can supply a title like this:

```TWIG
    <select class="select2">
        {{ foreach option in options }}
        <option value="{{ option.value }}" title="{{ option.text }}">
            {{ option.text }}
        </option>
        {{ endforeach }}
    </select>
```

But this feels a bit redundant.

I would like to suggest an optional ```trim()``` for the title.

PR #6374, opened 2025-06-23
Optional trimTitle in options.

antoineveldhoven:
Introduce optional trimTitle in options in order to avoid leading and trailing whitespaces in tooltip.

This pull request includes a

- [X] New feature

- Added trimTitle top options (defaults to false)
- Trims title on update() in order to compact tooltip when options set to true

Before:
```TWIG
    <select class="select2">
        {{ foreach option in options }}
        <option value="{{ option.value }}">
            {{ option.text }}
        </option>
        {{ endforeach }}
    </select>
```
afbeelding

After

```JavaScript
    $('.select2').select2({
        trimTitle: true
    });
```

afbeelding

Aiming to fix / contribute to #6373


biniyamchane25-jpg:
$('.select2').select2({
        trimTitle: true
    });

kevin-brown:
Unfortunately I'm going to have to close this pull request (1 year later) since it's not something we intend to introduce into Select2 with first-party support. This is something we recommend doing ahead of time prior to initializing Select2.

Issue #6409, opened 2026-05-30
pre-commit hooks


codingjoe:
Howdy 🤠,

great to see that the project is picking up steam again. Im sending my love from the Django community, will update to the latest release again. 

Now I have been quietly working on JavaScript version of pyupdate recently called [esupgrade](https://github.com/codingjoe/esupgrade). Since I know that  @kevin-brown is a fellow Python person too, this probably is a very familiar concept. 

Since I maintain Django-Select2 I frequently used this repos code for real word testing and hardening.

Finally, I think I have reached a point where I am happy. As a side effect I am sitting on a mountain of non-breaking code updates for this repo. 

Are you interested in me sharing those?

I would also offer to setup a more sophisticated pre-commit suite, which might help to streamline future maintenance of this repo. Especially with pre-commit.ci giving new contributors a lift.

But that's always personal preference.

Cheers
-Joe

kevin-brown:
Feel free to submit PRs for that. Just call out anything unusual.

codingjoe:
Thanks for considering @kevin-brown! I'll do so in bite-sized chunks. Starting with the basic setup in #6410

PR #6412, opened 2026-05-31
Ref #6409  -- Add Biome pre-commit hook for JavaScript, JSON and CSS files

codingjoe:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

- Add Biome pre-commit hook for JavaScript, JSON and CSS files

> [!IMPORTANT]
> I chose Biome since JSHint is no longer maintained. However, this hook is currently only formatting, not linting. This would involce more code changes, which are better suited for a separate PR. Once Biome is used for linting, we can drop JSHint.

Ref #6409

### DIFF
--- a/src/js/jquery.select2.js
+++ b/src/js/jquery.select2.js
@@ -30,8 +30,10 @@ define([
 
           if (instance == null && window.console && console.error) {
             console.error(
-              'The select2(\'' + options + '\') method was called on an ' +
-              'element that is not using Select2.'
+              "The select2('" + // jshint ignore:line
+                options +
+                "') method was called on an " + // jshint ignore:line
+                'element that is not using Select2.'
             );
           }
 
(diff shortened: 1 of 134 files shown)


src/js/jquery.select2.js
```diff
@@ -30,8 +30,10 @@ define([
 
           if (instance == null && window.console && console.error) {
             console.error(
-              'The select2(\'' + options + '\') method was called on an ' +
-              'element that is not using Select2.'
+              "The select2('" + // jshint ignore:line
```
kevin-brown, line 33: Glancing at this on mobile, what's the reason for needing the ignore here? Is it the quotes?

kevin-brown: Yup it's definitely the quotes.

codingjoe: Yes, sorry. I forgot to highlight that. The comments will go with JSHint.

Issue #6417, opened 2026-06-02
Docs: Cleanup items


kevin-brown:
* Either drop the committers plugin or remove bots from being shown
* Dark mode theme does not look correct
* Material has weird padding/margins which mess up multi-select boxes
* Favicon was removed on accident and needs to be added back

/cc @codingjoe

codingjoe:
I'll get on it :)

codingjoe:
Thanks @kevin-brown for noticing. I just wrapped up my PR. I believe I caught all issues, but please double check.

codingjoe:
@kevin-brown can be closed… or amended if you find anything else.

PR #6419, opened 2026-06-02
Ref #6417 -- Address static docs regression

codingjoe:
This pull request includes a

- [x] Bug fix

- Exclude bots from docs' author widget
- Restore favicon
- Drop obsolete flag assets (I previously switched them to a CDN)
- Add dark mode theme for mkdocs
- Fix select box paddings

### single
CleanShot 2026-06-02 at 13 40 46@2x

### multiple
CleanShot 2026-06-02 at 13 39 49@2x

Ref #6417


kevin-brown:
Note for future me when reviewing this:

Mkdocs Material uses https://github.com/ojacques/mkdocs-git-committers-plugin-2, not the old git committers plugin.

I accidentally switched one of the templating pages away from using the CDN, that should be switched back before merging.

docs/README.md
```diff
@@ -23,7 +23,7 @@ _The jQuery replacement for select boxes_
   and many other highly used options.
 </p>
 
-<div class="grid cards s2-docs-features" markdown>
```
kevin-brown, line 26: I think this is the CSS class used in the custom CSS for arranging the icon/title/subtitle for each card.

codingjoe: Oh, you added this. Sorry. I thought I forgot to clean up code in my previous commit. I restored it :)

codingjoe:
> Mkdocs Material uses https://github.com/ojacques/mkdocs-git-committers-plugin-2, not the old git committers plugin.

@kevin-brown so are we. They just use the same config name.
https://github.com/select2/select2/blob/293c7511904d4238e497c973532df1c973763414/pyproject.toml#L10

codingjoe:
FYI, the template for MkDocs is a little crude. But I am hoping to introduce template variables for Select2 soon, which should simplify this whole process.

kevin-brown (approved):

kevin-brown:
Last night was a mad dash to switch to GitHub Pages since I broke Grav (whoops) and didn't have an easy way to fix it. So I expect there will be a handful of PRs to make things a bit cleaner in the docs.

PR #6432, opened 2026-06-28
Add Norwegian Nynorsk (nn) translation

arvindfroi:
Adds a Norwegian **Nynorsk** (`nn`) translation. Select2 already ships Bokmål (`nb`); this adds Norway's other official written standard, same message set.

Disclosure: prepared with AI assistance and reviewed by me as a native Norwegian speaker.

### DIFF
--- a/src/js/select2/i18n/nn.js
+++ b/src/js/select2/i18n/nn.js
@@ -1,0 +1,33 @@
+define(function () {
+  // Norwegian (Nynorsk)
+  return {
+    errorLoading: function () {
+      return 'Kunne ikkje hente resultat.';
+    },
+    inputTooLong: function (args) {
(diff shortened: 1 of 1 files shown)


kevin-brown:
I'm not a native speaker so I don't have as much information on the language as I wish I did.

I see ISO-639 has three language codes for Norwegian (`no`, `nb`, `nn`). I recognize that `nn` (this PR) is different from the existing `nb` translation. But for some reason I thought the Select2 translation for `no` matched `nn`? Many years ago when approving the `nb` translation I could have sworn that's what someone said.

I'm totally fine introducing `nn` and/or fixing the language code `no` to match what one would natively expect. I get the impression this is not an issue unique to Select2 when it comes to translations.

PR #6456, opened 2026-09-11
docs: clarify closeOnSelect applies to single selects

### LINKED ISSUE #6455: Documentation bug: closeOnSelect applicable to singe-select controls too
The [documentation](https://select2.org/dropdown/#forcing-the-dropdown-to-remain-open-after-selection) for closeOnSelect says 

> Note that this option is only applicable to multi-select controls.

But the option does affect singe-select controls.

Since fixing this in code would be a breaking change, I suppose it's a documentation bug.

sridharkalaibala:
This pull request includes a

- [x] Bug fix (documentation)
- [ ] New feature
- [ ] Translation

The documentation currently says `closeOnSelect` only applies to multi-select controls. Setting it to `false` also keeps a single-select dropdown open after selecting a result. Clarify that the option applies to both types.

Fixes #6455.

Validation:
- `python -m mkdocs build --strict` passed using the documentation dependencies from `pyproject.toml` in an isolated Python 3.14 environment. The generated dropdown page contains the corrected sentence. CI-only Git metadata plugins were disabled by the repository's default configuration.
- Checked `defaults.js`: the CloseOnSelect decorator is applied outside the single/multiple branch.
- Exercised the checked-in Select2 4.1.0 bundle with jQuery 3.5.1 in jsdom: selecting Beta in single and multiple controls leaves the dropdown open with `closeOnSelect: false` and closes it with `true` (four passing cases). This was a local behavior check, not a real-browser or full QUnit run.
- `git diff --check` passed. Only documentation changed.

AI assistance: OpenAI Codex helped investigate, validate, and prepare this change.

### DIFF
--- a/docs/dropdown.md
+++ b/docs/dropdown.md
@@ -72,7 +72,7 @@ $('#mySelect2').select2({
 });
 ```
 
-Note that this option is only applicable to multi-select controls.
+This option applies to both single-select and multi-select controls.
 
 > [!NOTE]
 > If the [`CloseOnSelect` decorator](advanced/default-adapters/dropdown.md#closeonselect) is not used (or `closeOnSelect` is set to <code>false</code>), the dropdown will not automatically close when a result is selected. The dropdown will also never close if the <kbd>ctrl</kbd> key is held down when the result is selected.


lbesecker195 (approved):
Verified in `defaults.js`: the `CloseOnSelect` decorator is applied whenever `options.closeOnSelect` is set, with no check on `multiple`, so the old "multi-select only" note was wrong. LGTM.

sridharkalaibala:
Thanks for checking the decorator path, @lbesecker195. That confirms why the note needed correcting; the PR keeps the change limited to the documented `closeOnSelect` behavior.

PR #6457, opened 2026-09-14
a11y: retain 'aria-label' and 'aria-labelledby' attributes from the origin 'select' element

### LINKED ISSUE #4930: Feature proposal: Copy aria-label attribute to select 2 input field
Currently if you set an aria-label attribute on your select menu for a multiple select, the aria-label is not copied to the select2 input field. It would be useful if this could be copied for improved accessibility.

jayaddison:
The absence of an ARIA label on the HTML `input` elements created by Select2 in `multiple` mode is flagged by Deque AXE as an accessibility problem, and prior to this change any `aria-label` attribute on the source HTML `select` element(s) are lost in the resulting Select2 input controls.

This change detects any existing `aria-label` [and `aria-labelledby`](https://github.com/select2/select2/issues/4930#issuecomment-329474566) attributes and copies them to the destination Select2 selection control, following the same approach used for the `tabindex` attribute.

This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

- Copy `aria-label` and `aria-labelledby` attributes, when present, from the origin `select` element onto the Select2 selection element.

May resolve #4930.

Edit: add `aria-labelledby` in addition to `aria-label`.


kevin-brown (approved):

jayaddison:
Thank you @kevin-brown

Issue #5246, opened 2018-03-08
Custom MultipleSelection with CustomAdapter no search inline 


cyrilus:
## Prerequisites
 [ x] I have searched for similar issues in both open and closed tickets and cannot find a duplicate
[x ] The issue still exists against the latest master branch of Select2
[x ] This is not a usage question (Those should be directed to the community)
[x ] I have attempted to find the simplest possible steps to reproduce the issue
[x ]I have included a failing test as a pull request (Optional)

##  The Problem
When I try to use CustomSelectionAdapter the search never appear in the input
the
```
 '<li class="select2-search select2-search--inline">' +
        '<input class="select2-search__field" type="search" tabindex="-1"' +
        ' autocomplete="off" autocorrect="off" autocapitalize="none"' +
        ' spellcheck="false" role="textbox" aria-autocomplete="list" />' +
      '</li>'
```
never appear all the rest is working perfectly

## Steps to reproduce the issue

```
$.fn.select2.amd.require([
        "select2/utils",
       "select2/selection/multiple",
       "select2/selection/placeholder",
    ], function (Utils, MultipleSelection, Placeholder) {

      var CustomSelectionAdapter = Utils.Decorate(MultipleSelection, Placeholder);
      CustomSelectionAdapter = Utils.Decorate(CustomSelectionAdapter, EventRelay);
   
      CustomSelectionAdapter.prototype.update = function(data) {
              // I vave tried lot of thing but it never work
        }
 return CustomSelectionAdapter;
  }); 
```

Don't know if it 's a bug or a lack of documentation

elieux:
The core builds up the adapter from provided options, incl. the search, placeholder etc., but only if you don't specify your own adapter [1]. If you do, it completely skips this. In this case, you have to add the search box decorator (from `select2/selection/search`) yourself.

[1] https://github.com/select2/select2/blob/c90adc102b5e9b2faea2b871b1c764416584ee07/src/js/select2/defaults.js#L187-L233

cyrilus:
Thanks for your help  but I really can't figure out  ofhow to do it
everything i have tried was just a dead end

elieux:
I figured from your code example that it would be enough information.  Try this then:

```
$.fn.select2.amd.require([
       "select2/utils",
       "select2/selection/multiple",
       "select2/selection/placeholder",
       "select2/selection/search",
    ], function (Utils, MultipleSelection, Placeholder, SelectionSearch) {
[...]

Issue #4797, opened 2017-02-16
disable search on select boxes with multiple selection


chlorophyllkid:
Hi everyone,

as the title indicates I'm looking for a possibility to disable the search on selects with multiple selection.

I already know there is: 
`minimumResultsForSearch: Infinity`

And it works fine on selects with single selection, but doesn't seem to have any effect on select boxes with multiple selection.

Oh, and I'm currently using select2 4.0.3..

Here is a fiddle [http://jsfiddle.net/jEADR/3608/](http://jsfiddle.net/jEADR/3608/)

thanks in advance :)

chlorophyllkid:
My current workaround is this:

` $('select').on('select2:opening select2:closing', function( event ) {`
    `var $searchfield = $( '#'+event.target.id ).parent().find('.select2-search__field');`
    `$searchfield.prop('disabled', true);`
 `});`

nflorentin:
+1

Sartoric:
+1

3xcellent:
Also hitting this.  Would really like to have the search option disabled for multiselects with finite options.  My work-around (not very elegant):

  setTimeout(function() {
    $('input[type=search]').attr('disabled', 'disabled');
  });

alexweissman:
@chlorophyllkid your solution seems to work well, so I am adding it to the documentation.

alexweissman:
Now documented here: https://select2.org/searching#multi-select

thenitai:
I somehow couldn't get the proposed solution to work. However, this one works for me every time:

```
$('#selectid').on('select2:open select2:opening select2:closing', function( event ) {
  $('.select2-results').css('display', 'none');
});
```

tabarjack:
For those who might wind up here from a google search like I did: 

I used this and it works even on a refresh. Note that I haven't tested with sending values to the server because it's only for an in-page filter (and that works so no reason sending to server wouldn't, mentioning just in case).

`obj.on('select2:opening', function(e){ $('.select2-search__field').prop(readonly, true); }`

slbteam08:
> My current workaround is this:
> 
> ```
> $('select').on('select2:opening select2:closing', function( event ) {
>     var $searchfield = $( '#'+event.target.id ).parent().find('.select2-search__field');
>     $searchfield.prop('disabled', true);
> });
> ```

This documented solution works only from the second click on iPad. On the first click before the input is disabled, the iPad keyboard occurs already. Are there any events that can be used before the first click?

[...]

PR #4480, opened 2016-07-14
Added an option to disable the selection search box in multiple mode

dimmelsw:
This pull request includes a
- [x] New feature

The following changes were made
The option `disableSelectionSearch` is added to hide the search input box in multiple selects.

Issue #2965, opened 2015-01-25
Combine multiple selections into a single summary (e.g., "3 items selected")


piernik:
Is it possible in select2 4.0 that  multiple select will be displayed only in one line (now with too many selected options field has two line - is higher).
Something like in bootstrap select (http://silviomoreto.github.io/bootstrap-select/ example with `<select class="selectpicker" multiple data-selected-text-format="count">`).
Ex if selected 3 option text will be displayed like `2 options selected` instead of showing them.

piernik:
Can anyone give me an anwser pealse?

kevin-brown:
Yes, it is possible. No, it is not possible by default.

You need a custom [`selectionAdapter`](https://select2.github.io/options.html#selectionAdapter) that extends from [`MultipleSelection`](https://github.com/select2/select2/blob/4c563a3d068d66a30bb949fe4abd71f7402dc53e/src/js/select2/selection/multiple.js#L79-L93). You need to override the `update` method to render it the way you are looking for, instead of rendering the individual options.

piernik:
ok - thanks

ova2:
I like how the multiple items can be chosen by bootstrap-select (http://silviomoreto.github.io/bootstrap-select/) Unlike Select2 they can be chosen at once. Select2 close the dropdown after each selection and you have to open it again to select second, third, etc. items. Do you plan to add this behavior to Select2?

chaimann:
@ova2 
you can pass an option `closeOnSelect: false` when initializing select2:

    $('.select').select2({
        ...,
        closeOnSelect: false,
        ...
    });

thejahweh:
Do I see it right that this has no priority and everybody who wants to select more than 5 options has to use another library?

I have some cases with more than 100 selectable options and have to say that this looks amazingly ugly.

kevin-brown:
> Do I see it right that this has no priority

Yes. Like most enhancements that aren't thinly veiled bugs, this ticket does not have any milestone currently attached to it.

> everybody who wants to select more than 5 options has to use another library

Not necessarily. Select2 doesn't prevent anyone from making this work, we just don't have any quick and simple recommendations for adjusting the CSS to make this work. If you have any CSS snippets that make this work (I have a feeling it's going to involve `overflow: auto` and preventing text wrapping), that would definitely help everyone who is looking to do this.

[...]

PR #6462, opened 2026-09-22
Add dropdownSearch and selectionSearch options for multiple selects

TheWitness:
## Summary

For multiple selects, the search box currently always renders inline inside the
selection container (the `SelectionSearch` decorator). That's fine for composing
a selection, but the control's own height changes whenever the dropdown opens
(the search box only exists while it's part of that container), which shifts
the layout of the surrounding page - noticeable in dense forms, filter bars,
and similar UIs.

Single selects don't have this problem, since their search box lives in the
floating dropdown panel (`DropdownSearch` + `AttachBody`) instead of inline.

This PR adds two options so a multiple select can opt into that same
dropdown-search behavior, and/or opt out of a search box entirely:

- **`dropdownSearch`** (default `false`): when `true`, a multiple select
  searches from the dropdown panel instead of inline - exactly like a single
  select. No effect on single selects (already the default there).
- **`selectionSearch`** (default `true`): set to `false` to hide a multiple
  select's search box entirely. This replaces the `select2:opening`/
  `select2:closing` + `disabled`-prop workaround previously documented for
  this (#4797) with a real, first-class option. No effect on single selects,
  or when `dropdownSearch` is `true`.

Both options are pure adapter composition in `defaults.js` - reusing the
existing `Dropdown`, `DropdownSearch`, `MinimumResultsForSearch`, `AttachBody`,
`MultipleSelection`, and `SelectionSearch` adapters. No new adapter classes,
no behavior changes to any existing adapter, and default configuration for
both single and multiple selects is unchanged.

Related issues this addresses: #189, #1345, #2965, #4480, #5246.

## Docs

- New "Searching from the dropdown for multi-select controls" section in
  `docs/searching.md`, with a live example.
- Updated the multi-select "Hiding the search box" example to use
  `selectionSearch: false` instead of the old `disabled`-prop workaround.
- Added the (previously undocumented) `SelectionSearch` decorator entry to
  the built-in adapters reference, and noted the new option tie-ins on both
  `DropdownSearch` and `SelectionSearch`.
- Added both new options to the configuration options table.

## Tests

`tests/options/dropdownSearch-tests.js` covers adapter composition for all
combinations of the two options, plus a regression guard that default
behavior (no options set) is unchanged.

## Context

I maintain [Cacti](https://github.com/Cacti/cacti), an open-source
network/SNMP graphing and monitoring platform, and we recently converted a
large batch of legacy jQuery UI selectmenu/multiselect controls to Select2.
One of our filter controls is a multi-select that displays a condensed
"N Selected" summary instead of per-option chips (see screenshot below) -
exactly the case where an inline search box growing/shrinking the control on
open was disruptive to the surrounding filter bar's layout. We worked around
it locally with manual adapter composition (essentially what this PR turns
into first-class options), and it seemed worth upstreaming properly with
tests and docs rather than leaving every consumer to reinvent it.

Thanks for keeping this library going after all these years - it's a
pleasure to finally get to use it properly in Cacti.

- TheWitness


TheWitness:
Here is a screen capture of how it looks in the update Cacti.  Will release with 1.2.32 version of Cacti.  Thanks.  Been at this for 25 years now :)

image

